PR #2, opened 2015-04-23
Add log entrt after each request

robyoung:


dmutils/logging.py
```diff
@@ -25,6 +25,11 @@ def init_app(app):
     def after_request(response):
         request_id_header = current_app.config['DM_REQUEST_ID_HEADER']
         response.headers[request_id_header] = request.request_id
+
+        current_app.logger.info('Response: %s %s %s',
+                                request.method,
+                                request.url,
+                                request.status)
```
allait: Shouldn't this be `response`?

robyoung: Yes it should, good spot.

PR #8, opened 2015-04-29
Add DataAPIClient and tests

robyoung:


dmutils/apiclient.py
```diff
@@ -6,6 +6,19 @@
 logger = logging.getLogger(__name__)
 
 
+class APIError(StandardError):
```
allait: Should probably inherit from `HTTPError`. Also, I think `StandardError` is causing the py3 tests to fail.

PR #14, opened 2015-05-05
Add `get_status` methods and tests.

pcraig3:
Added a `get_status` method which will be inherited by our (at present, two) API clients, and added a couple tests.

### DIFF
--- a/dmutils/apiclient.py
+++ b/dmutils/apiclient.py
@@ -57,6 +57,10 @@ class BaseAPIClient(object):
             logger.exception(e.message)
             raise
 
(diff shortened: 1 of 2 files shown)


pcraig3:
I'm only testing what's generic to both responses, and so both of them are identical.  
Is there any way to / value in generalising the test?
Or, alternatively, should I make the tests mirror more closely the specific responses each API sends back?

allait:
Should bump the `dmutils/__init__py` `__version__` and tag it so that we have something to use in the app requirements

PR #17, opened 2015-05-06
Add user related methods

robyoung:
These are largely taken from the supplier frontend.


TheDoubleK:
I'm not sure why we are moving these things into utils?  Both this and https://github.com/alphagov/digitalmarketplace-utils/pull/18 seem like they will only be used in the supplier app, and pulling them out into utils makes debugging harder and utils more cluttered than it needs to be.

robyoung:
The most important reason I see is that it's more confusing to have multiple ways of talking to the API that work differently.

Also, I think they will be used by other apps. The user model in the data API already has three roles, buyer, supplier and admin. While the login flow may only happen in one of the apps accessing the user data is likely to be needed in all apps (for authorisation). As for the supplier information in #18, this is already likely to be needed in the suppliers and the buyers app.

robyoung:
another good reason for having API clients centralised is for adding easy to forget things like the request ID (which I forgot to add).

PR #20, opened 2015-05-06
Better error handling in `get_status` method

pcraig3:
Apache generates an 'Internal Server Error' page if any of our upstream APIs return 500-level responses.
So we'll just have the API Client return its own error if that happens.
Also required requests package.

### DIFF
--- a/dmutils/apiclient.py
+++ b/dmutils/apiclient.py
@@ -3,7 +3,6 @@ import logging
 
 import six
 import requests
-from requests import ConnectionError  # noqa
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,5 +26,6 @@
     install_requires=[
         'Flask>=0.10',
         'six==1.9.0',
+        'requests==2.7.0',
```
TheDoubleK: Is there a reason why we are doing imports differently in this repository than in all the others?  In the others we have `requirements.txt` rather than a `setup.py`.  I'm not against this, but it would be nice to do things the same way in all apps for consistency.

dmutils/apiclient.py
```diff
@@ -64,13 +63,14 @@ def get_status(self):
...
-            return {
-                "status": "error",
-                "message": "{}".format(e.message),
-            }
+            try:
+                return e.response.json()
+            except ValueError:
+                return {
+                    "status": "error",
+                    "message": "{}".format(e.message),
```
TheDoubleK: Are we sure that `e` will always have a `.message`?

robyoung: It's part of the standard interface of `Exception` so it should do. (for very large values of should)

setup.py
robyoung: This project is a python package. In order to be installable by pip it must have a `setup.py` which tells pip how to install it (including it's dependencies). What we could do is have `setup.py` read the requirements from a `requirements.txt` file. Ultimately this is a different kind of thing to the other repos though.

PR #25, opened 2015-05-07
Add `delete` method to SearchApiClient.

pcraig3:
Also meant adding a `_delete` method to the BaseApiClient and some tests.
If we try to delete a service that doesn't exist, no (harm|foul).  We'll return `None` but won't raise an error.

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -85,7 +85,7 @@ def service():
         "datacentreTier": "TIA-942 Tier 3",
         "datacentresSpecifyLocation": True,
         "datacentresEUCode": False,
-    }
+        }
 
 
 class TestSearchApiClient(object):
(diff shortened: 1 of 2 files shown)


pcraig3:
Actually, is anyone still using the [`ConnectionError`](https://github.com/alphagov/digitalmarketplace-utils/pull/25/files#diff-fefaa03059c9b9ebb32ebdd9399fa74eR6) that we're importing?

I asked @allait to put it in last week for my `/_status` endpoints, but I'm no longer using them for anything, and my PyCharm is flagging it as an unused import.

tests/test_apiclient.py
```diff
@@ -85,7 +85,7 @@ def service():
         "datacentreTier": "TIA-942 Tier 3",
         "datacentresSpecifyLocation": True,
         "datacentresEUCode": False,
-    }
```
allait: It might be me, but I think this looked better :sob:

allait:
@pcraig3 I don't think anyone is using it right now, so it might be safe to remove. But it's also useful, because we give no other way for apps to handle the connection errors (and we don't handle them in utils) apart from importing the original exception from requests.

Usually "unused imports" are silenced with `# noqa`, I'm not sure why PyCharm doesn't respect it, but maybe there's another way to silence the message.

PR #28, opened 2015-05-08
Add metrics

robyoung:
Add a very basic client for sending metrics to CloudWatch. It tries to
follow the pystatsd interface but so far only implements the `inc`
method.

The timestamp is explicitly set rather than allowing CloudWatch to set
it so that events line up better with our logs.

### DIFF
--- a/dmutils/metrics.py
+++ b/dmutils/metrics.py
@@ -1,0 +1,43 @@
+import copy
+from datetime import datetime
+
+from boto.ec2.cloudwatch import connect_to_region
+
+
+def client(region, namespace, default_dimensions=None):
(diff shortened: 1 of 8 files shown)


dmutils/metrics.py
```diff
@@ -0,0 +1,84 @@
...
+        self.name = name
+
+    def __enter__(self):
+        self.start = datetime.now()
+
+    def __exit__(self, *exc):
+        elapsed = datetime.now() - self.start
+        self.client._put_metric(
+            self.name,
+            int(elapsed.total_seconds() * 1000),
```
allait: What about `elapsed.total_seconds() * 1000 + (elapsed.microseconds / 1000)`?

PR #39, opened 2015-05-22
[#94576802] Update Json api format

TheDoubleK:
See this story in Pivotal: https://www.pivotaltracker.com/story/show/94576802 - the main aim is to make API response formats consistent across the Data and Search APIs.

There are FOUR related pull-requests to complete this story.  The other three are:
https://github.com/alphagov/digitalmarketplace-api/pull/132
https://github.com/alphagov/digitalmarketplace-search-api/pull/42
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/60

Changes in the search API client here are to return the whole response now as metadata is under a separate key from the services found.

See these gists for examples of the new API responses:
[Example services data API response](https://gist.github.com/TheDoubleK/0d7cd9363440e700a50d)
[Example suppliers data API response](https://gist.github.com/TheDoubleK/4fde0400e052f20ed2c0)
[Example services search API response](https://gist.github.com/TheDoubleK/8e82782502e611faabcc)

### DIFF
--- a/dmutils/apiclient.py
+++ b/dmutils/apiclient.py
@@ -184,8 +184,7 @@ class SearchAPIClient(BaseAPIClient):
             params['filter_{}'.format(filter_name)] = filter_values
 
         response = self._get(self._url("/search"), params=params)
-
-        return response['search']
+        return response
 
     def _convert_service(
             self,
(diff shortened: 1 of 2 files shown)


robyoung:
Could we add code into the API client to uncouple all the other pull requests? As far as I can tell, at the moment they all need to go out together to avoid breaking things.

TheDoubleK:
Is there a user need? Given that none of this is live yet I think the breaking change would be Ok. I'll take a look at fixing it though.

robyoung:
The benefits I can see are more to do with practicing our process before we go live. We get to practice doing releases that have some backwards compatibility breaking change. I agree, it's not the end of the world if we don't but it's a convenient opportunity to test out how we approach that kind of change.

PR #50, opened 2015-06-08
Use PEP8 in tests and fix existing PEP8 errors

TheDoubleK:
All our other projects enforce PEP8; we should do the same here for consistency.


pcraig3:
Glad to see that all of my recent changes were causing problems. :+1:

PR #51, opened 2015-06-08
Pass in path for `get_version_label` function

pcraig3:
Versions files are added to the root of each app when packaged for deploy.
Moving `get_version_label` to dmutils means it was looking for a relative path in the wrong place.  

Solution is for each app to pass its respective path.


robyoung:
If we use the current working directory (`os.getcwd()`) we wouldn't have to apply any change to all the apps. The apps are always run from the root of the app and the `version_label` file is always placed in the root of the app.

PR #52, opened 2015-06-09
Adding drafts api calls

minglis:
Added wrappers for all the draft API calls.

- This enables client calls for the underlying APIs.

### DIFF
--- a/dmutils/apiclient.py
+++ b/dmutils/apiclient.py
@@ -13,7 +13,6 @@ from flask import has_request_context, request, current_app
 
 logger = logging.getLogger(__name__)
 
-
 REQUEST_ERROR_STATUS_CODE = 503
 REQUEST_ERROR_MESSAGE = "Request failed"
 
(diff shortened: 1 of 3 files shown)


dmutils/apiclient.py
```diff
@@ -210,6 +209,57 @@ def init_app(self, app):
         self.base_url = app.config['DM_DATA_API_URL']
         self.auth_token = app.config['DM_DATA_API_AUTH_TOKEN']
 
+    def list_draft_service(self, supplier_id):
```
allait: I like the `list_` prefix more, but the rest of the methods are named `find_{services, suppliers}`, so this should probably be `find_draft_services`

dmutils/apiclient.py
```diff
@@ -210,6 +209,57 @@ def init_app(self, app):
...
+        return self._get(
+            "/draft-services?supplier_id={}".format(supplier_id)
+        )
+
+    def fetch_draft_service(self, service_id):
```
allait: Same as with `find_`: `get_draft_service` would be more consistent

dmutils/apiclient.py
```diff
@@ -210,6 +209,57 @@ def init_app(self, app):
...
+        return self._put(
+            "/services/{}/draft".format(service_id),
+            data={
+                "update_details": {
+                    "updated_by": user,
+                    "update_reason": "deprecated",
+                },
+            })
+
+    def edit_draft_service(self, service_id, service, user):
```
allait: `update_`

PR #54, opened 2015-06-10
Require FeatureFlags extension from pypi

pcraig3:
Pull request was merged with the fix we asked for, so we can require the extension as normal instead of relying on my custom forked version.


quis:
Link to the relevant pull request for posterity: https://github.com/trustrachel/Flask-FeatureFlags/pull/13

pcraig3:
:+1:

PR #55, opened 2015-06-10
Added audit types as a unique enum

minglis:
Added audit type enum,

This is too hold allowed values for audit events so these can be shared.


TheDoubleK:
Just to clarify: Is this instead of using a database enum which would be a pain to update every time we have a new audit type? i.e. the audit event table will just have string values for audit type, but when we create audit events we should only use values from this enum?

Seems like a good plan to me. Much easier to add to the list here than to do a database migration to change a database type.

PR #57, opened 2015-06-11
Add presenter for boolean values

quis:
"Yes" and "No" is friendlier than "True" and "False"

Tests are better than no tests.

### DIFF
--- a/dmutils/presenters.py
+++ b/dmutils/presenters.py
@@ -28,10 +28,10 @@ class Presenters(object):
             "url": value or "",
             "filename": value.split("/")[-1] or ""
         }
-        
+
     def _boolean(self, value):
-        if value == True:
+        if value is True:
             return "Yes"
-        if value == False:
+        if value is False:
             return "No"
-        return None
+        return ""
--- a/dmutils/presenters.py
+++ b/dmutils/presenters.py
@@ -28,3 +28,10 @@ class Presenters(object):
             "url": value or "",
             "filename": value.split("/")[-1] or ""
         }
+        
+    def _boolean(self, value):
+        if value == True:
+            return "Yes"
+        if value == False:
+            return "No"
+        return None
--- a/dmutils/presenters.py
+++ b/dmutils/presenters.py
@@ -30,8 +30,8 @@ class Presenters(object):
         }
 
     def _boolean(self, value):
-        if value == True:
+        if value is True:
             return "Yes"
-        if value == False:
+        if value is False:
             return "No"
         return ""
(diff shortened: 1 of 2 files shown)


tests/test_presenters.py
```diff
@@ -0,0 +1,61 @@
...
+import unittest
+
+from dmutils.presenters import Presenters
+presenters = Presenters()
+
+
+class TestPresenters(unittest.TestCase):
+    def test_service_id(self):
+        G5 = presenters.present(
+            "5.G5.12345",
```
TheDoubleK: G5's are generally more like `5.G5.1234.567`, but probably doesn't really matter here.

dmutils/presenters.py
```diff
@@ -28,3 +28,10 @@ def _upload(self, value):
             "url": value or "",
             "filename": value.split("/")[-1] or ""
         }
+
+    def _boolean(self, value):
+        if value is True:
```
TheDoubleK: For booleans you can just do `if value`

tests/test_presenters.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        nothing = presenters.present(
+            None,
+            {"type": "boolean"}
+        )
+        empty = presenters.present(
+            "",
+            {"type": "boolean"}
+        )
+        self.assertEquals(yes, "Yes")
```
TheDoubleK: This suggests that, due to deprecation, "you should use whatever you like for Python 2.x, but tend toward assertEqual for Python 3".
http://stackoverflow.com/questions/930995/assertequals-vs-assertequal-in-python

dmutils/presenters.py
quis: Would that also return `True` if you sent it `"string which isn't blank"` or `1`? Don't think that's desired behaviour.

tests/test_presenters.py
quis: Good point, addressed in 986b2e8897b48991cb234ab5a3ea262917a1e4ff

dmutils/presenters.py
TheDoubleK: It would, yes, if any old stuff gets passed in to it.  In that case maybe it should be `if value == True` - the `is` operator is for testing object equality, not value equality. (It will almost certainly work here, but is not guaranteed to in future if implementation of Python changes).

quis: Thanks for the Python learnings @TheDoubleK, addressed in 95b04b9

quis: @TheDoubleK PEP8 disagrees with you :(

Reverted in fd31766

TheDoubleK: Sorry...

TheDoubleK: We should ditch PEP8

PR #62, opened 2015-06-15
Make content loader handle dependencies

quis:
The format of the question content has changed from this:
``` yaml
question: "What is your service name?"
dependsOnLots: "SaaS, PaaS"
```

to this:

``` yaml
question: "What is your service name?"
depends:
  -
    "on": "lot"
    "being":
      - SaaS
      - PaaS
```

(alphagov/digital-marketplace-ssp-content#42)

This pull request adds a `filter` method to the content loader, which works like this:
``` python
  content.filter({
    "lot": "SaaS",
    "serviceName": "My first service",
    …
  })
```

Getting `content.sections` will then have only the questions relevant to the
current service, which means that the views don't have to do this kind of thing:
``` jinja
{% if service_data['lot']|lower in question['depends_on_lots'] %}
```

This pull request also adds tests, which the content loader did not previously have.

_This represents a breaking change and so releasing it will have to be reflected in the version number of utils_

### DIFF
--- a/tests/test_content_loader.py
+++ b/tests/test_content_loader.py
@@ -1,0 +1,49 @@
+# coding=utf-8
+
+import unittest
+import mock
+
+import os
+import tempfile
+import yaml
+
+from dmutils.content_loader import ContentLoader
+
+
+class MonkeyPatch:
+    def __init__(self, mocked_content=""):
+        self.mocked_content = mocked_content
+
+    def read_yaml_file(self, yaml_file):
+        return yaml.load(self.mocked_content[yaml_file])
+
+    def yaml_file_exists(self, file):
+        return True
+
+
+class TestContentLoader(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        ContentLoader._yaml_file_exists = MonkeyPatch().yaml_file_exists
+
+    def test_a_simple_question(self):
+        ContentLoader._read_yaml_file = MonkeyPatch({
+            "manifest.yml": """
+                -
+                  name: First section
+                  questions:
+                    - firstQuestion
+            """,
+            "folder/firstQuestion.yml": """
+                question: 'First question'
+                dependsOnLots: 'SaaS'
+            """
+        }).read_yaml_file
+        content = ContentLoader(
+            "manifest.yml",
+            "folder/"
+        )
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


tests/test_content_loader.py
```diff
@@ -0,0 +1,270 @@
...
+        self.mocked_content = mocked_content
+
+    def read_yaml_file(self, yaml_file):
+        return yaml.load(self.mocked_content[yaml_file])
+
+    def yaml_file_exists(self, file):
+        return True
+
+
+class TestContentLoader(unittest.TestCase):
```
allait: Modifying `ContentLoader` here will keep the replaced methods after the test ends. I think the best way to do this is to use `mock.patch("dmutils.content_loader.ContentLoader._yaml_file_exists")` and `mock.patch("dmutils.content_loader.ContentLoader._read_yaml_file")` either in the `setUp`/`tearDown` methods or as class/method decorators

quis:
@allait Does this look better to you?
[/tests/test_content_loader.py @ e0ec9b7ce08f349a09edb2a91446aebeaf91932a](https://github.com/alphagov/digitalmarketplace-utils/blob/e0ec9b7ce08f349a09edb2a91446aebeaf91932a/tests/test_content_loader.py)

tests/test_content_loader.py
```diff
@@ -0,0 +1,304 @@
...
+        return yaml.load(mocked_content[yaml_file])
+    return read
+
+
+class TestContentLoader(unittest.TestCase):
+
+    @mock.patch("dmutils.content_loader.ContentLoader._yaml_file_exists")
+    @mock.patch(
+        "dmutils.content_loader.ContentLoader._read_yaml_file",
+        side_effect=get_mocked_yaml_reader({
```
allait: It looks good either way, but since you have a reference to the `mocked_read_yaml_file` object in test arguments you could assign the `side_effect` there too. That way I think you can move the now repeating `@mock.patch` decorators to wrap the whole class instead of repeating them for every test method, which as far as I remember should work the same way (class decorator will decorate each test method).

quis:
@allait Your advice worked perfectly: 9a5ee9da6e3769ab2ab90623811eeab382742330

allait:
:+1:

PR #68, opened 2015-06-17
Audit api endpoints

minglis:
Added API endpoints to handle interaction with audit events:

- Fetch
Uses page, audit_type, audit_date and acknowledged flags to retrive events from data api

- Acknowledged
Allows an admin to mark this event as being OK'd in essence


minglis:
Depends on https://github.com/alphagov/digitalmarketplace-api/pull/149

PR #72, opened 2015-06-18
Bump version to 1.0.3

quis:
Includes:
- #71 make-content-loader-thread-safe

I think this is a patch, not major change because it's a bugfix in something that no apps were yet using.


robyoung:
I've pushed the 1.0.3 tag.

PR #74, opened 2015-06-19
Add .get_next_section_id() method to content loader

quis:
_Reviewing the changes in this pull request will probably be a lot easier commit-by-commit_

---

Because files are slow to load, the intended use of ContentLoader was to do this once when starting the app.

This caused thread safety issues when using ContentLoader in an object-orientated way, because you couldn't reliably store state inside an instance.

Being able to store state in an instance is nice because it makes the interface cleaner.

This commit splits content loader into two parts:
- `YAMLLoader`, which is expected to instantiated once on startup and handles reading and parsing YAML files, and caching the results
- `ContentBuilder` which is expected to be instantiated once per request and therefore could maintain state

---

So now it's possible to bring back the `.filter()` method that was refactored out yesterday.

Once you have an instance of `ContentBuilder` this filters the content to only the questions relevant to the current service.

---

So it's now possible to add a `.get_next_section_id()` method which operates on the filtered content.

When passed the ID of the current section, it will return the ID of the next section. This will be useful when we want to go through the flow of pages for the SSP.

---

New usage:
``` python
from dmutils.content_loader import YAMLLoader, ContentBuilder

yaml_loader = YAMLLoader()

def view_service():
    content = ContentBuilder("manifest.yml", "g6/", yaml_loader)
    content.filter(service_data)
    render_template("service.html", sections=content.sections)
```
---
Goes towards [97275662](https://www.pivotaltracker.com/story/show/97275662)

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -4,14 +4,14 @@ import re
 import os
 
 
-class ContentLoader(object):
+class ContentBuilder(object):
 
-    def __init__(self, manifest, content_directory):
+    def __init__(self, manifest, content_directory, yaml_loader):
 
-        section_order = self._read_yaml_file(manifest)
-
+        self.yaml_loader = yaml_loader
+        section_order = yaml_loader.read(manifest)
         self._directory = content_directory
-        self._question_cache = {}
+
         self.sections = [
             self._populate_section(s) for s in section_order
         ]
@@ -23,20 +23,7 @@ class ContentLoader(object):
                 return section
 
     def get_question(self, question):
-
-        if question not in self._question_cache:
-
-            question_file = self._directory + question + ".yml"
-
-            if not self._yaml_file_exists(question_file):
-                self._question_cache[question] = {}
-                return {}
-
-            question_content = self._read_yaml_file(question_file)
-            question_content["id"] = question
-            self._question_cache[question] = question_content
-
-        return self._question_cache[question]
+        return self.yaml_loader.read(self._directory + question + ".yml")
 
     def get_sections_filtered_by(self, service_data):
         return [
@@ -75,14 +62,6 @@ class ContentLoader(object):
 
         return None
 
-    def _yaml_file_exists(self, yaml_file):
-        return os.path.isfile(yaml_file)
-
-    def _read_yaml_file(self, yaml_file):
-        with open(yaml_file, "r") as file:
-            question_content = yaml.load(file)
-            return question_content
-
     def _populate_section(self, section):
         section["questions"] = [
             self.get_question(q) for q in section["questions"]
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -92,3 +84,17 @@ def _question_should_be_shown(self, dependencies, service_data):
             if not service_data[depends["on"]] in depends["being"]:
                 return False
         return True
+
+
+class YAMLLoader(object):
+
+    def __init__(self):
+        self._cache = []
```
allait: Should this be a dictionary?

dmutils/content_loader.py
```diff
@@ -4,49 +4,49 @@
...
 
-        for section in self.sections:
+    def get_section(self, requested_section):
+        for section in self._all_sections:
             if section["id"] == requested_section:
                 return section
+        return None
 
     def get_question(self, question):
+        return self.yaml_loader.read(self._directory + question + ".yml")
```
allait: This fails when reusing the same `YAMLLoader` instance with two content loaders: `section["questions"]` contains full question dictionaries which leads to a `TypeError: cannot concatenate 'str' and 'dict' objects`

dmutils/content_loader.py
```diff
@@ -4,49 +4,49 @@
...
 
 
-class ContentLoader(object):
-
-    def __init__(self, manifest, content_directory):
-
-        section_order = self._read_yaml_file(manifest)
+class ContentBuilder(object):
 
+    def __init__(self, manifest, content_directory, yaml_loader):
```
allait: Since all the file loading is part of `YAMLLoader` content builder doesn't really need the `manifest` and `content_directory`. I think we could move the `.get_question` method to `YAMLLoader` and change this to:
```python
yaml_loader = YAMLLoader(manifest, content_directory)
content = ContentBuilder(yaml_loader)
```

dmutils/content_loader.py
```diff
@@ -4,49 +4,49 @@
...
-        return [
-            section for section in [
-                self.get_section_filtered_by(section["id"], service_data)
-                for section in self.sections
-            ] if section is not None
-        ]
+        return None
 
-    def get_section_filtered_by(self, section_id, service_data):
+    def _get_section_filtered_by(self, section_id, service_data):
```
allait:
```python
 if len(filtered_questions):
             section["questions"] = filtered_questions
             return section
```

modifies the shared `section` dictionary in the `yaml_loader._cache` – the filtered questions will be missing for all future content builders that would use the same yaml loader

dmutils/content_loader.py
```diff
@@ -63,14 +63,6 @@ def get_section_filtered_by(self, section_id, service_data):
...
-    def _yaml_file_exists(self, yaml_file):
-        return os.path.isfile(yaml_file)
-
-    def _read_yaml_file(self, yaml_file):
-        with open(yaml_file, "r") as file:
-            question_content = yaml.load(file)
-            return question_content
-
     def _populate_section(self, section):
         section["questions"] = [
```
allait: Modifies the shared `section` dictionary from yaml loader. Should probably be a method of `YAMLLoader`.

allait:
I like the idea of separating the classes, but it seems that `YAMLLoader` is not doing enough to protect the global shared state from modification. I think we should move more "loading" logic to the `YAMLLoader` and either make it return copies (instead of originals) of the section objects it returns to the content builder or make sure the places that content builder never modifies the values owned by yaml loader.

dmutils/content_loader.py
quis: Yes, changed in 26d4715

PR #82, opened 2015-06-23
Presenters should handle 'None' content

quis:
[Previously](https://github.com/alphagov/digitalmarketplace-utils/blob/1ebfc0ce311c220e00f3a3e49e9a0caedca4f79c/dmutils/content_loader.py#L33), ContentLoader returned `{}` for question content that didn't exist.

[Now](https://github.com/alphagov/digitalmarketplace-utils/blob/3a604744c628dc8c9f944cc6f0452ad7e7506f93/dmutils/content_loader.py#L99) it returns `None`, which the Presenters choke on.

This commit adds an extra check in the Presenters for ContentLoader passing `None`.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,7 +2,7 @@ from . import logging, config, proxy_fix
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '2.0.1'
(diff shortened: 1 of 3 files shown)


allait:
Is there a test for this? Do we need one?

quis:
@allait Not sure if we need one, but can't hurt: 28b2481

PR #85, opened 2015-06-23
Rename selection question audit type

robyoung:
The previous audit types made it sound like we were updating the
questions rather than providing answers to them.


TheDoubleK:
Make your mind up :)

PR #89, opened 2015-06-23
Add `create_new_draft_service` to DataAPIClient

pcraig3:
Anticipates @TheDoubleK's [pull request in the digitalmarketplace-api](https://github.com/alphagov/digitalmarketplace-api/pull/156).

Related to two stories in Pivotal: [#97275564](https://www.pivotaltracker.com/story/show/97275564) and [#97275432](https://www.pivotaltracker.com/story/show/97275432)

(As a final note, changes are easier to read by going commit-by-commit.)

### In Sum

* Adds `create_new_draft_service` method which corresponds to similarly named method in API.
* Renamed some of the DataAPIClient methods
```python
  - create_service()       -> import_service()
  - create_draft_service() -> copy_draft_service_from_existing_service()
  - launch_draft_service() -> publish_draft_service()
```
* Changed `service_id` parameter to `draft_id` in most of the draft methods
* Changed url structures from `/services/{id}/draft` to `/draft-services/{id}`
* Grouped together relevant methods by the framework models they operate on.
Order is as follows:
Audit Events  -> Suppliers -> Users -> Services.

I don't think anyone is using draft service methods yet or the `create_service` method (which no one *should* use anyway), so I'm pretty sure this is backwards compatible.

***
#### 1. Use of the word "publish"

I changed the `launch_draft_service()` method to `publish_draft_service()` because we use "publish" in the method name in the API as well as in the URL route itself.  So now we're using "publish" everywhere.
*However*, we also have a status called `published`, and currently "publishing" a draft service -- as it currently exists in @TheDoubleK's [pull request](https://github.com/alphagov/digitalmarketplace-api/pull/156) -- will set its status to `enabled`.  So this seems like not the best name to keep.

#### 2. `updated_reason`s

Currently a few of the methods are passing `updated_reason`s back to the API, and, in the case of the draft service methods (which don't have a `reason` parameter) are just sending `"deprecated"`.

The API itself doesn't want updated_reasons.  Neither the [JSON schema](https://github.com/alphagov/digitalmarketplace-api/blob/master/json_schemas/services-update.json), the [draft](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/drafts.py) routes, nor the [service](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py) routes use them. Can I get rid of them, or are upstream frontend apps still using them?

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -7,6 +7,8 @@ class DataAPIClient(BaseAPIClient):
         self.base_url = app.config['DM_DATA_API_URL']
         self.auth_token = app.config['DM_DATA_API_AUTH_TOKEN']
 
+    # Audit Events
+
     def find_audit_events(
             self,
             audit_type=None,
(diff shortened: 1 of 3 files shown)


allait:
I don't think we should rename the `create_service` method: API view function names aren't exposed outside the API itself so it doesn't feel like a big consistency issue and `create_service` is already used by some of the scripts. If we do want to sync them, renaming the view might be easier.

Another point is naming consistency within the client class itself: we use `create_` for PUT methods and `find_` for paginated/filtered lists. While I do like `list_` more, I think having the same prefix for similar methods is more important. We could always rename all of them at the same time later if we agree that `list_...` is a better name.

pcraig3:
Hmm, I suppose you're right about consistency in our client methods being more important than what we call the views in the API, but I don't like `create_service` because we're never actually creating a service.  Or we shouldn't be thinking of it like that, really.

The only way it should be possible to create a service is by "publishing" a draft.

allait:
@pcraig3 that's a good point. I don't think the create/import distinction is that important in this case, but I don't mind keeping the new name if we fix the scripts that use the old one.

Another thing to consider is that if the method does end up being import-only, then we'll probably get rid of it once we migrate the data.

I do think we should keep the `find_` prefix for drafts though.

dmutils/apiclient/data.py
```diff
@@ -244,3 +144,133 @@ def update_user_password(self, user_id, new_password):
...
+            "/draft-services/{}".format(draft_id)
+        )
+
+    def delete_draft_service(self, draft_id, user):
+        return self._delete(
+            "/draft-services/{}".format(draft_id),
+            data={
+                "update_details": {
+                    "updated_by": user,
+                    "update_reason": "deprecated",
```
TheDoubleK: Is it worth removing the "update_reason" fields everywhere?  The API completely ignores them now.

TheDoubleK:
The `deprecated` will keep... going to merge this to get things moving.

PR #94, opened 2015-06-25
Chainable content builder

allait:
### Change ContentLoader.filter to return a new object

* Replaces YAMLLoader with ContentLoader. ContentLoader instances should
  be created for each manifest/section file on app startup.
* Moved .get_question method to ContentLoader. Return a copy of the
  question dictionary.
* ContentLoader.get_builder() will return a new builder instance with a full list of sections.
* ContentBuilder.filter now returns a new ContentBuilder instance, so filter calls can be chained:

      content_builder.filter({'lot': "SaaS"}).filter({'lot': "PaaS"})

  would return questions and sections relevant for both SaaS and PaaS.

Example use of the new ContentLoader and ContentBuilder classes:

`app/__init__.py`:

    sections_content = ContentLoader('sections_order.yml', 'app/content/g6/')

Creating a filtered ContentBuilder in a view method:

    content = section_content.get_builder().filter({'lot': "SaaS"})

Getting individual question data:

    content = section_content.get_question("questionId")

### 	Make ContentBuilder an iterator
 
Instead of `for section in content_builder.sections` it's now possible
to do `for section in content_builder`

Paired with @quis

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -5,17 +5,8 @@ import os
 
 
 class ContentBuilder(object):
-
-    def __init__(self, manifest, content_directory, yaml_loader):
-        self.yaml_loader = yaml_loader
-        section_order = yaml_loader.read(manifest)
-        self._directory = content_directory
-
-        self._all_sections = [
-            self._populate_section(s) for s in section_order
-        ]
-
-        self.sections = self._all_sections
+    def __init__(self, sections):
+        self.sections = list(sections)
 
     def get_section(self, requested_section):
         for section in self.sections:
@@ -23,27 +14,10 @@ class ContentBuilder(object):
                 return section
         return None
 
-    def get_question(self, question):
-        question_content = self.yaml_loader.read(
-            self._directory + question + ".yml"
-        )
-        question_content["id"] = question
-        return question_content
-
-    def filter(self, service_data):
-        self.sections = [
-            section for section in [
-                self._get_section_filtered_by(section["id"], service_data)
-                for section in self._all_sections
-            ] if section is not None
-        ]
-
     def get_next_section_id(self, section_id=None, only_editable=False):
-
         previous_section_is_current = section_id is None
 
         for section in self.sections:
-
             if only_editable:
                 if (
                     previous_section_is_current and
@@ -63,15 +37,16 @@ class ContentBuilder(object):
     def get_next_editable_section_id(self, section_id=None):
         return self.get_next_section_id(section_id, True)
 
-    def _get_section_from_all(self, requested_section):
-        for section in self._all_sections:
-            if section["id"] == requested_section:
-                return section
-        return None
+    def filter(self, service_data):
+        sections = filter(None, [
+            self._get_section_filtered_by(section, service_data)
+            for section in self.sections
+        ])
 
-    def _get_section_filtered_by(self, section_id, service_data):
+        return ContentBuilder(sections)
 
-        section = self._get_section_from_all(section_id)
+    def _get_section_filtered_by(self, section, service_data):
+        section = section.copy()
 
         filtered_questions = [
             question for question in section["questions"]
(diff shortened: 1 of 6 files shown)


dmutils/formats.py
```diff
@@ -1,2 +1,39 @@
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 DATE_FORMAT = "%Y-%m-%d"
+
+LOTS = [
+    {
+        'lot': 'saas',
+        'lot_case': 'SaaS',
+        'label': u'Software as a Service',
```
TheDoubleK: Why do labels need the 'u' modifier but others don't?

dmutils/content_loader.py
```diff
@@ -111,15 +75,58 @@ def _question_should_be_shown(self, dependencies, service_data):
...
+        question_content["id"] = self._make_question_id(question)
+
+        return question_content
+
+    def _make_id(self, name):
+        return inflection.underscore(
+            re.sub("\s", "_", name)
+        )
+
+    def _make_question_id(self, question):
```
robyoung: This could be a staticmethod

dmutils/content_loader.py
```diff
@@ -111,15 +75,58 @@ def _question_should_be_shown(self, dependencies, service_data):
...
+
+    def _load_question(self, question):
+        question_content = read_yaml(
+            self._directory + question + ".yml"
+        )
+        question_content["id"] = self._make_question_id(question)
+
+        return question_content
+
+    def _make_id(self, name):
```
robyoung: This could be a staticmethod

dmutils/formats.py
```diff
@@ -1,2 +1,39 @@
...
+    {
+        'lot': 'scs',
+        'lot_case': 'SCS',
+        'label': u'Specialist Cloud Services',
+    },
+]
+
+
+def lot_to_lot_case(lot_to_check):
+    lot_i_found = [lot for lot in LOTS if lot['lot'] == lot_to_check]
```
TheDoubleK: This line isn't hard to parse at all.

dmutils/content_loader.py
```diff
@@ -111,15 +75,58 @@ def _question_should_be_shown(self, dependencies, service_data):
...
+
+    def _populate_section(self, section):
+        section["id"] = self._make_id(section["name"])
+        section["questions"] = [
+            self.get_question(q) for q in section["questions"]
+        ]
+
+        return section
+
+    def _load_question(self, question):
```
robyoung: If directory were passed in this could be a staticmethod.

dmutils/content_loader.py
```diff
@@ -111,15 +75,58 @@ def _question_should_be_shown(self, dependencies, service_data):
...
+    def __init__(self, manifest, content_directory):
         self._cache = {}
+        self._directory = content_directory
+
+        manifest_sections = read_yaml(manifest)
+
+        self._questions = {
+            q: self._load_question(q)
+            for section in manifest_sections
+            for q in section["questions"]
```
robyoung: Nested comprehensions make my brain hurt. This one isn't so bad because the variable names make it fairly obvious but I did have to stop and have a think about it.

dmutils/formats.py
allait: Labels are the only ones that end up in the template and I think that caused unicode errors on py3, but I don't remember what the exact issue was.

dmutils/formats.py
```diff
@@ -1,2 +1,39 @@
...
+    if lot_i_found:
+        return lot_i_found[0]['lot_case']
+    return None
+
+
+def get_label_for_lot_param(lot_to_check):
+    lot_i_found = [lot for lot in LOTS if lot['lot'] == lot_to_check]
```
robyoung: It's also duplicated here.

tests/test_content_loader.py
```diff
@@ -17,354 +17,255 @@
...
-        return yaml.load(mocked_content[yaml_file])
-    return read
-
-
-@mock.patch("dmutils.content_loader.YAMLLoader.read")
 class TestContentBuilder(unittest.TestCase):
+    def test_content_builder_init(self):
+        content = ContentBuilder([])
+
+        self.assertEqual(content.sections, [])
```
robyoung: Elsewhere in dmutils we're using py.test. Is it worth replacing these with asserts?

PR #95, opened 2015-06-29
Remove special minimumContractPeriod handling from search apiclient

allait:
Buyer frontend has access to the questions content so it should build
the correct OR/AND value for each filter type without relying on
special cases in apiclient.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,7 +2,7 @@ from . import logging, config, proxy_fix
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '3.1.0'
+__version__ = '3.1.1'
 
 
 def init_app(
--- a/dmutils/apiclient/search.py
+++ b/dmutils/apiclient/search.py
@@ -58,20 +58,15 @@ class SearchAPIClient(BaseAPIClient):
                 raise
         return None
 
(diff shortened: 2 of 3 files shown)


TheDoubleK:
Is there a difference between "should" and "does" here?  i.e. will this break the buyer frontend until a corresponding change is made there?

dmutils/__init__.py
```diff
@@ -2,7 +2,7 @@
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '3.0.0'
+__version__ = '3.0.1'
```
quis: Doesn't feel like a patch change if there's an app relying on it…

allait:
@quis @TheDoubleK the buyer-frontend compiles the query with grouped filters on its own now (https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/85). The params that the apiclient gets from the new buyer-frontend will already include the joined `minimumContractPeriod`, so the removed code in dmutils is no longer needed, but should still work with the new buyer app. I've set the dmutils version to `3.0.0` in the buyer-frontend PR, but depending on which one gets merged first I'll either change that or we could change it later.

quis:
OK, I think that makes sense—:+1: to merge from me once you've rebased.

PR #100, opened 2015-07-20
Add find_users method and test.

pcraig3:
**[Pull request #170](https://github.com/alphagov/digitalmarketplace-api/pull/170) in the API needs to be merged before these methods can be used**

Adds a `find_users` method to list users, but enforces a `supplier_id` as a parameter.  While the API doesn't care if a `supplier_id` is given, we probably don't want frontend apps to have access to that functionality at this point.

Additionally, the query parameter for getting a user's email has changed in the API (so I've changed that in the `get_user` method), but there's no change to the interface presented by the DataAPIClient.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/98628954)

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,7 +2,7 @@ from . import logging, config, proxy_fix
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '3.2.0'
+__version__ = '3.3.0'
 
 
 def init_app(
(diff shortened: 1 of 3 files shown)


allait:
I think this has to go out before https://github.com/alphagov/digitalmarketplace-api/pull/170, but it needs to include a fix for the single-user/user-list change for `email_address` look-up response.

pcraig3:
Not sure what you mean, unless you think the `find_users` method should also take email addresses.

allait:
When `.get_user` calls `/users` with `email_address` it used to get a `{'users': {...}}` response from the API. With the changes in the API PR it will now get a `{'users': [{...}]}`. It should either transform the new format to the old one or we need to support both in the supplier app so that we can update the API without breaking the supplier app (so for some time supplier app should be able to handle both API version responses, then we update the API, then we can remove the `{'users': {...}}` support from dmutils and the supplier app).

allait:
This looks good to go :+1:

PR #101, opened 2015-07-24
Update users

minglis:
Added:

- Handler for 404 on get supplier - so client gets a None not an exeption in case of missing resource

- Update user endpoint so can updated *active* and *locked* aspects of a service

- Added an active flag to the user


robyoung:
:+1:

PR #102, opened 2015-07-27
Add asset fingerprinting

quis:
## 1. 	Extract generation of asset fingerprints from apps

This util generates a hash of file contents. This enables us to serve assets from different URLs if they are changed, which means they wont be loaded from browser cache.

This is already [implemented in the buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/111) but it seems sensible to share it between the apps.

## 2. Generate URLs within the util

@robyoung pointed out that we need to generate URLs for about 6 asset files now (rather than two), which would mean quite a lot of repetition in each apps templates and `Config.py`s.

This commit makes 'asset fingerprinter' into a class, and gives it a method to return the path and querystring for an asset, rather than just the hash.

Usage

**`Config.py`:**
```
config.asset_fingerprinter = AssetFingerprinter()
```

**`_base_template.html`:**
```
{{ asset_fingerprinter.get_url('stylesheets/application.css') }}
```

**Result:**
```
static/stylesheets/application.css?418e6f4a6cdf1142e45c072ed3e1c90a
```
"""

Addresses [98936730](https://www.pivotaltracker.com/story/show/98936730)

### DIFF
--- a/dmutils/asset_fingerprint.py
+++ b/dmutils/asset_fingerprint.py
@@ -1,20 +1,45 @@
 import hashlib
 
 
-def get_asset_fingerprint(asset_file_path):
+class AssetFingerprinter():
     """
         Get a unique hash for an asset file, so that it doesn't stay cached
         when it changes
 
         Usage:
-        get('app/static/stylesheets/application.css')
+
+            Config.py:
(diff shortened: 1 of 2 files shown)


allait:
Looks good :+1: :shipit:

PR #104, opened 2015-07-27
Add display date formats and template filters

pcraig3:
Both in the admin app and the supplier app we need to display dates, so it was just a matter of time before the date display logic ended up here.

After this update, all apps using dmutils will be able to use `timeformat`, `dateformat`, and `datetimeformat` in their templates.


allait:
:+1:

PR #106, opened 2015-07-27
Remove 's'

quis:
…and bump version to 3.4.1


allait:
There's a version conflict, but other than that :+1:

PR #107, opened 2015-07-27
Handle passing Nones in to datetime filters

pcraig3:
If a supplier hasn't logged in, their `loggedInAt` value is `null`, which would break our jinja filter methods.
New update returns a user set default value or `None`.


allait:
:+1:

PR #109, opened 2015-07-27
Return all (in)active users from the API.

pcraig3:
Added optional `active` parameter to `find_users` in the DataAPIClient.
Now we can choose to return either all active or all inactive users belonging to a supplier.
Or we can just return them all as before, if we're not particularly bothered.


pcraig3:
Decided to remove the `active` filter.  We can have this logic in the frontend apps.

PR #110, opened 2015-07-30
Make User model better reflect auth state

robyoung:
This change makes the Flask-Login interface methods on the User model better reflect the intended authentication state of a user account. After this change the following statements hold true.

- A Flask-Login User cannot be considered active if the account is locked.
- A Flask-Login User cannot be considered authenticated if it is not active.

### DIFF
--- a/dmutils/user.py
+++ b/dmutils/user.py
@@ -58,3 +58,13 @@ class User():
             locked=user['locked'],
             active=user['active']
         )
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


allait:
I'm worried about future Flask-Login update breaking this if it switches to properties instead of methods. Is there a test we could write to check that Flask-Login returns the right status given an locked/deactivated user model?

allait:
:+1:

PR #111, opened 2015-07-30
Adding email integration into utils

minglis:
Pull request to support sending of emails in utils

This migrates the mandrill code out of the supplier app and into the utils classes. This supports pivotal stories:

https://www.pivotaltracker.com/n/projects/997836/stories/98628914
and
https://www.pivotaltracker.com/n/projects/997836/stories/100192466

Moves all mandrill code into an email.py class with associated tests.

Also adds some extensions to user functionality:

- adds name to response
- allows updating of supplier_id and role

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -173,7 +173,12 @@ class DataAPIClient(BaseAPIClient):
                         user_id, e.status_code)
             return False
 
-    def update_user(self, user_id, locked=None, active=None):
+    def update_user(self,
+                    user_id,
+                    locked=None,
+                    active=None,
+                    role=None,
+                    supplier_id=None):
         fields = {}
         if locked is not None:
             fields.update({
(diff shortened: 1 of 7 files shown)


dmutils/apiclient/data.py
```diff
@@ -173,7 +173,12 @@ def update_user_password(self, user_id, new_password):
                         user_id, e.status_code)
             return False
 
-    def update_user(self, user_id, locked=None, active=None):
+    def update_user(self,
+                    user_id,
+                    locked=None,
+                    active=None,
+                    role=None,
+                    supplier_id=None):
```
robyoung: What's the idea with these vertical argument lists? We use them sometimes but not others. I find them quite hard to read.

minglis: personal preference,often we go over pep8 line length, so we get some on one line and some on the next. I think thats espcially hard to read. Always wrote them like this in JVM languages when the param list got more than a couple.

I find them easier to read, and do it when I'm meddling in the code.....

robyoung:
:+1:

PR #113, opened 2015-08-03
Extract attribute helper from buyer frontend

quis:
The idea is that all frontend apps will use this to transform data from the API before sending it to the toolkit templates.

### DIFF
--- a/dmutils/service_attribute.py
+++ b/dmutils/service_attribute.py
@@ -1,0 +1,44 @@
+class Attribute(object):
+    """Wrapper to handle accessing an attribute in service_data"""
+
+    def __init__(
+        self,
+        value,
+        question_type,
+        label='',
+        optional=False,
+    ):
+        self.label = label
+        self.answer_required = False
+        if value in ['', [], None]:
+            self.value = ''
+            self.type = 'text'
+            self.assurance = False
+            if not optional:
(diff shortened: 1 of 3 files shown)


dmutils/service_attribute.py
```diff
@@ -0,0 +1,44 @@
+class Attribute(object):
+    """Wrapper to handle accessing an attribute in service_data"""
+
+    def __init__(
+        self,
+        value,
+        question_type,
+        label='',
+        optional=False,
+    ):
```
robyoung: ):

tests/test_service_attribute.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+
+class TestAttribute(unittest.TestCase):
+
+    def test_Attribute_works_if_the_key_is_in_service_data(self):
+        try:
+            attribute = Attribute('supportAvailability', 'text')
+        except KeyError:
+            self.fail("'supportAvailability' key should be in service data")
```
robyoung: I'm struggling to see what this test tells us.

quis: @robyoung Good point. It's a hangover. Removed in 4bd9829.

robyoung:
:+1:

PR #114, opened 2015-08-03
Add function to format a price string

robyoung:
The priceString field is slowly going to be removed and replaced with rendering the price from it's constituent parts where needed.


allait:
:+1:

PR #115, opened 2015-08-03
Send draft update page_questions outside service data

allait:
Adds additional argument to draft update method to pass
`page_questions`.

This way, `page_questions` is more explicit as a meta field and is
not mixed with the service data.

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -246,16 +246,19 @@ class DataAPIClient(BaseAPIClient):
                 }
             })
 
-    def update_draft_service(self, draft_id, service, user):
-        return self._post(
-            "/draft-services/{}".format(draft_id),
-            data={
-                "update_details": {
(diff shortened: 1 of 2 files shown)


dmutils/apiclient/data.py
```diff
@@ -246,16 +246,19 @@ def copy_draft_service(self, draft_id, user):
...
+                "updated_by": user,
+                "update_reason": "deprecated",
+            },
+            "services": service,
+        }
+
+        if page_questions is not None:
+            data['page_questions'] = page_questions
+
+        return self._post("/draft-services/{}".format(draft_id), data=data)
```
robyoung: What situations do we have where `page_questions` shouldn't be provided for a draft update?

allait: When you're updating a complete draft? You could provide them, but there should be no need to do this.

robyoung:
:+1:

PR #116, opened 2015-08-03
Move file validation functions to dmutils.documents

allait:
## Move file validation functions to dmutils.documents
Extracts file validation checks from validator class to standalone
functions. Validate methods are preserved and call added functions.

Adds `validate_documents` and `filter_empty_files` methods used to
check `request.files` for any errors before uploading documents to
S3.

## Use framework slug as the S3 documents folder
Separating documents for each framework would allow us to control
access using nginx - e.g. redirecting /g-cloud-7 asset URLs to a
static page until G-Cloud 7 goes live.

All existing documents are still in the '/documents' folder, but
future document uploads for G5 and G6 will end up in separate
folders.

## Add `public` argument to document_upload
Allows setting the S3 file object ACL:
* `public=True` sets ACL to `"public-read"`
* `public=False` sets ACL to `"private"`

### DIFF
--- a/dmutils/documents.py
+++ b/dmutils/documents.py
@@ -62,6 +62,7 @@ def upload_document(uploader, documents_url, service, field, file_contents):
 
     """
     file_path = generate_file_name(
+        service['frameworkSlug'],
         service['supplierId'],
         service['id'],
         field,
@@ -101,7 +102,7 @@ def file_is_open_document_format(file_object):
     ]
 
 
-def generate_file_name(supplier_id, service_id, field, filename, suffix=None):
+def generate_file_name(framework_slug, supplier_id, service_id, field, filename, suffix=None):
     if suffix is None:
         suffix = default_file_suffix()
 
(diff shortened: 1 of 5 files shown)


dmutils/documents.py
```diff
@@ -0,0 +1,136 @@
...
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
+
+from .s3 import S3ResponseError
+
+
+def filter_empty_files(files):
+    """Removes any empty files from the list.
```
robyoung: [PEP257](https://www.python.org/dev/peps/pep-0257/) says this should be written as "Remove any empty files from the list." :trollface:

dmutils/documents.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def file_is_not_empty(file_contents):
+    not_empty = len(file_contents.read(1)) > 0
+    file_contents.seek(0)
+    return not_empty
+
+
+def file_is_less_than_5mb(file_contents):
+    size_limit = 5400000
```
robyoung: What's `5400000`? It's not 5Mb.

robyoung:
:+1:

PR #117, opened 2015-08-04
Remove use of base_url from DataAPIClient

robyoung:
This is no longer needed as `_get` in the `BaseAPIClient` prepends it. It hasn't been causing any problems because `urlparse.urljoin` is sensible.


allait:
:+1:

PR #120, opened 2015-08-04
Improve find users

robyoung:
## Add query parameters to find_users method

The API endpoint allows users to be listed by supplier_id, active and page. Being able to iterate through all active users is needed for generating an email list.

## Add page iteration to the DataAPIClient

This allows any 'find' method to be called in a way that automatically iterates through pages and retries failed requests.

Any method that starts with 'find_' and ends with '_iter' it will be called with an exponential backoff. From the response the records will be yielded and then the 'next' URL will be fetched and yielded until there are no pages.

This is most often needed for scripts that export or migrate large numbers of records where occasional requests may fail.

## Example

The following will iterate through all pages of active users, recovering from 503s if they happen.

```python
client = DataAPIClient('http://host', 'token')

for user in client.find_users_iter(active=True):
  print(user)
```

### DIFF
--- a/dmutils/apiclient/base.py
+++ b/dmutils/apiclient/base.py
@@ -16,6 +16,31 @@ from .errors import APIError, HTTPError, HTTP503Error, InvalidResponse
 logger = logging.getLogger(__name__)
 
 
+def make_iter_method(method_name, model_name, url_path):
+    """Make an iterator method from a find method
+
+    :param method_name: The name of the find method to decorate
+    :param model_name: The name of the model as it appears in the JSON response
+    :param url_path: The URL path for the API
+    """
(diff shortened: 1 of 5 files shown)


dmutils/apiclient/base.py
```diff
@@ -83,3 +84,42 @@ def get_status(self):
...
+        model_name = self._get_model_name(method_name)
+        method = getattr(self, method_name)
+
+        backoff_decorator = backoff.on_exception(backoff.expo,
+                                                 HTTP503Error,
+                                                 max_tries=5)
+
+        def iter_method(*args, **kwargs):
+            result = backoff_decorator(method)(*args, **kwargs)
+            print(result)
```
allait: Probably don't need this anymore

allait:
:+1:

PR #124, opened 2015-08-05
Add S3.get_signed_url method

allait:
Creates a temporary link to the private S3 document. Link expires
in `expires_at` seconds (defaults to 30 seconds).

Can be used for documents that require authorization and should not
be shared publicly (eg service submission documents).

### DIFF
--- a/dmutils/s3.py
+++ b/dmutils/s3.py
@@ -27,6 +27,21 @@ class S3(object):
         key.set_acl(acl)
         return key
 
+    def get_signed_url(self, path, expires_in=30):
+        """Create a signed S3 document URL
+
+        :param path: S3 object path within the bucket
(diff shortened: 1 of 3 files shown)


dmutils/s3.py
```diff
@@ -27,6 +27,22 @@ def save(self, path, file, acl='public-read', move_prefix=None):
...
+                           for, in seconds
+
+        :return: signed URL or ``None`` if object was not found
+
+        """
+
+        key = self.bucket.get_key(path)
+        if key is None:
+            return key
+        return key.generate_url(expires_in)
```
robyoung: Couldn't this just be the following?
```python
if not is not None:
    return key.generate_url(expires_in)
```

robyoung:
:+1:

PR #125, opened 2015-08-06
New endpoint for creating suppliers

minglis:
Breaking change to supplier interface

- create_supplier renamed to import_supplier
- new create_supplier endpoing

import_supplier is the PUT endpoint used in migrations
- expects a supplier_id

create_supplier is the POST endpoing for new suppliers
- does not expect a supplier_id


robyoung:
:+1: with conditions
- Depends on https://github.com/alphagov/digitalmarketplace-api/pull/203
- Should be a major version bump, hello `5.0.0`

PR #126, opened 2015-08-06
Get form data from content builder

robyoung:
This moves the form parsing and filtering logic implemented in the supplier frontend into the `ContentBuilder` so that it can be used on other content manifests and in other frontends. I have left a couple of TODOs in around allowing questions with multiple fields (ie. pricing). I don't want to implement it right now but I would like to when we have done pre-pag.

This will require changes in the supplier frontend to make use of it.

**Before this change:**
```python
posted_data = get_formatted_section_data(section)
```

**After this change:**
```python
posted_data = content.get_section_data(section, request.form)
```

## Pull out a ContentSection class

The `ContentBuilder.get_section_data` method made more sense on a specific section object as we're already getting hold of the section.

**Rather than:**
```python
section = content.get_section(section_id)
if not section:
    abort(404)
section_data = content.get_section_data(section_id, request.form)
```

**We do:**
```python
section = content.get_section(section_id)
if not section:
    abort(404)
section_data = section.get_data(request.form)
```

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -114,7 +114,7 @@ class ContentSection(object):
     @classmethod
     def create(cls, section):
         if isinstance(section, ContentSection):
-            return section
+            return section.copy()
         else:
             return ContentSection(
                 id=section['id'],
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -228,10 +228,7 @@ class ContentLoader(object):
         ]
 
     def get_question(self, question):
-        q = self._questions.get(question, {}).copy()
-        if q:
-            q['id'] = question
-        return q
+        return self._questions.get(question, {}).copy()
 
     def get_builder(self):
         return ContentBuilder(self._sections)
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -10,11 +10,13 @@ from .config import convert_to_boolean, convert_to_number
 class ContentBuilder(object):
     """An ordered set of sections each made up of one or more questions.
 
-    Usage::
+    Example::
+        # Get hold of a section
+        content = ContentBuilder(sections)
+        section = content.get_section(section_id)
 
-        >>> content = ContentBuilder(sections)
-        >>> content.get_section_data(section_id, form_data)
-        {'field': 'value', 'field2': 'value2'}
+        # Extract data from form data
+        section.get_data(form_data)
     """
     def __init__(self, sections):
         self.sections = [ContentSection(section) for section in sections]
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -1,32 +1,54 @@
...
 
+from .config import convert_to_boolean, convert_to_number
+
 
 class ContentBuilder(object):
+    """An ordered set of sections each made up of one or more questions.
+
+    Usage::
+
+        >>> content = ContentBuilder(sections)
```
allait: Running `py.test` with `--doctest-modules` picks this up as a doctest which fails. Making this work as a test probably wouldn' t make a useful usage example, so I think we should just drop the `>>>` and explain the return value separately.

robyoung: Ah, bums. Thanks though, this also highlighted I was referencing the wrong method in some of the docstrings.

quis:
Had a proper read through this now. :+1:

dmutils/content_loader.py
```diff
@@ -90,7 +217,10 @@ def __init__(self, manifest, content_directory):
         ]
 
     def get_question(self, question):
-        return self._questions.get(question, {}).copy()
+        q = self._questions.get(question, {}).copy()
+        if q:
+            q['id'] = question
```
allait: I think this would benefit from an explanation of what this bit is for.
All items in `_questions` come from `_load_questions` which adds the `'id'`, so does this have something to do with `serviceTypes*`?

dmutils/content_loader.py
```diff
@@ -74,6 +103,115 @@ def _question_should_be_shown(self, dependencies, service_data):
...
+            question = section.get_question(question_id)
+            if question:
+                return question
+
+
+class ContentSection(object):
+    @classmethod
+    def create(cls, section):
+        if isinstance(section, ContentSection):
+            return section
```
allait: This should make a copy as section `.questions` might be modified by additional filters. I think we can rely on `__getitem__` for now to reuse the code from the `else:` branch

dmutils/content_loader.py
robyoung: I think this may have been an example of the test data not matching the real data. I have since added `id`s to the test data so I think this can be removed.

dmutils/content_loader.py
robyoung: That's explicitly handled by `_get_section_filtered_by` [here](https://github.com/alphagov/digitalmarketplace-utils/pull/126/files#diff-346b98f8c34fcc00d15a2cc1629cfb6aL52)

robyoung: Although, I suppose it doesn't hurt and it does make the interface a bit more obvious.

allait:
:+1:

PR #127, opened 2015-08-07
Fixed incorrect logging on send email

minglis:
logged stated "password reset"

- now uses type tags to record actual email sent.


robyoung:
:+1:

PR #128, opened 2015-08-07
Add method to ContentSection to get field names

robyoung:
This moves `get_section_questions` and `get_raw_section_questions` from the supplier frontend to avoid exposing the `_is_pricing_type` method externally. It also brings in a distinction between question IDs, which are used to identify questions in the SSP content. And field names, which are the keys in the API payload that is returned by `get_data`.


quis:
:+1:

PR #131, opened 2015-08-10
Monotonic timing and timing API requests

robyoung:
## Use monotonic time for metric time

Non-monotonic time can jump back or forward. AWS tends to be quite good at this, slewing time changes slowly. However, using a monotonic time function is more robust.

## Time all API requests.

This gives us the ability to see how long requests are taking from the perspective of the caller.

### DIFF
--- a/dmutils/apiclient/base.py
+++ b/dmutils/apiclient/base.py
@@ -9,6 +9,7 @@ except ImportError:
 import requests
 from flask import has_request_context, request, current_app
 import backoff
+from monotonic import monotonic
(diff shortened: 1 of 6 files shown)


minglis:
👍

dmutils/apiclient/base.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+import datetime
```
allait: Doesn't seem to be used

PR #132, opened 2015-08-12
Complete draft service audit type and API client method

allait:
### Add "complete_draft_service" audit type
Used when the service submission draft is marked as complete

### Add complete_draft_service apiclient method

Sending this before the API endpoint because the API needs the new audit type


robyoung:
:+1:

PR #133, opened 2015-08-12
Duns number added to fetch suppliers endpoint

minglis:
client updates to support:

https://github.com/alphagov/digitalmarketplace-api/pull/206

Adds duns number as a param on fetching suppliers


robyoung:
:+1: If this goes in before https://github.com/alphagov/digitalmarketplace-utils/pull/134 it doesn't need a version bump.

PR #134, opened 2015-08-12
Create audit events

robyoung:
Add object type and id to find audit events. Add create_audit_event. Add audit type for register_framework_interest, a new type for registering when a supplier has shown interest in a framework.

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -48,6 +48,21 @@ class DataAPIClient(BaseAPIClient):
                 }
             })
 
+    def create_audit_event(self, audit_type, user, data, object_type=None, object_id=None):
+        payload = {
+            "type": audit_type,
+            "user": user,
+            "data": data,
+        }
+        if object_type is not None:
+            payload['objectType'] = object_type
+        if object_id is not None:
+            payload['objectId'] = object_id
+
+        return self._post(
+            '/audit-events',
+            data={'auditEvents': payload})
+
     # Suppliers
 
     def find_suppliers(self, prefix=None, page=None, framework=None):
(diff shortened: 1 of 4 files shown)


dmutils/apiclient/data.py
```diff
@@ -42,6 +48,21 @@ def acknowledge_audit_event(self, audit_event_id, user):
                 }
             })
 
+    def create_audit_event(self, audit_type, user, data, object_type=None, object_id=None):
```
minglis: The object type/id is slightly odd in that these are DB model objects normally. I don't know we'd know from an API point of view? Wondering whether these can be derived from audit-type in some way? That is if I make a "register-interest" audit type, those are always associcated with a supplier, and what the API takes is the ID of whatever object the audit code will associate  the event with.

What do you think? Is it not scalable? 

I guess the value here is a string, and I'd know what object I want to associate the event with, so I have control?

robyoung: I've explained it a bit in the API pull request (https://github.com/alphagov/digitalmarketplace-api/pull/208). This is the object type as it appears in the URL and the JSON payload. So, for `Service` it is `services` which keeps it away from the DB implementation.

I think that inferring them from the audit type could end up being a bit magical. In order to know what's happening in the client you have to know what the server is going to do behind the scenes.

minglis: Makes sense. Was worried about knowing what an object type may be when I want to create the audit event.

For example, if I was to audit sending an invitation email? I guess that would be the supplier - ie the invite relates to the supplier, but has a property of the email address that it was sent too. And I have the supplier so I can build the token, so that all works out.

I think this is OK. I guess not all audit events will map to domain objects cleanly, but we can think about each one. I suppose it's what I would query on.

👍

PR #136, opened 2015-08-14
[#100210316] Add invite user audit type

robyoung:
https://www.pivotaltracker.com/story/show/100210316

All invitations need to be audited:

- Who sent the invite
- When it was sent
- To whom
- By which supplier


allait:
:+1:

PR #137, opened 2015-08-17
[#100214054] Return a list of file keys from an s3 bucket.

pcraig3:
Returns a list of all file keys in an S3 bucket ordered by date, optionally filtered by files whose names contain a prefix and/or don't contain a delimiter.

For each key, we return a dictionary containing its:
- path
- filename
- extension
- last_modified date
- size

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/100214054)


robyoung:
:+1:

PR #139, opened 2015-08-18
Force AuditTypes instances to be used as arugments

robyoung:
For the `find_audit_events` and `create_audit_event` methods, force instances of `AuditTypes` be used as an argument.

Also add new audit types for clarification questions.


TheDoubleK:
:+1:

PR #140, opened 2015-08-19
Allow audit events to be created with no user

robyoung:
This is allowed by the API and needed by the supplier frontend.


TheDoubleK:
:+1:

PR #141, opened 2015-08-21
Update date format and to timezones

robyoung:
The date format in the designs is Tuesday 12 December 2012.

## Move date/time format filters into formats module
So that they can be more easily tested and moves them closer to the formats they wrap.

## Always convert datetimes to Europe/London timezone
As all our internal date times are UTC, if a date time has no time zone attached assume it to be UTC.


robyoung:
Closing, I'm going to put the timezone work in here as well.

allait:
:+1:

PR #142, opened 2015-08-21
Add role to the User class

allait:
#### Add User.has_role method
Returns True if user.role matches the argument


robyoung:
:+1:

PR #144, opened 2015-08-24
Bugfix: Correctly add template filters

robyoung:
The `template_filter` method can only be used as a decorator. To add a
filter function directly the `add_template_filter` method should be
used.


allait:
:+1:

PR #145, opened 2015-08-24
Add `status` field to declaration responses

TheDoubleK:
This will be either `started` or `complete`, depending on whether all declaration questions have been answered.

This will need to go in (along with a following version number bump), once this PR is updated: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/206


TheDoubleK:
Hmm... maybe not.

PR #147, opened 2015-08-25
Add snapshot_framework_stats audit event type

allait:
After some discussion we've decided that the easiest way to store
framework stats is to save them as audit events in the API.

This adds a new audit type to use for these events.

Since audit events have a creation date we can store the framework
stats response as audit event data without any changes. The event
should be associated with the related framework object.


robyoung:
:+1:

PR #148, opened 2015-08-25
Unformat data in the content loader

robyoung:
This moves the unformat_section_data function from the supplier frontend
into the ContentSection class.

See:
https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/d7e89ddc0abdc0d8e0e9d72035c1311ef948ad37/app/main/helpers/services.py#L83


allait:
:+1:

PR #149, opened 2015-08-26
Add DISPLAY_DAY_FORMAT

pcraig3:
Adding a day format like `26 August` to be used in the frameworks dashboard.  It's shorter than `Wednesday 26 August 2015` and in a lot of cases we don't gain anything by having the extra information.


quis:
:+1:

PR #150, opened 2015-08-26
Empty max price

allait:
#### Don't filter empty values from price fields
Allows removing optional priceMax value by sending `{"priceMax": ""}`
update to the API

Related to https://github.com/alphagov/digitalmarketplace-api/pull/224, but can be merged in any order (as long as both are merged)


robyoung:
:+1:

PR #151, opened 2015-08-27
Add JSON application logs

robyoung:
This adds a JSON log file when not in development. CloudWatch logs
supports much more sophisticated searching over JSON logs than it does
over plain text logs [1]. Most notably compound conditions (AND and OR).

The JSON log file path is the same as the line by line one with a
.json appended, as it seems simpler than having to explicitly
configure a separate JSON log path.

This uses python-json-logger [2]. I am pulling the v0.1.3 tag directly
from git because there is no v0.1.3 package file in PyPi. We need v0.1.3
for logging excecption track traces. There is an issue open for this
[3].

Some of the keys are rewritten to match the field names used in other
logs such as the apache and nginx access logs.

[1] http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/FilterAndPatternSyntax.html
[2] https://github.com/madzak/python-json-logger
[3] madzak/python-json-logger#30

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '6.5.2'
+__version__ = '6.6.0'
 
 
 def init_app(
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -2,10 +2,13 @@ from __future__ import absolute_import
 import logging
 import uuid
 import sys
+from itertools import product
 
 from flask import request, current_app
 from flask.wrappers import Request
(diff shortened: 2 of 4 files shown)


TheDoubleK:
Just wondering.... why do we want JSON logs? Will these also be browsable in the CloudWatch logs area?

robyoung:
@TheDoubleK yes, they're browsable in CloudWatch logs. We have some there already for Nginx and Apache access logs. The main benefit is that you can do much more sophisticated filtering with the JSON logs. See http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/FilterAndPatternSyntax.html

That's a good point though, I'll add a bit more info on why this should be useful in the commit message.

TheDoubleK:
Looks good to me :+1:

PR #152, opened 2015-08-27
Add dayformat filter, bump version

pcraig3:
##### Add dayformat template filter
Forgot to do this earlier.  Gives the templates a 'DD Month' display date filter (for example '27 August').

and version bump to 6.5.2


allait:
:+1:

PR #153, opened 2015-08-27
Add `max_age` parameter to `decode_token` method

TheDoubleK:
We want to have different ages for different types of token (invite vs password-reset).  We can do that if we add this parameter.

### DIFF
--- a/dmutils/email.py
+++ b/dmutils/email.py
@@ -57,12 +57,12 @@ def generate_token(data, secret_key, salt):
     return ts.dumps(data, salt=salt)
 
 
-def decode_token(token, secret_key, salt):
+def decode_token(token, secret_key, salt, max_age=86400):
     ts = URLSafeTimedSerializer(secret_key)
     decoded, timestamp = ts.loads(
         token,
         salt=salt,
-        max_age=86400,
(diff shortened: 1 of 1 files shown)


dmutils/email.py
```diff
@@ -57,12 +57,12 @@ def generate_token(data, secret_key, salt):
     return ts.dumps(data, salt=salt)
 
 
-def decode_token(token, secret_key, salt):
+def decode_token(token, secret_key, salt, max_age=86400):
```
quis: Could you name this `max_age_in_seconds`, if the unit is seconds?

Or call the function with a sensible unit, eg `decode_token(max_age_in_hours=48)` and calculate the seconds inside the function?

robyoung: I would have said max-age as a number of seconds is pretty common on the web (HTTP cache headers, cookie lifetimes). I'd rather it stays as seconds as it's then more instantly comparable to these other numbers.

TheDoubleK:
Yes - better to make the units explicit.  I think sticking with seconds as the unit is good (even if it means using some big numbers).

minglis:
👍

PR #155, opened 2015-09-01
Fingerprint unicode files without error

quis:
When running in Python 2, the asset fingerprinter throws an exception if given a file with unicode contents.

The only way I found to replicate this error in a test was to read a file, rather than mocking. The one file that a) can reliably expect to exist and b) makes most sense to have an arbitrary unicode character in it is the test itself.

Fixing the error meant using the `codecs` module from the standard library to to match Python 3's unicode-by-default file reading.


allait:
:+1:

PR #158, opened 2015-09-04
Include orphaned assurances

TheDoubleK:
Part of the fix for this bug: https://www.pivotaltracker.com/story/show/102562916

When getting form data any assurance responses that do not have their associated question also included in the form data are lost by the content loader. This means that responses cannot be re-populated on the page in the case of a page validation error.

This adds a check for these orphaned assurance responses and includes them in the extracted form data.

The changes to the API here: https://github.com/alphagov/digitalmarketplace-api/pull/238 will prevent any assurances being saved without a corresponding value. So that should be deployed before this new Utils version is used.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '6.8.0'
+__version__ = '6.8.1'
 
 
 def init_app(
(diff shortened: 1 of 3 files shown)


quis:
Would be good to add a couple of lines to the tests here to cover this: https://github.com/alphagov/digitalmarketplace-utils/blob/include-orphaned-assurances/tests/test_content_loader.py#L438-L475

TheDoubleK:
Added a test that an orphaned assurance is returned by the content loader.

quis:
:+1:

PR #159, opened 2015-09-08
[#102845416] Fix 500s for bad service ids

TheDoubleK:
Should fix this bug: https://www.pivotaltracker.com/story/show/102845416

If parameters passed in URLs contain unicode characters then the `.format` method on raw strings will fail in Python 2. By specifying unicode literals we can avoid this possibility.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '6.8.1'
+__version__ = '6.8.2'
 
 
 def init_app(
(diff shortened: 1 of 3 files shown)


quis:
Is it worth abstracting-out the strings into a dictionary and then testing that the dictionary only contains unicode strings? Or is that over-engineering it?

My only worry is that it's pretty trivial to reintroduce this bug when writing a new method.

TheDoubleK:
I can see your point, but it would be pretty ugly... it's nice that the individual API Client methods only know about and are responsible for the API endpoint that they interact with. To have them pick a URL out of a shared dictionary wouldn't be as nice.

TheDoubleK:
I'll do it if you insist, but I'm not sure it's necessary.

quis:
Yeah, it's a trade off. I think your original approach is the most sensible. :+1:

PR #160, opened 2015-09-09
Add string formatting to logging

robyoung:
Allow the log message to be passed in as a format string that is formatted with the log record. This allows additional parameters to be passed in once and used for both JSON log fields and the main log message. This has been implemented for both the JSON logs and the plain text logs.

This gives an overview of the Python logging architecture
https://docs.python.org/2/howto/logging.html#logging-flow

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 import logging
 import uuid
 import sys
+import re
 from itertools import product
 
 from flask import request, current_app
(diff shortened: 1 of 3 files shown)


dmutils/logging.py
```diff
@@ -120,6 +123,25 @@ def filter(self, record):
         return record
 
 
+class CustomLogFormatter(logging.Formatter):
+    """Accepts a format string for the message and formats it with the extra fields"""
+
+    FORMAT_STRING_FIELDS_PATTERN = re.compile(r'\((.+?)\)', re.IGNORECASE)
```
TheDoubleK: I'm not sure this needs the extra brackets - i.e. I think `\(.+?\)` will do the same thing?

Do we have to use a regex here - having a single string list of things in brackets seems weird, but is that a constraint of the Formatter?

TheDoubleK:
After discussion :+1:

PR #161, opened 2015-09-10
Unicode literals instead of u prefix

TheDoubleK:
As suggested by @robyoung here: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/157


robyoung:
:+1:

quis:
This is much nicer.

PR #164, opened 2015-09-21
Log stack trace for failed logging

robyoung:
## Log stack traces for logging format error messages
When the logging message formatting fails it logs an error. This change uses the `exception` method to add a stack trace to the error so that we can more easily trace the root cause.

## Move hash_email function into dmutils
This was originally implemented in the supplier frontend to hash emails before putting them in logs to avoid logging raw email addresses but still retain the ability to search for them.

## 	Use formatted log messages for email
This was causing errors with the new logging style because `result` is a JSON object. This change pulls out the useful information and formats it into the message directly.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -137,8 +137,8 @@ class CustomLogFormatter(logging.Formatter):
         record = self.add_fields(record)
         try:
             record.msg = record.msg.format(**record.__dict__)
-        except KeyError:
-            logger.exception("failed to format log message")
+        except KeyError as e:
+            logger.exception("failed to format log message: {} not found".format(e))
         return super(CustomLogFormatter, self).format(record)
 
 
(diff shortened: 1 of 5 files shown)


dmutils/logging.py
```diff
@@ -138,7 +138,7 @@ def format(self, record):
         try:
             record.msg = record.msg.format(**record.__dict__)
         except KeyError:
-            logger.error("failed to format log message")
+            logger.exception("failed to format log message")
```
TheDoubleK: Is there any information we can get from the KeyError to add to the message here - e.g. what was the key that broke?

robyoung: We sure can. I've added a commit.

allait: Wouldn't `logger.exception` add a traceback with the key that caused it anyway?

robyoung: Yes, but it wouldn't add it to the message so it would be a little harder to spot immediately.

TheDoubleK:
:+1:

PR #166, opened 2015-09-24
[#103779820] Fix user-update auditing

TheDoubleK:
Part of this bug fix: https://www.pivotaltracker.com/story/show/103779820

Currently our auditing of user updates is broken because we only audit the user who was changed, not who made the change.  We need both to have a proper record of changes that have been made.

This commit also removes the long-deprecated `update reason` parameter from update details.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
-__version__ = '6.11.0'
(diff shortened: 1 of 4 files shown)


quis:
Would be good to have a test that mocks `current_user` and ensures that it's added to the update.

TheDoubleK:
Changed it to have the updater passed in as a parameter.  Seems better this way - no need for the API client to know about the current session - that should be down to the frontend apps to manage.

quis:
That's a better solution anyway—more decoupled. :+1:

PR #167, opened 2015-09-24
Delete presenters.py

quis:
This is only used in the admin app.

When the admin app is changed to use [the summary table pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/summary-table.html) from the toolkit, this will no longer be needed.

--

Version number assumes that this will be merged after https://github.com/alphagov/digitalmarketplace-utils/pull/166


robyoung:
:+1:

PR #168, opened 2015-09-28
Add template filter for chunking the service ID

robyoung:
This adds a template filter for chunking the service ID for display. If a service ID is all numeric (ie. not G-Cloud 5) it will be split 4-digit chunks for display.

Currently this chunking is done in a couple of different places; in the buyer frontend it is done with service_presenters and in the admin app dmutils/presenters.

This does create a dependency between the frontend toolkit and dmutils but I can't see a better way around it. Any ideas would be gratefully received.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -47,7 +47,4 @@ def init_app(
         response.headers['X-Frame-Options'] = 'DENY'
         return response
 
-    application.add_template_filter(formats.timeformat)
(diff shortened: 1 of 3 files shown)


allait:
Do we still need this?

quis:
Closing this now that https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/166 is merged

PR #169, opened 2015-09-28
[#103403706] Set Request ID header on error responses

allait:
[#103403706](https://www.pivotaltracker.com/story/show/103403706)

Replaces @after_request header setter with a WSGI middleware.
Unlike after_request, middleware is run on error responses, even if
they're caused by an uncaught exception.

Middleware is not executed when DEBUG is set to True, so development
servers won't have Request-ID headers set by default.

Request-ID related code, including the custom request class is moved
to the request_id module to remove implicit `app.config.setdefault`
dependency between logging and request_id modules.


robyoung:
:+1:

PR #170, opened 2015-09-29
𝐔𝐑𝐆𝐄𝐍𝐓: Fix update password

quis:
The API has changed to require who is updating a user: alphagov/digitalmarketplace-api@7f578ea

This change was reflected in utils for updating a user's information: 8a4bd48

…but wasn't changed for updating a user's password. This is currently causing a "Could not update password due to an error" error on production.

This commit allows the email address of the current user to be passed in, fixing the bug.

--

Addresses: https://www.pivotaltracker.com/story/show/104342384


TheDoubleK:
:+1:

PR #171, opened 2015-09-29
Fix apiclient logging

robyoung:
This updates the apiclient logging to use the formatted logging introduced with #160.

In addition this changes the logging so that we only get one message per
API request (either successful or unsucessful) and the elapsed time is
added to both.


allait:
:+1:

PR #172, opened 2015-09-30
Fix mod_wsgi error with header encoding

allait:
Request ID value taken from the incoming request header is a unicode
object instead of a string. Headers set on flask response objects
might be encoded by flask before being sent to the WSGI server, but
middleware headers are probably added after that step, which means
a unicode string ends up as header value, which makes Apache mod_wsgi
respond with internal error.


robyoung:
:+1:

PR #174, opened 2015-10-01
Move upload_documents to dmutils

robyoung:
## Remove redundant _get_fields method
This is exactly the same as get_question_ids but without using the same language as the rest of the class.

## Allow question IDs to be filtered by question type
This is useful for doing actions on a questions of a particular type.

## Move upload_documents into dmutils
This was previously implemented in the supplier and admin frontends. This also introduces a couple of tests for it.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -156,8 +156,11 @@ class ContentSection(object):
     def _has_pricing_type(self):
         return any(self._is_pricing_type(q) for q in self.get_question_ids())
 
-    def get_question_ids(self):
-        return [question['id'] for question in self.questions]
+    def get_question_ids(self, type=None):
+        return [
+            question['id'] for question in self.questions
+            if type is None or question.get('type') == type
+        ]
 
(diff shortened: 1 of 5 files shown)


dmutils/content_loader.py
```diff
@@ -156,8 +156,11 @@ def get_field_names(self):
     def _has_pricing_type(self):
         return any(self._is_pricing_type(q) for q in self.get_question_ids())
 
-    def get_question_ids(self):
-        return [question['id'] for question in self.questions]
+    def get_question_ids(self, type=None):
+        return [
+            question['id'] for question in self.questions
+            if type is None or question.get('type') == type
```
TheDoubleK: ~~Does every question definitely have a type?  If not then is there a sensible default that could be added to the `.get()` here?~~

OK, yep questions must have a type or the frontend wouldn't be able to render them and all sorts of things would break.

TheDoubleK:
:+1:

robyoung:
Ooops, yea, changed title and added version bump.

PR #175, opened 2015-10-01
Add index name to SearchAPIClient init

allait:
Allows creating SearchAPIClient instance for a given index.

#### Add SearchAPIClient index creation methods
Adds methods for index and alias creation.

#### Set __str__ on APIError instances
Fixes the empty error output in Python shell and logged tracebacks.

### DIFF
--- a/dmutils/apiclient/search.py
+++ b/dmutils/apiclient/search.py
@@ -5,13 +5,17 @@ from .errors import HTTPError
 
 
 class SearchAPIClient(BaseAPIClient):
+    def __init__(self, *args, **kwargs):
+        self.index_name = kwargs.pop('index_name', 'g-cloud')
+        super(SearchAPIClient, self).__init__(*args, **kwargs)
+
     def init_app(self, app):
         self.base_url = app.config['DM_SEARCH_API_URL']
         self.auth_token = app.config['DM_SEARCH_API_AUTH_TOKEN']
         self.enabled = app.config['ES_ENABLED']
 
     def _url(self, path):
-        return u"/g-cloud/services{}".format(path)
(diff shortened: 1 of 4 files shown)


dmutils/apiclient/search.py
```diff
@@ -5,13 +5,29 @@
 
 
 class SearchAPIClient(BaseAPIClient):
+    def __init__(self, *args, **kwargs):
+        self.index_name = kwargs.pop('index_name', 'g-cloud')
```
robyoung: Providing this as a constructor argument doesn't work very well with the way we handle API clients at the moment. Could we just pass it to the functions?

allait: What do you mean? Is there an issue caused by adding a keyword argument to the constructor or is this about how we  don't want to create a separate client for each index?

I think we should definitely add index as a method argument where it will make sense, I'm trying to avoid this for now since I don't really need multi-index clients for scripts and constructor argument seems like it's less likely to break compatibility.

robyoung: I'm thinking more about how this would be used in the buyer frontend. At the moment it would be the same but I can see it very quickly turning to a map of API clients.

allait:
#### Add index argument to SearchAPIClient methods instead of constructor
One SearchAPIClient instance can be used to send requests to multiple
search indexes. Index needs to be set in each method call.

Adds a temprorary default value for the index to keep compatibility
with existing client usage. Once support for g-cloud and dos indexes
is added to the apps and index argument is passed on every method call
the default value can be removed.

robyoung:
:+1:

PR #176, opened 2015-10-05
Allow service document uploads to be private

robyoung:
The supplier frontend needs draft document uploads to be private until the framework goes live. This just passes the `public` down from `upload_service_documents` to the `upload_document` call.


allait:
:+1:

PR #177, opened 2015-10-06
[#103325898] Strip whitespace from submitted form values

pcraig3:
This pull request adds a whitespace-stripping WTF Field that we can use instead of a string field, and it strips surrounding whitespace from input submitted through our content-built forms.

[>> Story in Pivotal](https://www.pivotaltracker.com/story/show/103325898)

***

### Content Builder Data

[`ImmutableMultiDict`](http://werkzeug.pocoo.org/docs/0.10/datastructures/#werkzeug.datastructures.ImmutableMultiDict)s can be converted into a regular dict (with duplicate keys), as well as created from them.
So I've written a leading/trailing whitespace-stripping function to recursively `strip()` all values of a `dict`, ( or `list` or `tuple`, or `str`).  Not sure if this is the best solution, but it looks cool.

It addresses the problem of suppliers in their declaration copy+pasting emails with surrounding whitespace that then fail to validate with an unhelpful error message, [as pointed out](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/297#issuecomment-144679574) by @quis.

### WTF Field

<sub>Most of the explanation I've just cribbed from my other open pull request [in the supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/297#issuecomment-144679574).</sub>

I've subclassed the WTF `StringField` class with one that preprocesses all user-inputted values.  Specifically, all submitted values will be stripped of leading and trailing whitespace.

What happens if I'm trying to login with my valid email and password:

Submitted values | Processed values | Result
------------ | ------------ | -------------
`'email@email.com'`, `'password'` | `'email@email.com'`, `'password'` | :white_check_mark:
`' email@email.com '`, `'password'` | `'email@email.com'`, `'password'` | :white_check_mark:
`'email@email.com'`, `' password '` | `'email@email.com'`, `' password '` | :x:
`' email@email.com '`, `' password '` | `'email@email.com'`, `' password '` | :x:

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -176,7 +176,7 @@ class ContentSection(object):
         to their type in the section data.
         """
         # strip trailing and leading whitespace from form values
-        form_data = ImmutableMultiDict(_strip_values(form_data.to_dict(flat=False)))
+        form_data = ImmutableMultiDict((k, v.strip()) for k, v in form_data.items(multi=True))
 
         section_data = {}
         for key in set(form_data) & set(self.get_question_ids()):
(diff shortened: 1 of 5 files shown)


dmutils/content_loader.py
```diff
@@ -407,3 +410,44 @@ def read_yaml(yaml_file):
...
+
+        return s
+
+    def _strip_list(l):
+        return [_strip_values(v) for v in l]
+
+    def _strip_tuple(t):
+        return _strip_list(t)
+
+    def _strip_dict(d):
```
allait: Do we ever end up with dictionaries as values in `form_data`?

dmutils/content_loader.py
```diff
@@ -175,6 +175,9 @@ def get_data(self, form_data):
         in the form data are ignored. Fields in the form data are parsed according
         to their type in the section data.
         """
+        # strip trailing and leading whitespace from form values
+        form_data = ImmutableMultiDict(_strip_values(form_data.to_dict(flat=False)))
```
allait: I think using `form_data.iteritems(multi=True)` here might simplify `_strip_values` a bit. My first though would be that:

    form_data = ImmutableMultiDict((k,v.strip()) for k,v in form_data.iteritems(multi=True))

should be enough, but I'd test this with assertions, features and benefits, pricing, and file uploads to check that it covers the edge cases.

robyoung:
I couldn't think of any real reasons not to do this so I'm going to go with the majority. Code looks good :+1:

pcraig3:
@TheDoubleK agrees with us as well.

:grinning:

PR #178, opened 2015-10-07
Rename selection answers to supplier declaration

robyoung:
The language we have settled on for this is supplier declaration. This reflects the changes in alphagov/digitalmarketplace-api#262

## Depends on
- [x] alphagov/digitalmarketplace-api#262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# Digital marketplace utils changelog
+
+Records breaking changes from major version bumps
+
+## 9.0.0
+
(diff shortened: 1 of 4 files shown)


allait:
This looks like a good time to suggest starting a `CHANGELOG.md` file for listing breaking changes in utils. Something as simple as: version, what was the breaking change, where to look for things that need to be changed in the app, example of before/after usage.

I'm happy to do this as a separate PR if everyone agrees this might be useful.

:+1:

TheDoubleK:
@allait - yes, that would definitely be useful!

PR #179, opened 2015-10-07
Add user has_any_role method

robyoung:
Add a method to check if a user has any one of a list of roles.
This is particularly useful in the admin app where sections are available to multiple user roles.


TheDoubleK:
:+1:

PR #180, opened 2015-10-07
Convert config ints to ints

robyoung:
When a configuration option is defined as an integer it should be converted to an integer when passed in from the environment. If an invalid value is passed on the environment then it should fail.

This change also makes boolean configuration options fail if passed a non-truthy value on the environment.


allait:
:+1:

allait:
Note, this shouldn't be merged without a version bump, I think git doesn't warn because both changes are `8.4.0` to `8.5.0`

PR #181, opened 2015-10-07
More logging fixes

robyoung:
The correct way of logging is now to provide a format string and then the arguments in the `extra` parameter so that they get picked up and added as distinct JSON fields.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -23,10 +23,12 @@ def init_app(app):
 
     @app.after_request
     def after_request(response):
-        current_app.logger.info('%s %s %s',
-                                request.method,
-                                request.url,
-                                response.status_code)
+        current_app.logger.info('{method} {url} {status}',
+                                extra={
+                                    'method': request.method,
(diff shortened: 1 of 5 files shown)


dmutils/logging.py
```diff
@@ -23,10 +23,12 @@ def init_app(app):
...
     def after_request(response):
-        current_app.logger.info('%s %s %s',
-                                request.method,
-                                request.url,
-                                response.status_code)
+        current_app.logger.info('{method} {url} {status_code}',
+                                extra={
+                                    'method': request.method,
+                                    'url': request.url,
+                                    'status_code': response.status_code
```
allait: Should we go with `status` to match apache/nginx log field name?

allait:
There are a few more cases in https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/deprecation.py#L19-L21

robyoung:
Good shout, I've updated.

allait:
:+1:

PR #182, opened 2015-10-09
Support frameworks and manifests in ContentLoader

robyoung:
As we move to handling multiple frameworks the `ContentLoader` needs to be able to support this. This commit changes the `ContentLoader`'s role so that it is responsible for managing multiple manifests across multiple frameworks. Rather than creating multiple `ContentLoader` instances we create one, pointing it at the root content directory (`digitalmarketplace-frameworks`). On application startup we load the manifests we need and then get builders by framework slug and manifest name. If a manifest is not found either when loading or when getting a buidler we error with a `ContentNotFound` error and should fail fast.

Note that this changes `read_yaml` so that it raises an error if the file is not found. `read_yaml` only seems to be used by the content loader so this should not cause errors.
#### At application startup

Old
```python
existing_service_content = ContentLoader(
    'app/content/frameworks/g-cloud-6/manifests/edit_service.yml',
    'app/content/frameworks/g-cloud-6/questions/services/'
)
new_service_content = ContentLoader(
    'app/content/frameworks/g-cloud-7/manifests/edit_submission.yml',
    'app/content/frameworks/g-cloud-7/questions/services/'
)
declaration_content = ContentLoader(
    'app/content/frameworks/g-cloud-7/manifests/declaration.yml',
    'app/content/frameworks/g-cloud-7/questions/declaration/'
)
```

New
```python
content_loader = ContentLoader('app/content')
content_loader.load_manifest('g-cloud-6', 'edit_service', 'services')
content_loader.load_manifest('g-cloud-7', 'edit_submission', 'services')
content_loader.load_manifest('g-cloud-7', 'declaration', 'declaration')
```

#### In the views

Old
```python
content = declaration_content.get_builder()
```

New
```python
content = content_loader.get_builder(framework_slug, 'declaration')
```

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -2,6 +2,8 @@ import yaml
 import inflection
 import re
 import os
+from collections import defaultdict
+from functools import partial
 from werkzeug.datastructures import ImmutableMultiDict
 
 from .config import convert_to_boolean, convert_to_number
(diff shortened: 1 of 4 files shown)


robyoung:
NOT READY: I still need to update the CHANGELOG but needed a PR link

dmutils/content_loader.py
```diff
@@ -344,32 +346,72 @@ def _has_assurance(self, key):
...
+    """
+    def __init__(self, content_path):
+        self.content_path = content_path
+        self._content = defaultdict(dict)
+        # A defaultdict that defaults to a defaultdict of dicts
+        self._questions = defaultdict(partial(defaultdict, dict))
+
+    def get_builder(self, framework_slug, manifest):
+        try:
+            return ContentBuilder(self._content[framework_slug][manifest])
```
allait: I usually try to avoid defaultdicts since e.g. this will add a `framework_slug` key to `_content` whenever it's called with a missing slug. This might be not an issue for our use case, but I'd consider making `load_manifest` create framework keys explicitly

quis:
Got this working locally in the supplier app. Didn't encounter any problems. I think the level of sophistication is about right when we're still dealing with a low number of frameworks.

dmutils/content_loader.py
```diff
@@ -344,32 +346,72 @@ def _has_assurance(self, key):
...
+        except KeyError:
+            raise ContentNotFoundError("Content not found for {} and {}".format(framework_slug, manifest))
+
+    def load_manifest(self, framework_slug, manifest, question_set):
```
allait: I'd go with `framework_slug, question_set, manifest` argument order, since manifests can only cover questions from one question set, and this would highlight the hierarchical order

quis: I like your second point, eg changing the manifest from this:

```
-
  name: Page 1
  editable: True
  questions:
    - example
```

to something like this:
```
questions_path: declaration
sections:
  -
    name: Page 1
    editable: True
    questions:
      - example
```

robyoung: First point yes, I'll do that.

Second point yes, but I didn't want to do that as part of this PR.

allait:
:+1:

PR #183, opened 2015-10-09
Add find frameworks

robyoung:
Add methods to find all frameworks and get a single framework.

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -412,6 +412,14 @@ class DataAPIClient(BaseAPIClient):
                 },
             })
 
+    def find_frameworks(self):
+        return self._get("/frameworks")
+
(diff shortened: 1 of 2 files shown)


dmutils/apiclient/data.py
```diff
@@ -412,6 +412,14 @@ def update_service_status(self, service_id, status, user):
                 },
             })
 
+    def find_frameworks(self):
+        return self._get("/frameworks")
+
+    find_frameworks_iter = make_iter_method('find_frameworks', 'frameworks', 'frameworks')
```
allait: Is the `/frameworks` endpoint paginated? I doubt we really need the `_iter` for it, but I guess it doesn't hurt as long as it works.

allait:
:+1:

PR #184, opened 2015-10-13
Framework and status params in find_services

allait:
#### Add framework and status parameters to apiclient .find_services
Adds support for the framework and status parameters from
alphagov/digitalmarketplace-api#266

`framework` and `status` should contain comma-separated strings of
values, but similar to OR filters in search API the formatting is
left to the caller of the method.

#### Log apiclient 404 responses with INFO level
Data and search APIs return 404 for a lot of common requests:

* Requesting a missing service ID by user request
* Removing a service that was never published from the search index

Logging them as INFO to make more important log messages more visible
and reduce the logging output of scripts.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import flask_featureflags
 from flask_featureflags.contrib.inline import InlineFeatureFlag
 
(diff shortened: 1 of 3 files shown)


TheDoubleK:
:+1:

PR #186, opened 2015-10-14
Add question numbers automatically

quis:
## Add question numbering method to content builder

This will be useful for:
- numbering the questions in the declaration, rather than doing it in Jinja
- looking up a question's number when it's referred to from another question, eg "if you answered 'no' to question 23…"

## Add question number as a property of each question

This extends the above to be more immediately useful, eg in a template:

``` Jinja
<label>
  {{question.number}}. {{question.question}}
</label>
```

It preserves the numbering after any filtering.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -111,6 +111,14 @@ class ContentBuilder(object):
             if question:
                 return question
 
+    def get_question_number(self, question_id):
+        index = 0
+        for section in self.sections:
+            for question in section.questions:
+                index += 1
+                if question['id'] == question_id:
+                    return index
+
 
 class ContentSection(object):
     @classmethod
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -22,6 +22,9 @@ class ContentBuilder(object):
     """
     def __init__(self, sections):
         self.sections = [ContentSection.create(section) for section in sections]
+        for section in self.sections:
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -83,7 +86,7 @@ def _get_section_filtered_by(self, section, service_data):
         section = section.copy()
 
         filtered_questions = [
-            question for question in section.questions
+            self._add_question_number(question) for question in section.questions
```
robyoung: This isn't needed, the question numbers are added when a new `ContentBuilder` instance is created in [`filter`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/content_loader.py#L80).

dmutils/content_loader.py
```diff
@@ -111,6 +114,19 @@ def get_question(self, question_id):
             if question:
                 return question
 
+    def get_question_number(self, question_id):
+        index = 0
+        for section in self.sections:
+            for question in section.questions:
+                index += 1
+                if question['id'] == question_id:
```
robyoung: This is a lot of looping, for every question we loop through every question. If this is just being done in the constructor we could do this with one pass.

quis:
@robyoung You're right on both comments. Fixed in aa0bc82.

robyoung:
:+1:

PR #187, opened 2015-10-14
Move init app

robyoung:
## Move init_app into flask_init module
We try to keep our __init__ modules as clear as possible and import from sub modules where possible. With the addition of init_manager in the next commit the motivation becomes more.

## Add init_manager
This adds a function to initialize a flask manager object. It adds a single command to list all routes in all flask applications.


allait:
:+1:, but should be merged after #191

PR #188, opened 2015-10-15
Allow each content section to have a description

quis:
This will be useful for the declaration, where in the guidance each page has an introductory paragraph. Being able to have this in the interface might mean we don't have to publish the documentation.

Suggested usage, `manifest.yml`:
```YAML
-
  name: Grounds for mandatory exclusion
  description: You must be able to answer all of these questions
  questions:
    - …
```


TheDoubleK:
:+1: 

and :+1: to not publishing guidance.

PR #189, opened 2015-10-16
Have content loader manage framework messages

quis:
This is to let us manage pieces of content which are specific to a combination
of:
- a place on a page
- a framework
- the status of that framework
- the status of a supplier‘s application to a framework

eg, on the supplier dashboard there is a block telling you about your,
declaration which has different content through different stages of the process.
Currently we manage this with lots of {% if %} statements in the templates,
which is:
- hacky
- not scaleable

These pieces of content will be stored in the frameworks repo, in a directory
called 'messages', eg
`./frameworks/g-cloud-7/messages/index.yml`

These file will contain a dictionary of statuses, eg
```
g-cloud-7/messages/index.yml
---
- open-not_interested:
    message: please register interest
- open-registered_interest:
    message: please complete your application
- pending-registered_interest:
    message: you only did the registering interest bit
```

The format of the filename is:
`<block>.yml`

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -367,10 +367,15 @@ class ContentLoader(object):
     >>> loader.load_manifest('framework-1', 'manifest-2', 'question-set-1')
     >>> loader.load_manifest('framework-1', 'manifest-3', 'question-set-2')
     >>> loader.load_manifest('framework-2', 'manifest-1', 'question-set-1')
+    >>>
+    >>> # preload messages
+    >>> loader.load_messages('framework-1', ['homepage_sidebar', 'dashboard'])
+    >>>
     >>> # get a builder
-    >>> loader.get_builder(framework_slug, 'manifest-1')
+    >>> loader.get_builder('framework-1', 'manifest-1')
+    >>>
     >>> # get a message
-    >>> loader.get_message(framework_slug, 'homepage_sidebar', 'in_review')
+    >>> loader.get_message('framework-1', 'homepage_sidebar', 'in_review')
     """
     def __init__(self, content_path):
         self.content_path = content_path
@@ -422,17 +427,25 @@ class ContentLoader(object):
     def get_message(
         self, framework_slug, block, framework_status, supplier_status=None
     ):
-        if block not in self._messages[framework_slug]:
-            try:
-                self._messages[framework_slug][block] = read_yaml(
-                    self._message_path(framework_slug, block)
-                )
-            except IOError:
-                raise ContentNotFoundError("No message file at {}".format(self._message_path(framework_slug, block)))
+        self.load_messages(framework_slug, [block])
 
         return self._messages[framework_slug][block].get(
             self._message_key(framework_status, supplier_status), None
         )
+
+    def load_messages(
+        self, framework_slug, blocks
+    ):
+        for block in blocks:
+            if block not in self._messages[framework_slug]:
+                try:
+                    self._messages[framework_slug][block] = read_yaml(
+                        self._message_path(framework_slug, block)
+                    )
+                except IOError:
+                    raise ContentNotFoundError(
+                        "No message file at {}".format(self._message_path(framework_slug, block))
+                    )
 
     def _root_path(self, framework_slug):
         return os.path.join(self.content_path, 'frameworks', framework_slug)
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -369,10 +369,13 @@ class ContentLoader(object):
     >>> loader.load_manifest('framework-2', 'manifest-1', 'question-set-1')
     >>> # get a builder
     >>> loader.get_builder(framework_slug, 'manifest-1')
+    >>> # get a message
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -369,6 +369,8 @@ class ContentLoader(object):
     >>> loader.load_manifest('framework-2', 'manifest-1', 'question-set-1')
     >>> # get a builder
     >>> loader.get_builder(framework_slug, 'manifest-1')
+    >>> # get a message
+    >>> messages.get(framework_slug, 'homepage_sidebar', 'in_review')
```
robyoung: I think this should be
```python
>>> loader.get_message(framework_slug, 'homepage_sidebar', 'in_review')
```

dmutils/content_loader.py
```diff
@@ -416,6 +418,27 @@ def get_question(self, framework_slug, question_set, question):
 
         return self._questions[framework_slug][question_set][question].copy()
 
+    def get_message(
+        self, framework_slug, block, framework_status, supplier_status=None
+    ):
+
+        framework = self._content[framework_slug]
+
+        if 'messages' not in framework:
```
allait: I feel like I'm missing some context here.
As far as I understand `_content[framework_slug]` is a dictionary of manifest files (which are lists of sections), why are we adding a special non-manifest dictionary item to it? And what does the message yaml file look like? I have a suspicion that `ContentBuilder` won't really apply to them.

dmutils/content_loader.py
```diff
@@ -432,6 +458,20 @@ def _populate_section(self, framework_slug, question_set, section):
...
+        self, block, framework_status, supplier_status
+    ):
+        return '{}-{}{}'.format(
+            block,
+            framework_status,
+            '-{}'.format(supplier_status) if supplier_status else ''
+        )
+
+    def _message_error(self, framework_slug, message):
+        self._content[framework_slug]['messages'][message] = None
```
allait: Why is this needed? No one should be catching the `ContentNotFoundError`, so the process will termniate once the exception is thrown.

allait:
What does the content of each `<block>-<status>-<optional:supplier_status>.yml` file look like? Is there a way we could join them into `<block>.yml` and avoid having an additional structure of nested keys within the file names?

pcraig3:
Here's what I've done so far as far as messages go: https://github.com/alphagov/digitalmarketplace-frameworks/commit/4d6fffec066d6fdb05f4360f05e9468ca9028df7

dmutils/content_loader.py
quis: Fixed this and squashed.

allait:
@pcraig3 they look small enough so far to be merged into a single `index.yml` with a top-level framework status dictionary and avoid the name formatting

dmutils/content_loader.py
quis: Intention is that if there isn't a file for a specific block and status then the front end won't render anything. So the app needs to be able to ask for a file which doesn't exist without throwing an exception. And we should cache this as `None` so that it doesn't make a disk operation every time it does this.

dmutils/content_loader.py
quis: `ContentBuilder` won't apply. Maybe the messages should be stored in their own `dict`.

dmutils/content_loader.py
allait: This sounds like another reason to merge the files together. I think that "all files that we attempt to load should exist" is a nice property and it would be good if we could keep it.

dmutils/content_loader.py
allait: Yeah, `.get_builder` would be the only reason to add things to `_content`. We can just add a new property to store message.

quis:
@allait rewrote things to:
- store messages separately from questions
- remove the statuses from filenames, leaving them as `<block>.yml`

allait:
We could split `get_message` from `load_message` so that we can read the message files on startup similar to how we deal with manifests. It's possible to just call `.get_message` for this, but it would be a bit less clear.

quis:
How would we read in all the messages? Walk the directory? Have some kind of manifest? As a call to the constructor?

Don't see what's too bad about reading them as and if needed.

allait:
@quis we shouldn't have that many blocks so listing them the same way we do manifests should work. There's nothing a running app can do about a missing file, so it makes sense to make sure everything is in place before starting. Reading files when needed isn't a big problem as long as all related views have CI tests though.

quis:
Would it be better if content_loader was just given a list of frameworks, eg
``` python
frameworks = api.get_all_the_frameworks()  # ['g-cloud-7', 'g-cloud-8', 'dos-2', …]
content_loader = ContentLoader(frameworks)
```

Then we read each manifest for that framework, plus any messages that exist?

allait:
@quis the reason we list manifests (and schema files in the API) is that it forces the app to load the given filename, while e.g. just listing the directory doesn't mean that the file we actually expect to exist is there.

quis:
Suppose it's OK while we have a small number of frameworks. Added `.load_messages()` in 659a1ac.

allait:
:+1:

PR #191, opened 2015-10-19
Bump version to 10.3.0

quis:
https://github.com/alphagov/digitalmarketplace-utils/pull/188 got tagged as 10.2.0

Then https://github.com/alphagov/digitalmarketplace-utils/pull/184 overwrote this tag

This commit creates a new version which includes both


allait:
I think it's the other way around, #184 was merged first with 10.2.0, then #188 didn't get a tag since it already existed. So this just adds a tag to #188, which is the last merged PR right now.

:+1:

PR #192, opened 2015-10-22
Add 'markdown' jinja2 filter. 

pcraig3:
Added a markdown filter converts markdown-formatted (which we [occasionally store in .yml files](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/questions/declaration/environmentalSocialLabourLaw.yml)) strings to HTML.

Flask apps which make use of dmutils' `init_app` method will have the `markdown` filter added to their jinja templates.  

Other apps which want access to a `markdown` filter (the frontend toolkit, for example) can manually include it.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '10.5.0'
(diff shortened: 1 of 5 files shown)


quis:
:+1:

PR #195, opened 2015-10-27
Have flask.init init the Manager

quis:
The code that does this is currently repeated across all apps, and mean that the apps have Flask script as an extra dependency.

This brings the code and the dependency into utils, so the usage will look like:

## Before

``` Python
from flask.ext.script import Manager

manager = Manager(application)
init_manager(manager, 5003, ['./app/content/frameworks'])
```

## After

``` Python
init_manager(application, 5003, ['./app/content/frameworks'])


## Bump version to 11.0.0

Breaking changes because the expected arguments to Flask init have changed.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '10.7.0'
+__version__ = '11.0.0'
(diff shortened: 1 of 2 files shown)


tombye:
I've tested this across the apps and it LGTM :+1:

allait:
@quis could we add a record to `CHANGELOG.md` with the description of the breaking change for the major version bump

PR #196, opened 2015-10-27
Use new /draft-service POST endpoint for creating drafts

allait:
`/draft-service/<framework-slug>/create` is deprecated in favour of
using /draft-service and passing frameworkSlug in the payload, similar
to lot and supplierId

(https://github.com/alphagov/digitalmarketplace-api/commit/dd2ed8)


pcraig3:
:+1:

PR #199, opened 2015-10-28
Add `path_exists` method to S3

robyoung:
Add a method that returns true if the provided path exists in the bucket and false otherwise. This is useful in the supplier app for when a link to a file should be shown only if it exists but that link is to a proxy route in the supplier app, eg. agreement documents.


TheDoubleK:
Looks good, :+1:

PR #201, opened 2015-11-02
Agreements documents

robyoung:
## Add method to get signed document URL
This is a wrapper around the S3 get_signed_url method that replicates the behaviour of get_draft_document_url in the supplier frontend for replacing the S3 URL with our assets URL.

This has been moved to utils so that it can be used in the admin app.

## Add method to get agreement document path
This is copied over from the supplier frontend so that it can be used in the admin app as well. It has been added the the documents module as this is where other utility functions relating to documents have been added.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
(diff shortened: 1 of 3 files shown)


allait:
:+1:

PR #202, opened 2015-11-02
Create new draft page questions

allait:
### Remove get_framework_status apiclient method
Framework status is returned as part of the get_framework response,
so there's no need for a separate apiclient method and API endpoint.

Method isn't currently used by any of the apps.

### Add service data and page_questions arguments to create_new_draft_service

Allows creating a draft with initial service data and validating the
required page_questions fields in the API.

This changes the argument order for create_new_draft_service, which is
used by the supplier-frontend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,36 @@
 
 Records breaking changes from major version bumps
 
+## 12.0.0
+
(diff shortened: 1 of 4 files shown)


robyoung:
:+1:

PR #203, opened 2015-11-02
Include sanitised supplier name in agreement document filename

TheDoubleK:
CCS use supplier name to organise their work, so they need it to be included in the filename for the returned agreement documents.

This is exactly the same method used in the script to generate the framework agreement data.

I have confirmed with Chris at CCS that the filename format generated here is good for them to work with.

### DIFF
--- a/dmutils/documents.py
+++ b/dmutils/documents.py
@@ -179,12 +179,12 @@ def get_agreement_document_path(framework_slug, supplier_id, supplier_name, docu
     return '{0}/agreements/{1}/{2}-{1}-{3}'.format(
         framework_slug,
         supplier_id,
-        _sanitise_supplier_name(supplier_name),
+        sanitise_supplier_name(supplier_name),
         document_name
     )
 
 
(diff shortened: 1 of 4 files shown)


dmutils/documents.py
```diff
@@ -171,5 +175,19 @@ def get_signed_url(bucket, path, base_url):
...
+def get_agreement_document_path(framework_slug, supplier_id, supplier_name, document_name):
+    return '{0}/agreements/{1}/{2}-{1}-{3}'.format(
+        framework_slug,
+        supplier_id,
+        _sanitise_supplier_name(supplier_name),
+        document_name
+    )
+
+
+def _sanitise_supplier_name(supplier_name):
```
robyoung: I feel like this shouldn't be private, it's useful in it's own right at least in scripts.

TheDoubleK: Yep, fair point.  Sorted.

robyoung:
:+1:

PR #204, opened 2015-11-03
Add API client for supplier framework information

quis:
For alphagov/digitalmarketplace-api#282

Bumps version to 12.1.0


allait:
New method looks good, but needs a test

quis:
@allait f18542c

allait:
:+1:

PR #205, opened 2015-11-06
Add method for find framework suppliers

robyoung:
This is needed for listing suppliers on a framework with agreements returned in the admin app.

### DIFF
--- a/dmutils/apiclient/data.py
+++ b/dmutils/apiclient/data.py
@@ -197,6 +197,15 @@ class DataAPIClient(BaseAPIClient):
             },
         )
 
+    def find_framework_suppliers(self, framework_slug, agreement_returned=None):
+        params = {}
+        if agreement_returned is not None:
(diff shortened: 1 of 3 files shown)


dmutils/apiclient/data.py
```diff
@@ -197,6 +197,15 @@ def register_framework_agreement_returned(self, supplier_id, framework_slug, use
             },
         )
 
+    def find_framework_suppliers(self, framework_slug, agreement_returned=None):
+        params = {}
+        if agreement_returned is not None:
+            params['agreement_returned'] = 'true' if agreement_returned else 'false'
```
quis: This could default to `agreement_return=False` and then cast it to a string a couple of lines below.

robyoung: It can't default to `agreement_return=False` because that would only return supplier frameworks that have not had an agreement returned. The default is to return all supplier frameworks for this framework. The cast to the string would work though.

quis:
:+1:

PR #206, opened 2015-11-16
Update document url generation

robyoung:
This updates document URL generation to include /documents/ after the framework slug as per the discussion about URL naming. The document can be found in Google Drive with the title Digital Marketplace URLs.


TheDoubleK:
This looks good to me - I think I'm right in thinking that this is only ever used to generate a save URL for a newly uploaded document, which is then stored in the service JSON as a download link? If so then :+1:

robyoung:
Yes, that is correct.

PR #207, opened 2015-11-17
Documents and S3 updates

robyoung:
## Log file size on S3 upload

We have been having some problems with unexpected files being uploaded. This change logs the path, size and ACL of a saved file.

## Add file_is_empty function

In cases where we want to test if a file is empty using the double negative form (`if not file_is_not_empty`) is confusing.

### DIFF
--- a/dmutils/documents.py
+++ b/dmutils/documents.py
@@ -6,7 +6,7 @@ try:
 except ImportError:
     import urllib.parse as urlparse
 
-from .s3 import S3ResponseError
+from .s3 import S3ResponseError, get_file_size_up_to_maximum, FILE_SIZE_LIMIT
 
 
 BAD_SUPPLIER_NAME_CHARACTERS = ['#', '%', '&', '{', '}', '\\', '<', '>', '*', '?', '/', '$',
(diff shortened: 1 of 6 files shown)


dmutils/documents.py
```diff
@@ -116,17 +116,17 @@ def upload_service_documents(uploader, documents_url, service, request_files, se
...
-    size_limit = 5400000
-    below_size_limit = len(file_contents.read(size_limit)) < size_limit
+def file_is_empty(file_contents):
+    empty = len(file_contents.read(1)) == 0
     file_contents.seek(0)
+    return empty
 
-    return below_size_limit
+
+def file_is_less_than_5mb(file_contents):
```
TheDoubleK: `FILE_SIZE_LIMIT` could change, so this should probably be called `file_is_not_over_size_limit` or something else more generic like that.  But that would be a breaking change, so let's stick with it for now.

dmutils/s3.py
```diff
@@ -96,5 +112,12 @@ def _get_mimetype(self, filename):
         return mimetype
 
 
+def get_file_size(file_contents):
+    size = len(file_contents.read(FILE_SIZE_LIMIT))
```
TheDoubleK: It looks like this will only read a maximum of `FILE_SIZE_LIMIT` bytes?  In which case when testing the file size for a large file it will only ever be `FILE_SIZE_LIMIT`... the `file_is_less_than_5mb`  method will work OK because the test is for less than, but it seems odd, or at least wrongly named - could it be `get_file_size_up_to_maximum` to be clearer about what it does?

TheDoubleK:
Given it's idiosyncratic behaviour it might be nice to have a couple of tests for `get_file_size` - one below and one above the limit?

TheDoubleK:
Let's do it once Travis is happy :+1:

PR #208, opened 2015-11-18
Save submission documents seperately

robyoung:
With the new bucket layout service submission documents should be saved
separately to service documents. In order for Nginx to properly route
requests the paths must also be different. This change adds a property
to S3 instances that attempts to parse a digitalmarketplace bucket name
and return the short name. This is then used when generating the
document path in S3.


dmutils/documents.py
```diff
@@ -49,10 +49,11 @@ def validate_documents(files):
     return errors
 
 
-def upload_document(uploader, documents_url, service, field, file_contents, public=True):
+def upload_document(uploader, bucket_name, documents_url, service, field, file_contents, public=True):
```
allait: Should `bucket_name` be an attribute of the uploader?

robyoung: There is an attribute of the uploader called bucket name, this is the short name. I guess it could be a property. Actually, that would work really nicely.

TheDoubleK:
:+1:

PR #210, opened 2015-11-19
New pricing fields functionality

tommorris:
In Content Loader, instead of using `get_question`, a new method (`get_question_by_form_field_name`) provides a fields-compatible equivalent.

In the formatting code, a new function has been created to format pricing fields.

Existing methods have been retained until we are sure they are no longer needed.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -206,8 +206,6 @@ class ContentSection(object):
         This list of field names corresponds to the keys of the data returned
         by :func:`ContentSection.get_data`.
         """
-        return list(self.get_valid_params())
-
         return [
             form_field for question in self.questions for form_field in question.form_fields
         ]
@@ -217,15 +215,6 @@ class ContentSection(object):
         return [
             form_field for question in self.questions for form_field in question.get_question_ids(type)
         ]
-
-    def get_valid_params(self):
-        valid_params = set(self.get_question_ids())
-        for question in self.questions:
-            # if the question has fields, swap the id out for them as a valid param
-            if question.get('fields'):
-                valid_params.update(question.get('fields').values())
-                valid_params.remove(question.get('id'))
-        return valid_params
 
     def get_data(self, form_data):
         """Extract data for a section from a submitted form
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -486,6 +486,9 @@ class ContentQuestionSummary(ContentQuestion):
             self.questions = [q.summary(service_data) for q in self.questions]
         self.fields = question.fields
 
+    def _default_for_field(self, field_key):
+        return self.get('field_defaults', {}).get(field_key)
+
     @property
     def value(self):
         if self.questions:
@@ -494,8 +497,8 @@ class ContentQuestionSummary(ContentQuestion):
             return format_price(
                 self._service_data.get(self.fields.get('minimum_price')),
                 self._service_data.get(self.fields.get('maximum_price')),
-                self._service_data.get(self.fields.get('price_unit')),
-                self._service_data.get(self.fields.get('price_interval')),
+                self._service_data.get(self.fields.get('price_unit'), self._default_for_field('price_unit')),
+                self._service_data.get(self.fields.get('price_interval'), self._default_for_field('price_interval')),
             )
         return self._service_data.get(self.id, '')
 
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -203,18 +203,28 @@ class ContentSection(object):
         """Return a list of field names that this section returns
 
         This list of field names corresponds to the keys of the data returned
-        by :func:`ContentSection.get_data`. This only affects the pricing question
-        that gets expanded into the the pricing fields.
+        by :func:`ContentSection.get_data`.
         """
+        return list(self.get_valid_params())
 
         return [
             form_field for question in self.questions for form_field in question.form_fields
         ]
 
     def get_question_ids(self, type=None):
+        # TODO: variable name reconsider?
         return [
             form_field for question in self.questions for form_field in question.get_question_ids(type)
         ]
+
+    def get_valid_params(self):
+        valid_params = set(self.get_question_ids())
+        for question in self.questions:
+            # if the question has fields, swap the id out for them as a valid param
+            if question.get('fields'):
+                valid_params.update(question.get('fields').values())
+                valid_params.remove(question.get('id'))
+        return valid_params
 
     def get_data(self, form_data):
         """Extract data for a section from a submitted form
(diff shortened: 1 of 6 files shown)


dmutils/content_loader.py
```diff
@@ -187,9 +189,10 @@ def get_data(self, form_data):
         """
         # strip trailing and leading whitespace from form values
         form_data = ImmutableMultiDict((k, v.strip()) for k, v in form_data.items(multi=True))
+        valid_params = self.get_valid_params()
```
TheDoubleK: It looks like `valid_params` isn't used anywhere later in this method?

tommorris: @TheDoubleK fixed.

dmutils/formats.py
```diff
@@ -76,10 +76,44 @@ def get_label_for_lot_param(lot_to_check):
...
+        max_price_f = fields.get('maximum_price', None)
+        price_unit_f = fields.get('price_unit', None)
+        price_interval_f = fields.get('price_interval', None)
+
+        min_price = service.get(min_price_f, None)
+        max_price = service.get(max_price_f, None)
+        price_unit = service.get(price_unit_f, None)
+        price_interval = service.get(price_interval_f, None)
+
+        if min_price_f and price_interval_f and min_price and max_price:
```
TheDoubleK: For G6 and G7 services `max_price` and `price_interval` were optional fields, so they could be `None` but we'd still want to get a price string returned.

The `format_price` function will deal with `None`s for these fields: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/formats.py#L96

TheDoubleK: But I may have misunderstood... is this for DOS prices only?

TheDoubleK: I guess so because G6 and G7 aren't field-based prices?

tommorris: It is for DOS based prices, but the plan I had is to switch G7 prices over to field-based pricing and then when we are happy with it, deprecate non-field based prices.

dmutils/formats.py
```diff
@@ -76,10 +76,44 @@ def get_label_for_lot_param(lot_to_check):
...
+             fields are set in the service dictionary.
+    """
+    if 'fields' in question:
+        fields = question['fields']
+        min_price_f = fields.get('minimum_price', None)
+        max_price_f = fields.get('maximum_price', None)
+        price_unit_f = fields.get('price_unit', None)
+        price_interval_f = fields.get('price_interval', None)
+
+        min_price = service.get(min_price_f, None)
```
allait: With `None` being the default, I think we could shorten this to
```
min_price = service.get(fields.get('minimum_price'))
```
 while still keeping it readable. The presence of the field is not really used in the price string rendering logic.

dmutils/content_loader.py
```diff
@@ -203,19 +203,29 @@ def get_field_names(self):
         """Return a list of field names that this section returns
 
         This list of field names corresponds to the keys of the data returned
-        by :func:`ContentSection.get_data`. This only affects the pricing question
-        that gets expanded into the the pricing fields.
+        by :func:`ContentSection.get_data`.
         """
+        return list(self.get_valid_params())
```
robyoung: Is this still correct now that the form_fields property knows about fields

dmutils/content_loader.py
```diff
@@ -419,17 +398,52 @@ def _get_single_question_data(self, form_data):
...
     def form_fields(self):
-        if self.type == 'pricing':
-            return PRICE_FIELDS
+        if self.get('fields', False):
+            return self.get('fields').values()
         else:
+            # pricing fields should have fields.
+            # throw an assertion error if they don't.
+            # TODO: maybe we can check this elsewhere?
+            assert self.type != "pricing"
```
allait: This should be validated by the framework question schemas, but it's not easy to add (the "if fields is set type should be pricing" is validated though). I don't think there's much value in this assert here, template rendering will fail without `fields` anyway

dmutils/content_loader.py
```diff
@@ -419,17 +398,52 @@ def _get_single_question_data(self, form_data):
...
             return self.get_question_ids()
```
allait: I don't think this is right anymore: pricing questions will be nested within multiquestions, so `.form_fields` from nested questions should be returned, not question ids

dmutils/content_loader.py
```diff
@@ -265,50 +264,24 @@ def get_error_messages(self, errors, lot):
...
-                field_name = question_id = 'priceString'
-            elif message_key == 'assurance_required':
-                field_name = '{}--assurance'.format(question_id)
-
-            validation_message = self.get_error_message(question_id, message_key)
-
-            errors_map[field_name] = {
+        for field_name, message_key in errors.items():
+            if field_name == 'serviceTypes':
+                field_name = 'serviceType{}'.format(lot)
```
allait: I don't think this is right. We rewrite the `serviceType*` to `serviceType` in `question.id` on import (https://github.com/alphagov/digitalmarketplace-utils/blob/price-formatting-changes/dmutils/content_loader.py#L654), and now that each question is asked to look up the id (as opposed to searching for a match in `section.questions`) I don't think we need the id rewrite here. But `serviceTypes` is confusing, and I might be misunderstanding what happens here.

dmutils/content_loader.py
robyoung: Agreed, this assertion is intended to help debugging rather than runtime safety, until we find a better way of handling it explicitly.

dmutils/content_loader.py
robyoung: Good point, updated in 25a8c

dmutils/content_loader.py
robyoung: I'm a bit confused as to when that would have changed then because this change doesn't change that logic at all.

robyoung: I've just had a dig through the API and the rendered frontend and it looks like you're right. All I can think is that error handling for service types has been broken since June.

dmutils/content_loader.py
```diff
@@ -419,16 +401,53 @@ def _get_single_question_data(self, form_data):
...
+
+    @property
+    def _optional_form_fields(self):
+        if self.get('optional'):
+            return self.form_fields
+        elif self.get('optional_fields'):
+            return [self.fields[key] for key in self['optional_fields']]
         else:
-            return self.get_question_ids()
+            if self.questions:
```
allait: I don't think this needs to be nested, could be another `elif ... else`

dmutils/content_loader.py
```diff
@@ -464,18 +483,33 @@ def __init__(self, question, service_data):
...
         return self._service_data.get(self.id, '')
 
     @property
     def answer_required(self):
         if self.questions:
             return any(question.answer_required for question in self.questions)
-        if self.value in ['', [], None]:
+        elif self.fields:
+            return any(
+                self._service_data.get(field_name) in ['', [], None]
```
allait: If `format_price` returned an empty string when none of the fields are set we could just check `self.value` here like we do for normal questions

allait:
:+1: 

I'm not confident that `.form_fields` and `.get_data` can be safely combined with file uploads at this point (in terms of protecting from upload URLs being written directly as opposed to requiring a file to be attached and uploaded). I think it's important to capture this since it's a potential security issue for supplier/admin service editing, but since dos doesn't have document attachments this shouldn't block us right now.

robyoung:
That would only be an issue if a question has fields and also is an upload, which we don't currently support.

PR #211, opened 2015-11-19
Multiquestion questions support

allait:
### Rename ContentBuilder to ContentManifest
Class name is not used outside `dmutils` and `ContentManifest` feels
like a more descriptive name, especially since `ContentSection` was
added.

### Add support for loading nested questions
Reuses the existing section loading to load `questions:` for question
files. Since the original function also added an `id` to the section
and we need a way to create URLs addressing `multiquestions` we kept
the existing logic but replaced `id` key with `slug` to differentiate
between addressable and non-addressable questions.

Section `slug` is only used internally for now: once section data is
transformed into `ContentSection` instance it can be addressed as
`section.id`. Since renaming it is a breaking change it's kept as
TODO for now. Questions should use `.slug` for URL referencing.

### Add ContentQuestionSummary to replace service_attribute
Adds a new class for a question containing a reference to the service
data. This allows adding additional methods to the question: `.value`
and `.answer_required`.

`ContentQuestionSummary` instances are created by calling `.summary`
on manifest or section objects, which returns a copy of the object
with all questions transformed to `ContentQuestionSummary` instances.

This replaces `service_attribute.Attribute` class and removes the need
for creating section look-alike dictionaries with "rows" key for summary
table display since ordinary section objects can be used instead and
there's no need to propagate the changes in section attributes.

### Add support for individually editable questions
Adds a helper method that allows returning a multiquestion question
as a section object containing all the nested questions. Parent
section `edit_questions` is stored to the new `section.edtiable`
so we can reuse the existing write protection/display section logic.

### Allow access to question data keys as attributes on ContentQuestion
Similar to ContentSection we want to access question keys as
attributes. Since questions have a lot of keys, instead of listing
them in the constructor arguments we just set __getattr__ to continue
attribute lookup in _data dictionary keys.

### Use question.name or .question as ContentQuestion.label
Question content 'name' key can be used for short question names to
be displayed in the summary tables instead of the full question.

ContentQuestion.label attribute is now present on all questions and will
try to use name instead of `question` if set.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -114,9 +114,9 @@ class ContentManifest(object):
                 return False
         return True
 
-    def get_question(self, question_id):
+    def get_question(self, field_name):
         for section in self.sections:
-            question = section.get_question(question_id)
+            question = section.get_question(field_name)
             if question:
                 return question
 
@@ -317,12 +317,13 @@ class ContentSection(object):
                 result[key] = data[key]
         return result
 
-    def get_question(self, question_id):
+    def get_question(self, field_name):
         """Return a question dictionary by question ID"""
-        # TODO: investigate how this would work as get by form field name
+
         for question in self.questions:
-            if question['id'] == question_id:
-                return question
+            field_question = question.get_question(field_name)
+            if field_question:
+                return field_question
 
     # Type checking
 
@@ -367,6 +368,15 @@ class ContentQuestion(object):
             self.questions = None
         self.number = number
         self.data = data
+
+    def get_question(self, field_name):
+        if self.id == field_name:
+            return self
+        elif self.questions:
+            return next(
+                (question for question in self.questions if question.id == field_name),
+                None
+            )
 
     def fields(self, type=None):
         if self.questions:
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -13,12 +13,12 @@ class ContentNotFoundError(Exception):
     pass
 
 
-class ContentBuilder(object):
+class ContentManifest(object):
     """An ordered set of sections each made up of one or more questions.
 
     Example::
         # Get hold of a section
-        content = ContentBuilder(sections)
+        content = ContentManifest(sections)
         section = content.get_section(section_id)
 
         # Extract data from form data
@@ -76,7 +76,7 @@ class ContentBuilder(object):
         return self.get_next_section_id(section_id, True)
 
     def filter(self, service_data):
-        """Return a new :class:`ContentBuilder` filtered by service data
+        """Return a new :class:`ContentManifest` filtered by service data
 
         Only includes the questions that should be shown for the provided
         service data. This is calculated by resolving the dependencies
@@ -86,7 +86,7 @@ class ContentBuilder(object):
             for section in self.sections
         ])
 
-        return ContentBuilder(sections)
+        return ContentManifest(sections)
 
     def _get_section_filtered_by(self, section, service_data):
         section = section.copy()
@@ -390,7 +390,7 @@ class ContentLoader(object):
         try:
             if framework_slug not in self._content:
                 raise KeyError
-            return ContentBuilder(self._content[framework_slug][manifest])
+            return ContentManifest(self._content[framework_slug][manifest])
         except KeyError:
             raise ContentNotFoundError("Content not found for {} and {}".format(framework_slug, manifest))
 
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -35,6 +35,22 @@ class ContentManifest(object):
     def __iter__(self):
         return self.sections.__iter__()
 
+    def summary(self, service_data):
+        """Create a manifest instance for service summary display
+
+        Return a new :class:`ContentManifest` instance with all
+        questions replaced with :class:`ContentQuestionSummary`.
+
+        :class:`ContentQuestionSummary` instances in addition to
+        question data contain a reference to the service data
+        dictionary and so have additional properties used by the
+        summary tables.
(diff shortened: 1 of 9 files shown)


robyoung:
I agree that `ContentManifest` is a better name now but it does mean that `get_builder` is a really bad method name.

dmutils/content_loader.py
```diff
@@ -388,9 +498,12 @@ def get_builder(self, framework_slug, manifest):
         try:
             if framework_slug not in self._content:
                 raise KeyError
-            return ContentBuilder(self._content[framework_slug][manifest])
+            manifest = self._content[framework_slug][manifest]
         except KeyError:
             raise ContentNotFoundError("Content not found for {} and {}".format(framework_slug, manifest))
+            return ContentManifest()
```
robyoung: I've not come across this pattern before ;)

allait: I've noticed that line in diffs too, just didn't realise I was the one who added it :sob: Fixed.

allait:
@robyoung renamed and added a note to CHANGELOG, but kept `.get_builder` as an alias (https://github.com/alphagov/digitalmarketplace-utils/commit/d98c8ef919714e4379ddd4c227a16d5c79938ed3#diff-346b98f8c34fcc00d15a2cc1629cfb6aR507) just so we don't have to update all apps right now.

dmutils/content_loader.py
```diff
@@ -26,11 +30,16 @@ def __init__(self, sections):
         for section in self.sections:
             for question in section.questions:
                 question_index += 1
-                question['number'] = question_index
+                question.number = question_index
 
     def __iter__(self):
         return self.sections.__iter__()
 
+    def summary(self, service_data):
```
robyoung: I think this could do with a docstring. Without reading the commit comment it wasn't immediately obvious what this was for.

dmutils/content_loader.py
```diff
@@ -9,12 +9,16 @@
...
+class ContentManifest(object):
     """An ordered set of sections each made up of one or more questions.
 
     Example::
         # Get hold of a section
-        content = ContentBuilder(sections)
+        content = ContentManifest(sections)
         section = content.get_section(section_id)
 
         # Extract data from form data
```
robyoung: Not really to do with this pull request but this docstring style doesn't match the style used on `ContentLoader`

dmutils/content_loader.py
```diff
@@ -314,68 +315,174 @@ def unformat_data(self, data):
...
+            if self.get('assuranceApproach') and '{}--assurance'.format(self.id) in form_data:
+                return {self.id: {'assurance': form_data.get('{}--assurance'.format(self.id))}}
+            else:
+                return {}
+
+        if self.id == 'serviceTypes' or self.type in ['list', 'checkboxes']:
+            value = form_data.getlist(self.id)
+        elif self.type in ['boolean']:
+            value = convert_to_boolean(form_data[self.id])
+        elif self.type in ['percentage']:
```
robyoung: Could these just be `self.type == "whatever"`

dmutils/content_loader.py
```diff
@@ -314,68 +315,174 @@ def unformat_data(self, data):
...
+            value = convert_to_number(form_data[self.id])
+        elif not self.type == 'upload':
```
robyoung: Could this just be `self.type != 'upload'`

dmutils/content_loader.py
```diff
@@ -496,16 +607,16 @@ def _load_question(question, directory):
...
 
     return question_content
 
 
-def _make_section_id(name):
+def _make_slug(name):
     return inflection.underscore(
         re.sub(r"\s", "_", name)
-    )
+    ).replace('_', '-')
```
robyoung: This will have effect previously existing URLs. I don't think this will have any actual user impact because they should all be dead URLs anyway but worth having a think about.

dmutils/content_loader.py
```diff
@@ -314,68 +315,174 @@ def unformat_data(self, data):
...
-    >>> loader.load_manifest('framework-2', 'manifest-1', 'question-set-1')
+    >>> loader.load_manifest('framework-1', 'question-set-1', 'manifest-1')
+    >>> loader.load_manifest('framework-1', 'question-set-1', 'manifest-2')
+    >>> loader.load_manifest('framework-1', 'question-set-2', 'manifest-3')
+    >>> loader.load_manifest('framework-2', 'question-set-1', 'manifest-1')
     >>>
     >>> # preload messages
     >>> loader.load_messages('framework-1', ['homepage_sidebar', 'dashboard'])
     >>>
     >>> # get a builder
```
robyoung: Should be 'get a manifest' now :trollface:

robyoung:
The code looks great, I've left a few comments around documentation but code wise I think it's good to go. :+1:

dmutils/content_loader.py
TheDoubleK: https://www.youtube.com/watch?v=eT-HLFOS7qc

dmutils/content_loader.py
allait: PR for the supplier frontend updates the URLs in tests. We're moving most of the submission URLs under `/<lot_slug>`, so this seems like a good opportunity to change both

dmutils/content_loader.py
allait: It could be. I kept `in` checks because they're easier to add to (if we might get more boolean/numeric types) and keep the same pattern with `['list', 'checkboxes']`, but I don't think it matters too much

dmutils/content_loader.py
allait: Added a docstring

PR #213, opened 2015-11-23
Rename `fields` property in ContentQuestion.

pcraig3:
Some of our questions' .yml descriptions contain [a `fields` key](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/questions/services/agileCoachDayRate.yml) which was being overwritten by the `fields` property in ContentQuestion.  

Renaming `fields` -> `form_fields` should fix it.


allait:
:+1:

PR #214, opened 2015-11-24
Add method to get question by slug from a manifest

TheDoubleK:
There is already a method to get a question by slug from a section, but it is useful to be able to do this at the manifest level too - this avoids having to pass section IDs as well as question slugs into views that deal with subsections.


allait:
:+1:

PR #215, opened 2015-11-24
Genericise and document `get_message`

quis:
We need to use messages to store some nouns pertaining to a `lot`. For example

> You have added 1 user research lab

…means know that for the 'user research studios` lot contains things called 'labs', the singular of which is a 'lab'.

This commit genericises the `get_message` method so that using it in this way makes sense. It also adds a docstring to explain what the parameters are.


robyoung:
:+1:

PR #216, opened 2015-11-24
Multiquestion validation related fixes

allait:
### Only return non-empty questions for multiquestion .value
Filters out empty questions from multiquestion .value. This makes
it easier to check whether `empty_message` should be displayed
by checking if the multiquestion is empty (which is true only if
all nested question are empty).

### Send empty list value to the API for list/checkbox questions
New JSON schemas always define minItems for checkbox/list questions.
Since questions can be required to have a value even if the field name
is not listed in the schema required keys (e.g. questions that are part
of a multiquestions and have to be answered if any other related
question was answered) we can no longer send empty update to the API and
rely on the API returning an error based on mismatch between required
and request page_questions.

Instead, for list and checkbox questions we send the empty list value to
the API, so that it's validated against the `minItems` requirement.

### Set `answer_required=False` for optional multiquestions
If a question is optional `answer_required` message should never
be displayed, even if nested questions don't specify the optional
flag.

### Use question.hint as question_as_section description
Since 'description' is not a defined field on question content data
we use question hint instead

### Use question.form_fields for section.has_changes_to_save check
`question.id` no longer represents the service data fields writable
by the question, so there's a chance of missing answer required
API errors by not sending the update to the API or submitting the
same data as an update for sections with question.ids that don't
match any service data fields

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '15.0.0'
+__version__ = '15.1.0'
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -274,7 +274,7 @@ class ContentSection(object):
 
             errors_map[error_key] = {
                 'input_name': field_name,
-                'question': question['question'],
+                'question': question.label,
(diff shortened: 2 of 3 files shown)


robyoung:
:+1:

PR #217, opened 2015-11-25
Raise error on unrecognised error key

robyoung:
If an unrecognised error key is returned from the API then we should
raise an exception as it implies a form error.


TheDoubleK:
:+1:  once Travis pulls his finger out.

PR #218, opened 2015-11-27
Add send_application_question audit event type

allait:
A more general version of `send_g7_application_question` event for
other framework types.


robyoung:
:+1:

PR #219, opened 2015-11-27
Remove empty fields

robyoung:
When getting data from a form with the content loader empty fields
should be replaced with None to cause the field to be removed in the
API. This means that the API will throw the correct error
(answer_required) when all types of field are not filled in.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -373,7 +373,7 @@ class ContentQuestion(object):
     def get_data(self, form_data):
         if self.fields:
             return {
-                key: form_data[key]
+                key: form_data[key] if form_data[key] else None
                 for key in self.fields.values()
                 if key in form_data
             }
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -368,7 +368,7 @@ def get_question(self, field_name):
     def get_data(self, form_data):
         if self.fields:
             return {
-                key: form_data[key]
+                key: form_data[key] if form_data[key] else None
```
TheDoubleK: Do we definitely want `None` for empty arrays as well as strings?

tests/test_content_loader.py
```diff
@@ -777,11 +781,12 @@ def test_get_data(self):
             'q5': ['check 1', 'check 2'],
```
TheDoubleK: Maybe the test could also `None` an list value as well as strings, to be sure.

dmutils/content_loader.py
robyoung: This only sends `None` for questions with `fields`. Currently they can only be text fields.

TheDoubleK:
As discussed in real life, this should work :+1:

PR #221, opened 2015-12-02
Fix rendering assurance questions

pcraig3:
This pull request fixes the rendering of any questions with assurances on our service summary screens.

After rewriting the content loader, some of the logic around questions' assurances went missing.  This meant that calling `question.value` and `question.assurance` stopped working in
our templates. 

Now, `ContentQuestionSummary` objects can once again return their question `value`s and (optional) `assurance` values correctly.

***

## Broken
screen shot 2015-12-01 at 12 36 21

## Fixed
screen shot 2015-12-02 at 14 42 31

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -333,7 +333,7 @@ class ContentSection(object):
     def _has_assurance(self, key):
         """Return True if a question has an assurance component"""
         question = self.get_question(key)
-        return bool(question) and question.get('assuranceApproach', False)
+        return bool(question) and question.has_assurance()
 
 
 class ContentQuestion(object):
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -459,6 +459,9 @@ def _optional_form_fields(self):
         else:
             return []
 
+    def has_assurance(self):
+        return self.get('assuranceApproach', False)
```
quis: Would it be safer to explicitly return `True` or `False` here?

pcraig3: Yeah, I suppose.  I just cut and paste the existing logic, but you're probably right.

robyoung:
Looks good :+1:

pcraig3:
:rocket:

PR #222, opened 2015-12-02
Do not require a key for ContentLoader.get_message

robyoung:
The dates messages is just a flat mapping so it is easier to just return the whole thing.


pcraig3:
Good solution to the dates problem.
:+1:

PR #223, opened 2015-12-02
Fix error labels order and price fields links

allait:
### Fix pricing field error label not linking to the form
[#109072494](https://www.pivotaltracker.com/story/show/109072494)

Toolkit forms use question IDs for linking to the question input.
This is the same for multi-field questions: instead of using the
field name for priceMax/priceMin we should use the parent quesiton
id.

### Sort form errors to match the question order
errors_map is sorted according to the question order within the
section.

Returns an OrderedDict to preserve the question order since frontend
templates expect errors to be a dictionary.

### Move get_error_messages logic to the ContentQuestion
Similar to `.get_data`, `.get_error_messages` for individual quesiton
should be the responsibility of the question class, so we move the
logic to `ContentQuestion.get_error_messages`.

In order to check that the API didn't return any errors that don't
belong to the current section we compare the error fields with the
list of section fields, and if there are any errors that are not
part of the current section question we raise an exception.

This also takes care of ordering the errors to  match the question
order without sorting since the initial iteration is based on
section question order.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -4,7 +4,7 @@ import yaml
 import inflection
 import re
 import os
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from functools import partial
 from werkzeug.datastructures import ImmutableMultiDict
 
@@ -272,7 +272,7 @@ class ContentSection(object):
         :param lot: the lot of the service
         :return: error dictionary with human readable error messages
         """
-        errors_map = {}
+        question_errors = []
         for field_name, message_key in errors.items():
             question = self.get_question(field_name)
             if not question:
@@ -283,11 +283,17 @@ class ContentSection(object):
             if message_key == 'assurance_required':
                 error_key = '{}--assurance'.format(error_key)
 
-            errors_map[error_key] = {
+            question_errors.append((question, {
                 'input_name': error_key,
                 'question': question.label,
                 'message': validation_message,
-            }
+            }))
+
+        errors_map = OrderedDict(
+            (error['input_name'], error)
+            for _, error in sorted(question_errors, key=lambda e: self.questions.index(e[0]))
+        )
+
         return errors_map
 
     def unformat_data(self, data):
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -272,27 +272,12 @@ class ContentSection(object):
         :param lot: the lot of the service
         :return: error dictionary with human readable error messages
         """
-        question_errors = []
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -413,6 +404,27 @@ def _get_single_question_data(self, form_data):
...
+            question = self.get_question(field_name)
+            message_key = errors[field_name]
+
+            validation_message = question.get_error_message(message_key, field_name)
+
+            error_key = question.id
+            if message_key == 'assurance_required':
+                error_key = '{}--assurance'.format(error_key)
+
+            return {error_key: {
```
TheDoubleK: It looks like this will return after finding one error message? Are there questions (multiquestions?) that could have multiple error messages that need to be collected before returning? Or will it only ever be one error per question, in which case this should be OK?

allait: I don't think this could ever happen: question.id is acting as error_key for everything except assurance questions, but assurance error is returned in the same field key from the API – so it's impossible to get both value and assurance errors from the API. Multiquestions are acting as sections in form validation, so this doesn't apply to them either.

Having said that, that explanation relies on a lot of current behaviour, so I've changed this to collect all the errors for question fields. This might change the error we end up with for multi-field questions (the last field wins, as opposed to the first one), but that's similar to the previous behaviour anyway.

dmutils/content_loader.py
```diff
@@ -413,6 +404,30 @@ def _get_single_question_data(self, form_data):
 
         return {self.id: value if value else None}
 
+    def get_error_messages(self, errors):
+        error_fields = set(errors.keys()) & set(self.form_fields)
+        if not error_fields:
+            return {}
+
+        question_errors = {}
```
robyoung: Would this not mean that the order of errors within a multi-question would be undefined?

allait: Multi-questions act as sections when edited – so `ContentQuestion.get_error_messages` is only responsible for one nested question and the order is defined in the `ContentSection.questions`. This dictionary should always end up with one key right now anyway, there's a bit more explanation here: https://github.com/alphagov/digitalmarketplace-utils/pull/223#discussion_r46440837

Another thing about order of errors here is the value we end up with. Since `.form_fields` order is not defined (with fields being a dictionary) for multi-field questions that have multiple errors (eg both max and min price were empty) the error message we end up with is defined by alphabetical order rather than field order (so we could end up with max price error first).

robyoung:
:+1:

PR #224, opened 2015-12-04
Do not convert boolean false to None

robyoung:
This fixes a bug where selecting 'No' would result in None being sent to the data API. As a result of this fix Yes / No questions cannot be unanswered once answered.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -381,7 +381,7 @@ class ContentQuestion(object):
             if self.get('assuranceApproach') and '{}--assurance'.format(self.id) in form_data:
                 return {self.id: {'assurance': form_data.get('{}--assurance'.format(self.id))}}
             elif self.type in ['list', 'checkboxes']:
-                return {self.id: []}
+                return {self.id: None}
             else:
                 return {}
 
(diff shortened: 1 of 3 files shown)


dmutils/content_loader.py
```diff
@@ -402,7 +402,10 @@ def _get_single_question_data(self, form_data):
                 "assurance": form_data.get(self.id + '--assurance'),
             }
 
-        return {self.id: value if value else None}
```
allait: I missed the commit where this was added.
Should we be doing the same thing for checkboxes (line 384)? This makes it possible to delete subsections by unchecking values though.

TheDoubleK:
Could questions be unanswered before?  I thought once you've ticked Yes/No you can change the value but there's no way to uncheck all radio buttons?

Anyway, this fix looks :+1:

dmutils/content_loader.py
robyoung: Checkboxes and lists don't get passed through at all if none are checked. This does mean we currently don't have a way of unchecking the last checkbox.

allait: We check that and return an empty list at the moment (https://github.com/alphagov/digitalmarketplace-utils/pull/224/files#diff-346b98f8c34fcc00d15a2cc1629cfb6aR384). I'm wondering if we should return `None` there too.

allait:
:+1:

PR #225, opened 2015-12-04
Update python-json-logger to use PyPI

robyoung:
This package was broken in PyPI for a while, it is now fixed.


quis:
Link to the issue for posterity: https://github.com/madzak/python-json-logger/issues/30

:+1:

PR #226, opened 2015-12-04
Multiple recipient emails

allait:
### Add support for multiple recipients to send_email
`to_email_addresses` can be a string or a list of strings.
Emails to multiple recipients are sent with `preserve_recipients`
flag set to `False`, which makes Mandrill send a separate 'To:' email
for each address.

Removed ignored ip_pool option and optional recipient name.

### Set async=True on mandrill API send email requests
Since we're not checking the status of messages in the API response
there's no reason to wait for them to be sent. Emails with more than
10 recipients are always sent asynchronously, but for messages with
less than 10 to addresses the API response can still take some time.

From [mandrill docs](https://mandrillapp.com/api/docs/messages.python.html):

> enable a background sending mode that is optimized for bulk sending.
> In async mode, messages/send will immediately return a status of
> "queued" for every recipient.

### DIFF
--- a/dmutils/email.py
+++ b/dmutils/email.py
@@ -1,6 +1,8 @@
 import hashlib
 import base64
+
 from flask import current_app
+from flask._compat import string_types
 from mandrill import Mandrill, Error
 from itsdangerous import URLSafeTimedSerializer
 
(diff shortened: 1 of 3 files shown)


dmutils/email.py
```diff
@@ -27,25 +25,21 @@ def send_email(
             'from_name': from_name,
             'to': [{
                 'email': email_address,
-                'name': 'Recipient Name',
```
TheDoubleK: Now there's a placeholder that hung round for a long time.

TheDoubleK:
Looks good :+1:

dmutils/email.py
```diff
@@ -27,25 +25,21 @@ def send_email(
...
-            }]
+            } for email_address in to_email_addresses]
         }
 
-        result = mandrill_client.messages.send(
-            message=message,
-            async=False,
-            ip_pool='Main Pool'
-        )
+        result = mandrill_client.messages.send(message=message, async=True)
```
robyoung: We should set up a webhook so that we're at least notified of failures. I don't think it's a blocker for this PR.

PR #105, opened 2015-07-27
Bump version to 3.4.0

quis:
Includes 
- #102 Add asset fingerprinting

https://github.com/alphagov/digitalmarketplace-utils/compare/3.3.1...bump-version-3.4.0

PR #228, opened 2015-12-09
Allow a custom timestamp to be set on S3 files

robyoung:
When saving a file in S3 write the timestamp into a custom metadata field so that it can be overridden with a custom timestamp.

This is required for moving G7 documents between buckets but retaining the original upload timestamp.

### DIFF
--- a/dmutils/s3.py
+++ b/dmutils/s3.py
@@ -34,13 +34,26 @@ class S3(object):
 
         return match.group(1)
 
-    def save(self, path, file, acl='public-read', move_prefix=None):
+    def save(self, path, file, acl='public-read', move_prefix=None, timestamp=None):
+        """Save a file in an S3 bucket
+
+        canned ACL list: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
+
+        :param path:        location in S3 bucket at which to save the file
+        :param file:        file object to be saved in S3
+        :param acl:         S3 canned ACL
+        :param move_prefix: Prefix to give to existing file when moving it out of the way
+        :param timestamp:   Timestamp to set for this file rather than using utcnow
+
+        :return: S3 Key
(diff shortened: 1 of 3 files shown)


dmutils/s3.py
```diff
@@ -34,13 +34,26 @@ def bucket_short_name(self):
...
+
+        :return: S3 Key
+        """
         path = path.lstrip('/')
 
         self._move_existing(path, move_prefix)
 
         key = self.bucket.new_key(path)
         filesize = get_file_size_up_to_maximum(file)
+        key.set_metadata('timestamp', (timestamp or datetime.datetime.utcnow()).isoformat())
```
TheDoubleK: It seems weird to set the `timestamp` metadata here but then look at `last_modified` for the sort later (line #105) - are they both the same thing in S3?

dmutils/s3.py
```diff
@@ -99,12 +112,16 @@ def _format_key(self, key):
         :return:    dict
         """
         filename, ext = os.path.splitext(os.path.basename(key.name))
+        timestamp = None
+        if load_timestamps:
+            key = self.bucket.get_key(key.name)
```
allait: Why do we need to call `.get_key` when we already have a `Key` instance? Are objects returned by `bucket.list` different from `.get_key` ones?

dmutils/s3.py
robyoung: `timestamp` is a custom metadata that we can write and then we read out again on line 118. The reason I went with a different field name was to make it clear that this isn't necessarily the `last_modified` time.

I suppose what would be most accurate would be to add a `timestamp` field to the dict returned from `_format_key`  that falls back to `last_modified`. This would require more changes to the frontend apps though.

pcraig3:
:+1: Looks good.  Keys from S3 are weird.

PR #229, opened 2015-12-10
Use datetime format for timestamp metadata.

robyoung:
We use it elsewhere and expect timestamps to be formatted this way


TheDoubleK:
:+1:

PR #230, opened 2015-12-17
Add method for countersigned agreement document path

TheDoubleK:
Required for:
https://www.pivotaltracker.com/story/show/109718484
and
https://www.pivotaltracker.com/story/show/105012966

### DIFF
--- a/dmutils/documents.py
+++ b/dmutils/documents.py
@@ -204,6 +204,16 @@ def get_agreement_document_path(framework_slug, supplier_id, supplier_name, docu
     )
 
 
+def get_countersigned_agreement_document_path(framework_slug,
+                                              supplier_id,
+                                              document_name="countersigned-framework-agreement.pdf"):
+    return '{0}/agreements/{1}/{1}-{2}'.format(
+        framework_slug,
+        supplier_id,
+        document_name
+    )
+
+
 def sanitise_supplier_name(supplier_name):
     """Replace ampersands with 'and' and spaces with a single underscore."""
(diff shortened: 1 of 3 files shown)


dmutils/documents.py
```diff
@@ -190,6 +190,16 @@ def get_agreement_document_path(framework_slug, supplier_id, supplier_name, docu
     )
 
 
+def get_countersigned_agreement_document_path(framework_slug,
+                                              supplier_id,
+                                              document_name="countersigned-framework-agreement.pdf"):
```
robyoung: Why does `document_name` need to be parameterised? If the function is for just getting countersigned agreement document paths (as the name suggests) then the filename should always be the same shouldn't it?

TheDoubleK: Just in case?

robyoung: It feels like the wrong function name if we want the document name to be parameterised. It's a bit of a pain because `get_agreement_document_path` would be perfect but it's taken.

robyoung: Although, something that should be parameterised is the extension as signed framework agreements can be returned as JPEGs.

TheDoubleK: I thought about this but I don't think CCS will countersign things as a JPEG.  But I'll check to be sure.

TheDoubleK:
Removed paramaterisation of `document_name` - and confirmed that CCS will only give us PDFs as countersigned framework documents.

robyoung:
:+1:

PR #231, opened 2015-12-18
adding file is PDF and ZIP functions

tommorris:
Moving file format checking logic from admin frontend over to dmutils.


robyoung:
:+1:

PR #232, opened 2015-12-21
Add method to export users

pcraig3:
Add method that gets back the users for a certain framework. 
Add test that checks the method is constructing the right URL.

Depends on:

 - [ ] https://github.com/alphagov/digitalmarketplace-api/pull/312


TheDoubleK:
Needs a version bump, then :+1:

PR #233, opened 2015-12-31
Remove apiclient and audit modules

allait:
### Remove apiclient and audit modules in favour of dmapiclient
apiclient and audit were moved to the new dmapiclient package, so
local modules can be removed.

`dmutils.apiclient` and `dmutils.audit` import paths are preserved
for package-level names to simplify transition, but imports from
apiclient modules (eg `from dmutils.apiclient.errors ....`) will
have to be changed.

### Configure new dmapiclient logger together with dmutils
Since API clients use the module name as logger name it has changed
after the modules were moved to the new `dmapiclient` package.

The new logger has to be configured the same way we configure `dmutils`.


robyoung:
:+1:

PR #235, opened 2016-02-02
Add login utils

TheDoubleK:
This moves token-decoding functions into utils, as they are used by both buyer and supplier app now (well, the `decode_invitation_token` is used by both, and I think it is worth keeping them together so `decode_password_reset_token` is here too, even though that is only used by the buyer app.

I'm aware that this is missing tests for failure modes of `decode_password_reset_token`.  Suggestions for how to do this welcome.  I have lost quite a lot of time to it so far - I've tried using freezegun to pass a too-old token, and tried passing in an incomplete token but in both cases I get a weird "working outside of app context" error which I've failed miserably to debug.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '16.2.1'
+__version__ = '16.3.0'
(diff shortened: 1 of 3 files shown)


robyoung:
The functions could be made easier to test by splitting out the flask related stuff from the logic. However, freezegun should just work™

robyoung:
Needs a version bump, but other than that :+1:

PR #236, opened 2016-02-03
Add optional content disposition header when uploading a file to S3

pcraig3:
`Content-Disposition` header forces a file to be downloaded instead of viewed in the browser and suggests a different filename when the file is saved locally.


allait:
:+1:

PR #237, opened 2016-02-03
Encode Content-Disposition header

pcraig3:
Sending a unicode string would result in escaped spaces in our header value.  
ie, we were getting something that looked like:
`attachment;%20filename="DM_Functional_Test_Supplier-11111-signed-framework-agreement.pdf"`

Encoding in utf-8 fixes the bug.


allait:
:+1:

PR #238, opened 2016-02-03
No supplier name in agreement document path

allait:
### Don't add supplier name to the agreement document path
S3 file names now only have supplier id and document name, so that
they don't need to change if supplier name is changed.

Supplier name should be added to the uploaded file Content-Disposition
header by the app.


pcraig3:
:+1:

robyoung:
Coveralls is stupid! :+1:

PR #240, opened 2016-02-11
Make make content section slugs nicer

robyoung:
Strip ampersands and question marks from section names an trim them when converting them into slugs. Question marks in slugs make my day really bad.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '17.0.0'
+__version__ = '17.1.0'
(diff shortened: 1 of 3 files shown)


TheDoubleK:
The change looks good, but I just wonder if it might break some tests elsewhere?  It would be good to run both the SSP functional tests and the supplier app unit tests against this new version before merging.

robyoung:
Good call, I will look into that now.

robyoung:
Phew! Finally. After battling with my local environment I've managed to test this and it :sparkles: passes functional tests :sparkles:.

TheDoubleK:
Nice work :+1:

PR #242, opened 2016-02-15
[#112338109] ContentLoader accepts `yes_no_question` type questions

pcraig3:
Supplier responses to briefs are going to be returning a new data type to the `ContentLoader`.
I've called them `boolean_list` in the manifest files -- essentially they're lists of boolean values.  ([Pull request for `boolean_list` is open in digitalmarketlace-utils right now](https://github.com/alphagov/digitalmarketplace-frameworks/pull/196))

I've put them directly after `list` questions in our test because it seemed appropriate.
Then I reindexed all of the following questions, which makes the diff harder to look at.
<sub>(Sorry about that.) </sub>

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/112338109)

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -377,7 +377,7 @@ class ContentQuestion(object):
             return self._get_single_question_data(form_data)
 
     def _get_single_question_data(self, form_data):
-        if self.id not in form_data:
+        if self.id not in form_data and self.type != 'yes_no_question':
             if self.get('assuranceApproach') and '{}--assurance'.format(self.id) in form_data:
                 return {self.id: {'assurance': form_data.get('{}--assurance'.format(self.id))}}
             elif self.type in ['list', 'checkboxes']:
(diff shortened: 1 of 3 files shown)


carolinegreen:
Looks fine :+1:

dmutils/content_loader.py
```diff
@@ -387,6 +387,30 @@ def _get_single_question_data(self, form_data):
...
+            boolean_list_indexes = [
+                int(key.split('-')[-1]) for key in form_data.keys()
+                if key.startswith("{}-".format(self.id)) and
+                key.split('-')[-1].isdigit()
+            ]
+
+            if not len(boolean_list_indexes):
+                return {}
+
+            # we don't need each value in the list of indexes, we just want the max value
```
TheDoubleK: The wording of this comment could just be something like: `# check for values for all keys up to the max index`

It's a shame there's no way to do anything at this point for missing responses beyond the max index, but I guess we just need to rely on the API complaining that the list is too short.

tests/test_content_loader.py
```diff
@@ -836,6 +844,19 @@ def test_get_data(self):
         ])
         assert section.get_data(form)['q4'] == ['']
 
+        # if we have one empty value
+        form = ImmutableMultiDict([
+            ('q5-0', '')
+        ])
+        assert section.get_data(form)['q5'] == ['']
```
TheDoubleK: Should this be a `None` rather than an empty string? Or are we leaving non-truthy strings as whatever string got submitted?

TheDoubleK:
Looks good - :+1:
Let's suck-it-and see once we start building out the frontend :)

tests/test_content_loader.py
pcraig3: Not sure what we wanted, really, but this seemed a reasonable guess.

With [lists](https://github.com/alphagov/digitalmarketplace-utils/blob/master/tests/test_content_loader.py#L837), the rule we follow is that we always return a list with an empty string.
And with [single boolean questions](https://github.com/alphagov/digitalmarketplace-utils/blob/master/tests/test_content_loader.py#L807), we return whatever non-boolean value was passed in because the API will catch it.

dmutils/content_loader.py
```diff
@@ -387,6 +387,30 @@ def _get_single_question_data(self, form_data):
 
         if self.id == 'serviceTypes' or self.type in ['list', 'checkboxes']:
             value = form_data.getlist(self.id)
+        elif self.type == 'boolean_list':
+            value = []
+
+            # if self.id is 'q5', form values will come back as 'q5-0', 'q5-1', 'q5-3', ...
+            # here, we build a list of the values after the hyphen, eg [0, 1, 3, ...]
+            boolean_list_indexes = [
```
allait: We could build a `{index: value}` mapping here and then we won't have to rebuild the key again.
Building the list then becomes:
```python
value = [None] * max(boolean_indexes.keys() + 1)
for k, v in boolean_indexes.iteritems():
    value[k] = v
```

PR #243, opened 2016-02-26
Remove serviceTypes special case check in question.get_data

allait:
Special check for serviceTypes seems to have been carried over from
the old content_loader behaviour. It was required when question types
were looked up in the questions dictionary using the form field name
and the dictionary key didn't match the modified question.id.

Since then content loader was rewritten so that question ids are
set early on and questions are never looked up in the dictionary,
so `serviceTypes` question will preserve all question metadata
(including type) and shouldn't require any special handling.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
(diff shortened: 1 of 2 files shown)


robyoung:
:+1:

PR #244, opened 2016-02-29
Fix log message double formatting

allait:
Log records are mutable, so modifying `.msg` in plaintext log
formatter means that JSON logger tries to format the string that's
already been formatted, returning an error if any of the values
contained braces.

To avoid this, instead of modifying the record object we format the
message string returned by base formatter. This could cause issues if
the default format string returns a message containing braces, but this
shouldn't happen with our current format string.


robyoung:
:+1:

PR #245, opened 2016-02-29
Remove unicode from sanitised supplier names

robyoung:
Suppliers can have unicode characters in their names. However, we do not want to include these in the sanitised name as we're unsure whether this is supported in the `Content-disposition` header by all browsers.


allait:
:+1:

PR #246, opened 2016-03-02
Inject base template data in init

TheDoubleK:
By adding this here we will no longer need to pass in `BASE_TEMPLATE_DATA` to every single view in the frontend apps.

I have tested it locally and everything seems to still work OK with the views still explicitly passing BASE_TEMPLATE_DATA into the templates, so this isn't a breaking change.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '17.3.3'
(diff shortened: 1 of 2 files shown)


allait:
Great find! :+1: :sparkles:

allait:
Could use a link to http://flask.pocoo.org/docs/0.10/templating/#context-processors in the commit message

TheDoubleK:
Reworded - will merge once Travis gets going.

PR #247, opened 2016-03-04
Use ContentQuestion ID if available

robyoung:
By default the id of a content loader question is set to the name of the YAML file. This change uses the `id` field if it is available.


TheDoubleK:
:+1:

TheDoubleK:
This will need a version bump.

TheDoubleK:
This is a breaking change until the content version is updated in the frontend apps.

PR #248, opened 2016-03-04
[#112216471] Shove the logic for `boolean_list` questions into Content Loader

pcraig3:
After a _few_ false starts on how to handle the `boolean_list_question` type, this seems to me to be the neatest way to get all the non-reusable code into the ContentLoader.  

Two main things: (1) adding brief questions to the section data and (2) getting the right error messages.
- Boolean list questions need, at some point, to receive a list of questions typed by buyers into their brief.  Adding `.inject_brief_questions_into_boolean_list_question()` methods to both the `ContentSection` and the `ContentQuestion` is a fairly straightforward way to do this.
- Calling `.get_error_messages()` on a `ContentQuestionSummary` returns the right error messages for boolean lists
 - _if_ `self.id` key is returned by API's validation messages
 - _if_ `self.type == 'boolean_list'`
 - _if_ a `boolean_list_questions` key exists

Logic in the supplier app is [wayyy simpler](https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/59f5c68e708420b00029ad19c84a16dc204f62d5).

Wrote the tests.  Should be good to go.

### DIFF
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -318,6 +318,10 @@ class ContentSection(object):
             if question.get('slug') == question_slug:
                 return question
 
+    def inject_brief_questions_into_boolean_list_question(self, brief):
+        for question in self.questions:
+            question.inject_brief_questions_into_boolean_list_question(brief)
+
     # Type checking
 
     def _has_assurance(self, key):
@@ -493,6 +497,10 @@ class ContentQuestion(object):
         else:
             return []
 
+    def inject_brief_questions_into_boolean_list_question(self, brief):
+        if self.type == 'boolean_list':
+            self.boolean_list_questions = brief[self.id]
+
     def has_assurance(self):
         return True if self.get('assuranceApproach') else False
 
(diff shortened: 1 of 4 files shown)


TheDoubleK:
You're seriously asking us to look at this?
screen shot 2016-03-04 at 16 56 59

pcraig3:
Wouldn't dream of merging it, Kev.

I'll write the tests if we think it's a good direction to head in.

pcraig3:
screen shot 2016-03-07 at 10 58 21

dmutils/content_loader.py
```diff
@@ -319,6 +318,10 @@ def get_question_by_slug(self, question_slug):
             if question.get('slug') == question_slug:
                 return question
 
+    def inject_brief_questions_into_boolean_list_question(self, brief):
```
TheDoubleK: Not sure if this name should be more generic - we could in theory have boolean lists that aren't to do with briefs?
e.g. `inject_list_of_questions_into_boolean_list` or similar.

dmutils/content_loader.py
```diff
@@ -494,6 +497,10 @@ def _optional_form_fields(self):
         else:
             return []
 
+    def inject_brief_questions_into_boolean_list_question(self, brief):
```
TheDoubleK: Also here - maybe not mention `brief` in the name. And do we really want the method name to be exactly the same? I find it confusing having two methods wit the same name in the same file, even if they do belong to different classes.

TheDoubleK:
Nice work on the tests.

I'm not sure about the naming stuff - maybe just stick with briefs for now as it makes it clear what the methods expect and what it's all used for, and it can be made generic if it's ever re-used for something else?

Gets a :+1:  from me, but would be nice to have someone else look it over too I reckon.

dmutils/content_loader.py
pcraig3: Method names are exactly the same for several of the other method names (see `summary`, `get_question` or `get_error_messages`).

But I'll change it from `brief`, no problem.

dmutils/content_loader.py
pcraig3: Yeah, I tend to agree with your final comment.  We only need these for briefs, so with our really specific method naming it's obvious what they're doing.

PR #249, opened 2016-03-11
Accommodate for optional boolean_list questions

pcraig3:
`boolean_list` questions have buyers' essential/nice-to-have questions passed into them from published briefs so that they can build forms and do error messages properly.  
The `QuestionContent` object knows about itself.  It knows 
- Its `type` (ie, `boolean_list`) 
- Its `id` (ie, `niceToHaveRequirements`)

Once we pass in brief JSON data to a section, `boolean_list` questions will look for their `id` value in the brief (ie, they'll look for `brief['niceToHaveRequirements`]).  Unfortunately, since some `boolean_list` questions are optional, there was a possibility that the brief wouldn't contain the key and we'd see our 500-level error page.

This pull request recovers from situations where an optional `boolean_list` question can't be found in the brief.  It throws an error if a non-optional question is missing. 

#### Notes 

- Is it worth checking for an error?  I can't really see how a brief returned from the API could get into a state where a required bit of information is missing.
 - If not, we can stop checking if a question is optional (see [the first commit](https://github.com/alphagov/digitalmarketplace-utils/commit/35aa8dd6c3f6ff95232bbde9a6525fa04daf1cad))
 - If so, is `ContentNotFoundError` the right one to use?

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '19.0.0'
+__version__ = '19.0.1'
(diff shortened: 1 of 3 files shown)


robyoung:
:+1:

PR #250, opened 2016-03-29
Replace question value with matching option label if present

allait:
ContentQuestionSummary.value will return the matching option label for the question value instead of the value itself, so that questions that have human-readable labels for values display them in the summary tables.

This is similar to the processing we do for displaying price strings, so `.value` seems like a good place for it.


pcraig3:
:+1:

PR #251, opened 2016-03-30
Allow Reply-To to be overridden when sending email

robyoung:
MailChimp will soon only allow emails to be sent from correctly authorized domains. This change allows the Reply-To address to be different to the From address which means we can send emails from an address we own while sending the reply to the supplier.


allait:
Needs another version bump.
:+1: otherwise

PR #252, opened 2016-03-30
Add pluralize template function

robyoung:
Add a function to Jinja context for pluralization. This function works in a similar way to Rails's pluralize function (but without the magic understanding of English).
http://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize


allait:
:+1:

PR #253, opened 2016-04-06
Add has summary page property to content section

tombye:
As part of the DOS work, a need has come up for section pages to appear based on certain conditions.

1. [if a section has some description text](http://dm-prototype.herokuapp.com/outcomesqa) the user needs to read
2. [if a section holds more than one question](http://dm-prototype.herokuapp.com/outcomes-brief) the user needs to be able to see them all in one place

If neither of these are the case, and so there is only one question in the section, the user should be taken straight to that section's question.

These conditions are based on a relationship between properties on instances of the ContentSection class so it seems sensible for a property to be added which indicates the boolean result of that decision.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '19.3.0'
+__version__ = '19.4.0'
--- a/dmutils/content_loader.py
+++ b/dmutils/content_loader.py
@@ -331,7 +331,7 @@ class ContentSection(object):
 
     @property
     def has_summary_page(self):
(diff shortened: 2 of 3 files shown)


TheDoubleK:
The code here looks fine, and if it's what you need to make the briefs work then yes do it.

I'm not 100% sold on the logic though - surely a single question with a description could (at least in principle) all be rendered on a single page?  I'm just thinking of the future - might we have service submission questions with a description that should be rendered on the same page as the question? The G-Cloud lot selection question has `description`s already, and these all get shown on the same page (e.g. https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-7/questions/services/lot.yml#L7)

TheDoubleK:
:+1:

PR #255, opened 2016-04-13
Add `step` property to ContentSection

pcraig3:
Adds an optional `step` property to ContentSection objects. Pretty straightforward.

Steps are a (new!) key in our `edit_brief.yml` manifest.  Or they will be, once [the relevant pull request](https://github.com/alphagov/digitalmarketplace-frameworks/pull/220) goes in.


allait:
👍

PR #256, opened 2016-04-15
Add support for unpacking pricing 'price' field

allait:
'price' field is already part of the pricing form template. It was added when we thought specialist prices will be a number instead of a range, but it seems that utils support hasn't been finished.

This also makes unit an optional field, so that price string can be formatted with price alone. This shouldn't affect anything since questions with required price unit will have to validate it first before displaying summary price string.


TheDoubleK:
:+1:

PR #257, opened 2016-04-19
Allow abbreviations in our markdown filter

pcraig3:
Abbreviations are an extremely valuable and important part of HTML, and it's frankly shocking that we've not supported them before now.

We're using them [in some of our framework messages](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/messages/dates.yml), so this way we could reformat them as markdown and avoid having to fall back to the `|safe` filter.

[Syntax for abbreviations here](http://pythonhosted.org/Markdown/extensions/abbreviations.html).

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '19.6.0'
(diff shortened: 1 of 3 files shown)


tombye:
👍  (especially as extras included definition lists)

PR #258, opened 2016-04-19
Extend markdown less

pcraig3:
So I've made one mistake in my life.

This pull request updates the `|markdown` filter to only use the `abbr` extension.
`extra` includes 6 other extensions, none of which we need right now.


TheDoubleK:
:+1:

PR #259, opened 2016-04-20
Do not localize time when displaying dates only

TheDoubleK:
When displaying a date only from a datetime we shouldn't localize to London time.

Example
 =======
 Consider the datetime `2016-04-27T23:59:59.0Z`
 When displaying this as a date it should be "Wednesday 27 April 2016", but when localized to London, as it is BST the hour gets rolled forward and the displayed date becomes "Thursday 28 April 2016", which is wrong.

 When displaying the actual time it makes sense to continue to localize.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '19.7.1'
+__version__ = '19.7.2'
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Version numbering assumes that this gets merged first: https://github.com/alphagov/digitalmarketplace-utils/pull/258

allait:
👍

PR #261, opened 2016-04-28
Check for csv filetype

TheDoubleK:
Related to this story: https://www.pivotaltracker.com/story/show/118178617

Communications files can be PDF or CSV, so we need this additional test for document type.


pcraig3:
Looks okay to me.

👍

PR #262, opened 2016-05-03
Remove date format zero-padding

allait:
Doesn't add leading zeroes to day of the month when displaying the date.


TheDoubleK:
:+1: looks good.  I had a quick check of functional tests and I'm 99% sure there are no displayed dates that get checked, so should be fine.

PR #263, opened 2016-05-03
Replace 'percentage' question type with 'number'

allait:
Follows the change made in the frameworks repo to use a more general question type for number fields.

#### [#118085921] Don't replace 0 number values with None
Number questions submitted with '0' value showed `answer_required` validation error since the falsy value was replaced with `None`.

Fixing it so it's possible to submit '0'.

### DIFF
--- a/tests/test_content_loader.py
+++ b/tests/test_content_loader.py
@@ -820,8 +820,8 @@ class TestContentSection(object):
                 "type": "upload",
             }, {
                 "id": "q10",
-                "question": "Percentage question",
-                "type": "percentage",
+                "question": "number question",
+                "type": "number",
             }, {
                 "id": "q11",
                 "question": "Large text question",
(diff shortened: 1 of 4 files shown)


tests/test_content_loader.py
```diff
@@ -1544,6 +1550,34 @@ def test_question_value_returns_matching_option_label(self):
...
+            "type": "number",
+        }).summary({'example': '12.20'})
+
+        assert question.value == '12.20'
+
+    def test_number_questions_adds_unit_before(self):
+        question = ContentQuestion({
+            "id": "example",
+            "type": "number",
+            "unit": u"£",
```
TheDoubleK: I'm not sure if it would it be nicer to import `unicode_literals` for the whole test file?  Or better to keep it explicit?  Probably doesn't matter too much...

TheDoubleK:
:+1:

PR #264, opened 2016-05-09
Extract content loader

Wynndow:
**Removes the content loader**
Removes the `content_loader.py` file and it's tests. They've been moved to a new package here: https://github.com/alphagov/digitalmarketplace-content-loader

**Removes a couple of functions from `formats.py`**
`format_price` and `format_service_price` have both been moved to the new package as `format_price` is a dependency of the content loader, and `format_service_price` of `format_price`. Tests have also been moved.

**Mock dependency on `test_documents.py`**
`test_documents.py` was importing `ContentSection` from the content loader. As it's now been removed form `dmutils` it makes sense to mock the dependency.


allait:
Needs a rebase, but looks good otherwise 👍

PR #265, opened 2016-05-23
Add a new filter for correctly joining lists of strings

Wynndow:
Part of the work for [this](https://www.pivotaltracker.com/story/show/118472411) story on Pivotal.

Creates a new filter for joining a list of strings with a comma, except for the last item which is joined with 'and'.

It deals with a couple of edge cases. If the list only has one item, it just returns that item without a comma or 'and'. If the list has no items it just returns an empty string.

All strings are lower case.

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -3,3 +3,12 @@ from markdown import markdown
 
 def markdown_filter(text, *args, **kwargs):
     return markdown(text, ['markdown.extensions.abbr'], *args, **kwargs)
+
+
+def smartjoin(list_to_join):
+    if len(list_to_join) > 1:
+        return '{} and {}'.format(', '.join(list_to_join[:-1]), list_to_join[-1])
+    elif len(list_to_join) == 1:
+        return '{}'.format(list_to_join[0])
+    else:
+        return ''
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -5,7 +5,8 @@ def markdown_filter(text, *args, **kwargs):
     return markdown(text, ['markdown.extensions.abbr'], *args, **kwargs)
 
 
-def smartjoin(list_to_join):
+def smartjoin(input):
+    list_to_join = list(input)
     if len(list_to_join) > 1:
         return '{} and {}'.format(', '.join(list_to_join[:-1]), list_to_join[-1])
     elif len(list_to_join) == 1:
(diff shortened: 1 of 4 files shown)


allait:
Do we need to register this as a Jinja template filter in `dmutils/flask_init.py`?

Wynndow:
Yep. I was registering on the buyer front end; didn't realised they could be registered there instead. Have added.

dmutils/filters.py
```diff
@@ -3,3 +3,13 @@
 
 def markdown_filter(text, *args, **kwargs):
     return markdown(text, ['markdown.extensions.abbr'], *args, **kwargs)
+
+
+def smart_join_filter(list_to_join):
+    lower_case = list(map(lambda text: text.lower(), list_to_join))
```
allait: I suggest we remove lowercasing from this. There's already a way to lowercase a list of strings using Jinja
(` | map('lower')`) that can be chained with other filters, and only doing join would make this filter more general.

dmutils/filters.py
```diff
@@ -3,3 +3,13 @@
...
+    if len(lower_case) > 1:
+        return '{} and {}.'.format(', '.join(lower_case[:-1]), lower_case[-1])
```
allait: I'd skip the final dot here. It can always be added in the parent template and without it the filter can be used in the middle of a sentence.

dmutils/flask_init.py
```diff
@@ -55,6 +55,7 @@ def markdown_filter_flask(data):
     application.add_template_filter(formats.shortdateformat)
     application.add_template_filter(formats.dateformat)
     application.add_template_filter(formats.datetimeformat)
+    application.add_template_filter(filters.smart_join_filter)
```
allait: Having `_filter` in a filter is a bit redundant. We could make a wrapper with a different name, but I'd just rename the function to `smart_join`

Wynndow:
Yep, that all makes sense. Have updated. Will update my PR on the buyer app to reflect that changes.

dmutils/filters.py
```diff
@@ -3,3 +3,12 @@
 
 def markdown_filter(text, *args, **kwargs):
     return markdown(text, ['markdown.extensions.abbr'], *args, **kwargs)
+
+
+def smartjoin(list_to_join):
+    if len(list_to_join) > 1:
+        return '{} and {}'.format(', '.join(list_to_join[:-1]), list_to_join[-1])
+    elif len(list_to_join) == 1:
+        return '{}'.format(list_to_join[0])
```
allait: Extra dot here too

allait:
👍

PR #266, opened 2016-05-25
Send logs to STDERR only if DM_LOG_PATH is not set

allait:
At the moment, the decision to write logs to a file instead of to stdout/stderr is based on the value of DEBUG flag, with DEBUG=False environments writing to files and DEBUG=True to stderr. This worked
reasonably well when all non-DEBUG environments were hosted on AWS and the only DEBUG ones were development and testing. But this split is less useful for eg deploying to PaaS, since Cloud Foundry expects logs to be sent to stdout and stderr, but at the same time we don't necessarily want to switch apps to debug mode.

Basing the logging decision solely on DM_LOG_PATH makes more sense and would also allow us to switch debug on/off without redirecting the logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,42 @@
 
 Records breaking changes from major version bumps
 
+## 21.0.0
+
+PR: [#266](https://github.com/alphagov/digitalmarketplace-utils/pull/266)
+
+### What changed
+
+Logs will be written to a file or stdout/stderr based on the value of `DM_LOG_PATH`, even for environments with `DEBUG = True`.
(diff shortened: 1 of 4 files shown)


TheDoubleK:
I'm not sure this really counts as a breaking change?  It doesn't look like things would actually break if config was left as it is in the apps, it's just that the logging behaviour will change in a possibly surprising way.

TheDoubleK:
But yep this makes sense :+1:

allait:
It's likely to break locally, since we set the log path to `/var/log/digitalmarketplace` for all environments, and that folder probably doesn't exist for development.

TheDoubleK:
I've got it - would be fine for me :)

PR #267, opened 2016-06-17
Add filter to format links.

carolinegreen:
Add filter to format links.  Allows user inputted links to be marked up, and will prevent styled links in summary tables from breaking on mobile view.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '21.0.0'
+__version__ = '21.1.0'
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -1,4 +1,6 @@
+import re
 from markdown import markdown
+from flask import Markup
 
 
 def markdown_filter(text, *args, **kwargs):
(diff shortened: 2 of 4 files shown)


dmutils/__init__.py
```diff
@@ -3,4 +3,4 @@
 
 import flask_featureflags
 
-__version__ = '21.0.0'
+__version__ = '22.0.0'
```
TheDoubleK: If this is a major version (breaking) change it would be good to have a CHANGELOG entry to explain what has broken and how to fix apps that use it.

I'm not sure this is a breaking change though? In which case go to 21.1.0 for a new non-breaking feature.

(See the summary of version numbering at http://semver.org/)

TheDoubleK:
Is this related to any particular story? Would be good to note it in the PR description.

dmutils/__init__.py
carolinegreen: No, it's a typo :/.

dmutils/filters.py
```diff
@@ -13,3 +15,26 @@ def smartjoin(input):
         return '{}'.format(list_to_join[0])
     else:
         return ''
+
+
+def format_links(text):
+    pattern = re.compile(r"""((?:http[s]?://|www)(?:[a-zA-Z]|[0-9]|[$-_@.&#+]|[!*\(\),]|
```
TheDoubleK: This huge regex makes me nervous.  It might be worth adding a comment to explain the intent of each part of it, or maybe naming it something more descriptive than `pattern` to make it clear what it is supposed to match.

carolinegreen:
Yes, the story where summary tables break in the mobile view which I described?  I can link to the pivotal story but didn't think we were doing that in general?

dmutils/filters.py
carolinegreen: Right, it's matching urls but I can change the name.

dmutils/filters.py
```diff
@@ -13,3 +15,26 @@ def smartjoin(input):
...
+                        (?:%[0-9a-fA-F][0-9a-fA-F]))+)""", re.X)
```
TheDoubleK: This regex seems far too liberal in what it will match.
eg. `www.`, `http://.`, `www.a`

It should probably check for a minimum number of `.` separated things in the domain (at least two), and ideally that there is a valid top-level domain at the end.

Also, once a possible URL has been extracted it would be nice to sanity-check it by reading it in with something like `urlparse`.  That doesn't guarantee it's right, but would give a bit more confidence.

carolinegreen:
Paired with @TheDoubleK to write a nicer regular expression.

tombye:
👍

PR #268, opened 2016-06-21
[#121961157] Bugfix bloody unicode

TheDoubleK:
Work done with @carolinegreen 

Fixes this bug: https://www.pivotaltracker.com/story/show/121961157

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -1,4 +1,3 @@
-from __future__ import unicode_literals
 import re
 from markdown import markdown
 from flask import Markup
@@ -25,19 +24,20 @@ def format_links(text):
                                 \.                        # match dot in host name
                                 (?:[^\s/?\.#<>"']+\.?)+   # match rest of domain
                                 (?:/[^\s<>"']*)?          # rest of url
-                                [^\s<>"'\.]               # no dot at end
+                                [^\s<>,"'\.]               # no dot at end
                                 )""", re.X)
     matches = url_match.findall(text)
-    link = '{0}<a href={1} rel="external">{1}</a>'
     if matches:
-        textArray = url_match.split(text)
+        link = '{0}<a href={1} rel="external">{2}</a>'
         formattedText = ""
+        textArray = url_match.split(text)
         for text in textArray:
             if text in matches:
-                formattedText = link.format(formattedText, Markup.escape(text.decode('utf-8')))
+                url = "http://{}".format(text) if text.startswith('www') else text
+                formattedText = link.format(formattedText, Markup.escape(url), Markup.escape(text))
             else:
-                formattedText = '{}{}'.format(formattedText, Markup.escape(text.decode('utf-8')))
+                formattedText = '{}{}'.format(formattedText, Markup.escape(text))
         formattedText = Markup(formattedText)
-        return formattedText.encode('utf-8')
+        return formattedText
     else:
         return text
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -29,16 +29,19 @@ def format_links(text):
                                 )""", re.X)
     matched_urls = url_match.findall(text)
     if matched_urls:
-        link = '{0}<a href="{1}" rel="external">{2}</a>'
-        formattedText = ""
-        textArray = url_match.split(text)
-        for partial_text in textArray:
+        link = '<a href="{0}" rel="external">{1}</a>'
+        formatted_text = ""
+        text_array = url_match.split(text)
+        formatted_text_array = []
+        for partial_text in text_array:
             if partial_text in matched_urls:
-                url = "http://{}".format(partial_text) if partial_text.startswith('www') else partial_text
-                formattedText = link.format(formattedText, Markup.escape(url), Markup.escape(partial_text))
+                url = "https://{}".format(partial_text) if partial_text.startswith('www') else partial_text
+                url = link.format(Markup.escape(url), Markup.escape(partial_text))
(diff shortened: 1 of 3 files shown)


dmutils/filters.py
```diff
@@ -24,18 +25,19 @@ def format_links(text):
...
+    if matched_urls:
+        link = '{0}<a href="{1}" rel="external">{2}</a>'
         formattedText = ""
-        for text in textArray:
-            if text in matches:
-                formattedText = link.format(formattedText, Markup.escape(text))
+        textArray = url_match.split(text)
+        for partial_text in textArray:
+            if partial_text in matched_urls:
+                url = "http://{}".format(partial_text) if partial_text.startswith('www') else partial_text
```
allait: This bit is interesting since it means we're linking to HTTP by default even if there's an HTTPS version, which is not ideal. I don't think we can do the reverse either, since it'll likely break some links, while HTTP link has a chance of being promoted to HTTPS.

Can we avoid linking `www.` at all and just look for `http[s]://` so that we don't end up with broken links? I don't know how common `www.` is nowadays, and we're not looking for any other URL/domain formats.

dmutils/filters.py
```diff
@@ -24,18 +25,19 @@ def format_links(text):
...
-    if matches:
-        textArray = url_match.split(text)
+    matched_urls = url_match.findall(text)
+    if matched_urls:
+        link = '{0}<a href="{1}" rel="external">{2}</a>'
         formattedText = ""
-        for text in textArray:
-            if text in matches:
-                formattedText = link.format(formattedText, Markup.escape(text))
+        textArray = url_match.split(text)
```
allait: (nitpicking) Could we stick with underscore variable names for `textArray` and `formattedText`?

dmutils/filters.py
TheDoubleK: I know what you mean about http-only, but seemed like the only way to deal with `www` links.  I'm not sure if including `www` for linking was based on an existing need for things that are already breaking our pages.  @carolinegreen knows more than me about the original issues we've seen.

dmutils/filters.py
```diff
@@ -24,18 +25,19 @@ def format_links(text):
...
+        for partial_text in textArray:
+            if partial_text in matched_urls:
+                url = "http://{}".format(partial_text) if partial_text.startswith('www') else partial_text
+                formattedText = link.format(formattedText, Markup.escape(url), Markup.escape(partial_text))
             else:
-                formattedText = '{}{}'.format(formattedText, Markup.escape(text))
+                formattedText = '{}{}'.format(formattedText, Markup.escape(partial_text))
```
allait: Since python strings are immutable, concatenating parts one by one leads to a lot of copying. It's usually better to gather parts into an array and then `u''.join(fomatted_text_parts)` in the end

dmutils/filters.py
carolinegreen: Yeah it's bad.

dmutils/filters.py
carolinegreen: Wasn't sure of the best way of handling it, but that sounds a lot better.  Will change it.

dmutils/filters.py
```diff
@@ -24,19 +25,23 @@ def format_links(text):
...
-                formattedText = link.format(formattedText, Markup.escape(text))
+    matched_urls = url_match.findall(text)
+    if matched_urls:
+        link = '<a href="{0}" rel="external">{1}</a>'
+        formatted_text = ""
+        text_array = url_match.split(text)
+        formatted_text_array = []
+        for partial_text in text_array:
+            if partial_text in matched_urls:
+                url = "https://{}".format(partial_text) if partial_text.startswith('www') else partial_text
```
TheDoubleK: I don't think we can do https here, as it won't work for any sites that don't have SSL/TLS.  Doing http is possibly less secure, but probably slightly more reliable.

carolinegreen: I think it might actually be okay.  When you visit it without it gives you warning and also generally sites should be https these days. There was another way of doing it by putting // in front but this is advised against now and the suggestion was to always make it https instead.

I was going to ask about the http being promoted to https problem though when Alexey got back as I can't really find anything on it easily.

The problem with dropping www is that this is being put in so we could still end up with the same problem of the page not being readable because some weird long url has been put in.  I think that if we really can't make it into a link that it might actually be better to put a span around it so that it can still have the class that allows it break.

TheDoubleK: Ah yes, the warning page might be OK, I hadn't thought of that. 
But still that might put suppliers off clicking through if they think it's supposed to be https but isn't? I'm not sure what the best thing is 😕

allait: There's no warning when connecting to a web server that doesn't accept HTTPS requests at all (which feels like a more common scenario than invalid certificate/mixed content) so it will only say "This site can't be reached". `//` is relative to the origin, which means that since Digital Marketplace is https it will make all links https. 

I don't think there's a good answer to this. We're not going to build a link checker to test whether the linked page is accessible via HTTP or HTTPS, so this means we will end up with broken links in both cases (probably more if we go with `https://`). We could look at links in the existing briefs to get more data about this, but my suggestion is still to not link `www.` and let suppliers copy the link manually. This way we don't create an expectation that the link is correct and the website isn't working for some reason and might prompt the user to try a different approach (eg changing http to https manually or pasting the url to a search engine).

carolinegreen: Okay, will just put spans around the www links for now.

allait:
👍

PR #271, opened 2016-07-01
Make regex less catastrophic

TheDoubleK:
This replaces the old regular expression for URL matching with a new simpler one that doesn't die horribly on certain inputs.

The new version is not quite as precise (this one will match eg `www.foo` but the previous one required a further `.` before matching.  But we can live with this better than the whole site getting taken down by a bad brief.

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -20,18 +20,15 @@ def smartjoin(input):
 
 def format_links(text):
     url_match = re.compile(r"""(
-                                (?:https?://|www\.)       # start with http:// or www.
-                                (?:[^\s/?\.#<>"']+\.?)+   # first part of host name
-                                \.                        # match dot in host name
-                                (?:[^\s/?\.#<>"']+\.?)+   # match rest of domain
-                                (?:/[^\s<>"']*)?          # rest of url
-                                [^\s<>,"'\.]              # no dot at end
+                                (?:https?://|www\.)    # start with http:// or www.
+                                (?:[^\s<>"'/?#]+)      # domain doesn't have these characters
+                                (?:[^\s<>"']+)         # post-domain part of URL doesn't have these characters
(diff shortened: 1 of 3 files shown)


dmutils/filters.py
```diff
@@ -20,18 +20,15 @@ def smartjoin(input):
...
     url_match = re.compile(r"""(
-                                (?:https?://|www\.)       # start with http:// or www.
-                                (?:[^\s/?\.#<>"']+\.?)+   # first part of host name
-                                \.                        # match dot in host name
-                                (?:[^\s/?\.#<>"']+\.?)+   # match rest of domain
-                                (?:/[^\s<>"']*)?          # rest of url
-                                [^\s<>,"'\.]              # no dot at end
+                                (?:https?://|www\.)    # start with http:// or www.
+                                (?:[^\s<>"'/?#]+)      # domain doesn't have these characters
+                                (?:[^\s<>"']+)         # post-domain part of URL doesn't have these characters
```
risicle: This works, but it doesn't always come out as the comments suggest it does - because the post-domain part is now mandatory you now get `http://kev.gov.uk` matching with `u` as the "post domain part" and `k` as the "no dot at end".

risicle: http://pythex.org/?regex=((https%3F%3A%2F%2F%7Cwww%5C.)(%5B%5E%5Cs%3C%3E%22%27%2F%3F%23%5D%2B)(%5B%5E%5Cs%3C%3E%22%27%5D%2B)(%5B%5E%5Cs%3C%3E%2C%22%27%5C.%5D))&test_string=http%3A%2F%2Fkev.gov.uk&ignorecase=0&multiline=0&dotall=0&verbose=0

(ironic posting a long complex url in a text field about code for catching complex urls...)

allait:
👍

PR #272, opened 2016-07-12
[#124752631]format_links: apply specific (but customizable) classes to inserted html elements

risicle:
This is how I would go about tackling the issues in story https://www.pivotaltracker.com/story/show/124752631

Would clearly be an api-breakage - on that note could someone explain how we use `__version__` in -utils. Two-numbered versioning I get. Not sure what the third number is being used for (currently `21.2.2`?).

utils consumers would need to update their css to style `urlized-link` and `urlized-plaintext-link` to their desires. Or override the defaults, or whatever...

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -18,7 +18,16 @@ def smartjoin(input):
         return ''
 
 
-def format_links(text):
+def format_links(text, link_classes=("urlized-link",), plaintext_link_classes=("urlized-plaintext-link",)):
+    """
+        Replaces apparent urls in plaintext `text` with <a href="..."> elements, applying `link_classes` to the
+        resulting elements.
+
+        Apparent pseudo-urls (url-like substrings missing the scheme (beginning "www") are wrapped in a <span> element
+        to which `plaintext_link_classes` are applied.
+
+        `link_classes` and `plaintext_link_classes` are expected to be valid sensible html class names
+    """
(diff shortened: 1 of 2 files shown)


TheDoubleK:
I think this is overkill for what we need (this filter is highly unlikely to be used anywhere other than the very specific places it has been introduced for).

We already have the custom `break-link` class that does precisely what we need, so there's no need for ay new styles. The behaviour of `break-link` is already defined in the buyer app, which is the only place it is used.  So the only change we need here in utils is to add `break-link` to both plaintext and anchored detected links.

That way there would be no breaking change, and utils consumers won't need to care at all about the new behaviour.

The only change needed in the frontend is then to remove the `break-all` from `a.summary-item-body`, which is the thing that is making the `Edit` links split when they shouldn't.

I think the fix is literally a one-line change here (adding the `brek-link` class to linked URLs) and one line in the buyer app (removing the `a.summary-item-body` style). 

As this is a production bug that people can see right now I'd say go with that and ship it ASAP.

TheDoubleK:
As you suggested the other day, I would like us to investigate how well bleach (https://pypi.python.org/pypi/bleach) would fit our need here - would be much nicer than rolling our own regex as we have at the moment.

But I think that's for a future investigation, not for this bug fix.

risicle:
Fair 'nuff

PR #275, opened 2016-07-25
Adding date logic

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

The logic for working out publishing dates was being done in two places; the api and the buyer frontend. This PR is to extract the logic from the api and frontend and consolidate it into one place in the utils.

If a brief is open for 2 weeks, suppliers have 5 working days to submit questions. If it's open for 1 week they only have 2 working days. Questions need to be published and answered 1 working day before the closing date for both lengths.


dmutils/dates.py
```diff
@@ -0,0 +1,54 @@
...
+from workdays import workday
+
+
+def get_publishing_dates(brief):
+    APPLICATION_OPEN_DAYS = {'1 week': 7, '2 weeks': 14}
+    QUESTIONS_OPEN_DAYS = {'1 week': 2, '2 weeks': 5}
+    dates = {}
+
+    if brief_is_from_api_(brief):
+        dates['today'] = get_start_date_from_brief_api(brief)
```
idavidmcdonald: Would it make sense to call it `dates['published_date']` or something similar rather than `dates['today']`? If this method is run on a brief not on the day of publishing then you are not actually given todays date.

dmutils/dates.py
```diff
@@ -0,0 +1,54 @@
...
+
+def brief_is_from_frontend_app_and_published_(brief):
+    return brief.get('publishedAt') and isinstance(brief.get('publishedAt'), string_types)
+
+
+def get_start_date_from_brief_api(brief):
+    return brief.get('publishedAt').replace(hour=23, minute=59, second=59, microsecond=0)
+
+
+def get_start_date_from_published_frontend_brief(brief):
```
idavidmcdonald: Could be good to add a short unit test for this function

dmutils/dates.py
```diff
@@ -0,0 +1,54 @@
...
+    return datetime.strptime(brief['publishedAt'][0:10], '%Y-%m-%d') \
+        .replace(hour=23, minute=59, second=59, microsecond=0)
+
+
+def get_todays_date():
+    return datetime.utcnow().replace(hour=23, minute=59, second=59, microsecond=0)
+
+
+def get_length_of_brief(brief):
```
idavidmcdonald: Could be good to add a short unit test for this function

tests/test_dates.py
```diff
@@ -0,0 +1,222 @@
...
+            'publishedAt': u'2016-01-04T12:00:00.00000Z',
+        }
+
+        dates = dates_package.get_publishing_dates(brief)
+
+        assert dates['questions_close'] == datetime.datetime(2016, 1, 11, 23, 59, 59)
+        assert dates['answers_close'] == datetime.datetime(2016, 1, 15, 23, 59, 59)
+        assert dates['closing_date'] == datetime.datetime(2016, 1, 18, 23, 59, 59)
+
+    def test_get_publishing_dates_returns_correct_dates_if_publishedAt_is_a_date_object(self):
```
idavidmcdonald: `_if_published_at` seems to be convention

tests/test_dates.py
```diff
@@ -0,0 +1,222 @@
...
+                'requirementsLength': '2 weeks',
+            }
+
+            dates = dates_package.get_publishing_dates(brief)
+
+            assert dates['questions_close'] == datetime.datetime(2016, 7, 15, 23, 59, 59)
+            assert dates['answers_close'] == datetime.datetime(2016, 7, 22, 23, 59, 59)
+            assert dates['closing_date'] == datetime.datetime(2016, 7, 24, 23, 59, 59)
+
+    def test_get_publishing_dates_returns_correct_dates_if_brief_is_published_with_no_requirementLength(self):
```
idavidmcdonald: Would be good to split this into two tests as it's testing two things at the moment
- no requirement
- `publishedAt` being a string

dmutils/dates.py
Wynndow: Yep, totally. Have changed it to your suggestion and update the buyer frontend accordingly.

dmutils/dates.py
Wynndow: @idavidmcdonald My thinking here is that get_publishing_dates is the only public function of the dates package. The other functions are tested implicitly by testing get_publishing_dates. I've added a commit to mark the other functions as private which I should've done before. What do you think?

tests/test_dates.py
Wynndow: Yep!

dmutils/dates.py
idavidmcdonald: Ah OK, makes sense. I know some people may argue about whether you should or should  not test private methods, for which I haven't made an opinion. But happy to go with your call :)

tests/test_dates.py
Wynndow: @idavidmcdonald Yep, have added a couple more tests to cover the possibilities of published briefs and the different states requirementsLength could be in.

.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
allait: py.test creates `.cache` to rerun failed tests. You could add it to your global `.gitignore` so it doesn't get committed

Wynndow: Yep, added it to .gitignore

dmutils/dates.py
```diff
@@ -0,0 +1,54 @@
...
+    return datetime.strptime(brief['publishedAt'][0:10], '%Y-%m-%d') \
+        .replace(hour=23, minute=59, second=59, microsecond=0)
+
+
+def _get_todays_date():
+    return datetime.utcnow().replace(hour=23, minute=59, second=59, microsecond=0)
+
+
+def _get_length_of_brief(brief):
+    if brief.get('requirementsLength') == '1 week':
```
pcraig3: This could be `return brief['requirementsLength'] if brief.get('requirementsLength') else '2 weeks'`

pcraig3: Unless you're trying to make clear that there are only two possible `requirementsLength`s.

Wynndow: No, you're right. Your way seems a bit cleaner. Will update!

dmutils/dates.py
```diff
@@ -0,0 +1,54 @@
...
+        dates['published_date'] = _get_start_date_from_published_frontend_brief(brief)
+    else:
+        dates['published_date'] = _get_todays_date()
+
+    length = _get_length_of_brief(brief)
+
+    dates['closing_date'] = dates['published_date'] + timedelta(days=APPLICATION_OPEN_DAYS[length])
+    dates['questions_close'] = workday(dates['published_date'], QUESTIONS_OPEN_DAYS[length])
+    dates['answers_close'] = workday(dates['closing_date'], -1)
+    dates['application_open_weeks'] = APPLICATION_OPEN_DAYS[length]//7
```
pcraig3: It seems weird to keep this key in just for the purposes [of the template in the buyer frontend app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/7e6fdb56610df1b5d270b136ad3dcc08569cba9f/app/templates/buyers/brief_publish_confirmation.html#L48), when it's just recreating what our variable is already called.

If we return it in the serialise method, we can just use it directly in the template.

dmutils/dates.py
Wynndow: Or, could do `return brief.get('requirementsLength', '2 weeks')`

Or do you think it's not clear enough?

Wynndow: I've tried it both ways, and it causes an error if it encounters a brief with `requirementsLength == None` which non specialist briefs will all have. Might be safer to leave it the original way?

dmutils/dates.py
Wynndow: This is similar to the point above, if a brief doesn't have a requirementLength, like a non specialist, then setting this key an using it in the template makes sense. If requirementsLenght was called on a brief without it, the template would be off.

requirements.txt
```diff
@@ -12,3 +12,4 @@ pytz==2015.4
 Flask-WTF==0.12
 markdown==2.6.2
 Flask-Script==2.0.5
+workdays==1.4
```
pcraig3: This `workdays` package seems really cool.

Wynndow: Yeah, it'd be good if it handled bank holidays!

pcraig3: Seems like [you can feed it a list of holidays](https://pypi.python.org/pypi/workdays/). (but I don't know where we'd get them from.)

dmutils/dates.py
pcraig3: Actually, here's a parallel idea: what if you just passed the `'1 week'` or `'2 weeks'` string directly into `dates['application_open_weeks']`?  That way, we could minimize the template logic and make it easier to remove this key in the future.

dmutils/dates.py
Wynndow: The way you suggested works better, turns out that when the brief is coming form the API, requirementsLength can be set to None which makes the other way fall over. Gonna update it to your version.

dmutils/dates.py
Wynndow: Boom. That's a great idea. Will do it now.

requirements.txt
Wynndow: Ah nice, maybe one for another story in the future. I guess we can get a pretty reliable list from [here](https://www.gov.uk/bank-holidays).

allait:
Looks great 👍 Very good call about using an external library for figuring out workdays, makes everything a lot more straightforward!

PR #276, opened 2016-07-29
Move application logging section from the manual

allait:
Logging from applications belongs in the utils readme together with the logger code.


pcraig3:
👍

PR #277, opened 2016-08-17
[#119736241] Log 5xx responses with ERROR loglevel

allait:
Adds a check for the response_status and sets the log level to ERROR for 5xx responses.


TheDoubleK:
Nice :+1:

PR #279, opened 2016-08-25
[#125516661] Add metadata and logger parameters to send_email

allait:
Allows overwriting the logger to make it possible to use send_email outside app context.

Setting metadata argument will attach it to the messages in Mandrill, so that it can be used to find the relevant message using Mandrill message search API.

### DIFF
--- a/dmutils/email.py
+++ b/dmutils/email.py
@@ -19,7 +19,10 @@ class MandrillException(Exception):
     pass
 
 
-def send_email(to_email_addresses, email_body, api_key, subject, from_email, from_name, tags, reply_to=None):
+def send_email(to_email_addresses, email_body, api_key, subject, from_email, from_name, tags, reply_to=None,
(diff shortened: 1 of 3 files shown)


dmutils/email.py
```diff
@@ -40,6 +43,7 @@ def send_email(to_email_addresses, email_body, api_key, subject, from_email, fro
             'track_clicks': False,
             'auto_text': True,
             'tags': tags,
+            'metadata': metadata,
```
pcraig3: Is metadata just anything you want it to be, or does it have to be a specific thing (like a string or a dict or something else)?

allait: I think it could be anything that's serialisable to JSON. Dicts probably make the most sense.

pcraig3:
```py
{ "cool": True }
```

pcraig3:
👍

PR #280, opened 2016-09-01
add xmlutils, containing `squashed_element_text` & some whitespace-normalizing functions

risicle:
These are extremely useful in tests and it would be nice to have them available project-wide.

### DIFF
--- a/dmutils/xmlutils.py
+++ b/dmutils/xmlutils.py
@@ -1,0 +1,29 @@
+import re
+
+
+def squashed_element_text(element):
+    """
+        returns a concatenation of text contents of `element` and all its children
+    """
+    return (element.text or u"") + u"".join(
+        squashed_element_text(child_element)+(child_element.tail or u"") for child_element in element
+    )
+
+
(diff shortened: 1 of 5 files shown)


dmutils/xmlutils.py
```diff
@@ -0,0 +1,29 @@
+import re
+
+
+def squashed_element_text(element):
```
allait: [`.text_content()`](http://lxml.de/lxmlhtml.html) seems to do exactly this

risicle: God bleeding damn it.

risicle:
At least this PR taught me that lxml on my local machine and on travis do _not_ behave _quite_ alike.

allait:
@risicle whitespace normalisation seems quite useful, I think we can merge that without the text function

risicle:
I'm not sure about that - it's so trivial it doesn't seem worth it, especially seeing as I'm not even sure it is standards-compliant. Also think `lxml.html` might have some utility functions for this

PR #281, opened 2016-09-08
[#129915689]turn `_Line` utility class into more robust `_StringPipe` class

risicle:
Tangentially related to https://www.pivotaltracker.com/story/show/129915689

### DIFF
--- a/tests/test_csv_generator.py
+++ b/tests/test_csv_generator.py
@@ -2,14 +2,15 @@ from dmutils.csv_generator import iter_csv
 
 
 class TestIterCsv():
-    def test_it_creates_csv_lines_from_lit_of_rows(self):
+    def test_it_creates_csv_lines_from_list_of_rows(self):
         rows = [
             ['a', 'b', 'c', 'd'],
-            ['e', 'f', 'g', 'h']
+            ['e', u'\u00a3', 'g', 'h']
         ]
 
         result = iter_csv(rows)
         lines = [line for line in result]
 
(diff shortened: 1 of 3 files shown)


TheDoubleK:
I'm not sure I see the benefit here?  Why might we want to `iter_csv` with something other than a csv writer? What does `_StringPipe` do for us that the previous code wasn't doing?

risicle:
I was _going_ to use a `DictWriter` but it turned out I didn't in the end anyway, so much of that bit is unnecessary now.

`_StringPipe` (other than having a better name) doesn't flush its buffer every time it's written to, instead concatenating it to any existing input. `_Line` would work fine as long as your code was guaranteed to do `write` `read` `write` `read` `write` `read` `write` `read`... if you did two successive writes before a read you would end up with some of your stream missing.

risicle:
Ok, updated to remove now-redundant flexibility-adding features. It's just the `_StringPipe` fix now.

tests/test_csv_generator.py
```diff
@@ -5,11 +5,12 @@ class TestIterCsv():
     def test_it_creates_csv_lines_from_lit_of_rows(self):
```
TheDoubleK: Just spotted typo in the test name `lit` -> `list`.  I reckon we can live with it for now if you can't face waiting for Travis again.

TheDoubleK:
:+1:

tests/test_csv_generator.py
risicle: nah travis ain't too slow for utils

PR #282, opened 2016-10-03
Generate agreement upload filenames

TheDoubleK:
Add method to generate timestamped document names - this is to be used for new agreement upload process where we need a unique path for agreements.


idavidmcdonald:
👍

PR #284, opened 2016-10-04
Allow setting custom disposition-type on S3 saves with filename

TheDoubleK:
For G8 countersigning we want uploaded agreements to have `inline` disposition type but also have a filename set for when users download the files.

(The admin app inline display of PDFs breaks if Content-Disposition is set to `attachment`, but at present we can only set `attachment` with a filename.)

Currently you can only set a custom disposition-type when a filename is also supplied.  This could be changed in future if necessary to set disposition types without a filename.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
👍

PR #287, opened 2016-11-23
return email in password reset

leohemsted:
buyer-frontend previously relied on the fields `user` and `email` being in the
token that it passes into decode_password_reset_token - as that function just
passed them straight through. but by building the return dict from the user
object, we can remove the requirement on email_address being in the token
so it isn't left unencrypted in our logs

i think this is a minor change, but it might be a patch... it removes a previously required attribute of the `token` parameter but doesn't change the functionality or output


allait:
👍

PR #288, opened 2016-11-25
Use encrypted tokens to pass data in URLs

leohemsted:
inspired by AusDTO#20

Password reset/create account email tokens were previously base64
encoded and signed. Since they're part of the URL of a GET request,
they would show up in logs and google analytics, which we don't
really want. So we've replaced the use of `URLSafeTimedSerializer` with
`cryptography.fernet`. Fernet provides encryption, signing and TTL.

To implement this a few things have needed to happen.

* cryptography==1.6 is a new dependency
* generate_token now generates the new encrypted token
* decode_token, while old tokens are still in the system, has to decode
  both new and old tokens
  - it first tries to decode a signed b64 message.
  - if that raises, it then tries to decode using Fernet
  - if that raises, it raises a new dependency-agnostic
    InvalidTokenException
  - invite emails have a TTL of 7 days, so this code must remain in
    place for at least that long
* as cryptography.fernet does not distinguish invalid and expired data,
  we no longer distinguish token_expired and token_invalid errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 Records breaking changes from major version bumps
 
+## 23.0.0
+
+PR: [#288](https://github.com/alphagov/digitalmarketplace-utils/pull/288)
+
+### What changed
+
+`decode_invitation_token` previously accepted a `role` parameter, which it would switch on to assert keys of the encoded token's data. For example, if you passed in 'supplier' it would assert that the token contains 'email_address', 'supplier_id', and 'supplier_name'. The contents of the data encoded shouldn't be the responsibility of the utils app, and is left to the implementing code to either check or not.
+
+### Example app change
+
+Old:
+```python
+
+data = decode_invitation_token(token, role='supplier')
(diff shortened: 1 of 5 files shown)


allait:
I think it'd be useful to have some more tests for encrypt/decrypt with literal cypher-text (probably using `freezegun` to get consistent results). Otherwise tests don't really highlight the fact that we're not accidentally sending plaintext.

I think this also one of the places that would benefit from a good docstring for encrypt and decrypt functions explaining
* what the process is step by step
* what acts as secret key for encryption and decryption and why we need a certain format
* how TTL is checked
* why we return the timestamp and what it's used for
* why we need the `.encode`/`.decode` bits (and that it's all base64 rather than actual unicode)

Other than that this looks good to me, but with it being a security change I'd appreciate if someone else would take another look (maybe @TheDoubleK or @pcraig3) and gave the final thumbs up.

leohemsted:
can't test encryption in a consistent manner - it calls `urandom` https://github.com/pyca/cryptography/blob/master/src/cryptography/fernet.py#L50 .

can write a test for decrypt though

pcraig3:
Yeah, looks good I think. 👍

PR #289, opened 2016-12-01
Generate download filename method

TheDoubleK:
I'm moving this method here from scripts repo because the admin frontend (and possibly other apps) also need to set a download filename for supplier documents.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '23.0.0'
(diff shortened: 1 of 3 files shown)


risicle:
Shouldn't the "bump" commit come last if you want the COUNTERPART_FILENAME change to get picked up?

TheDoubleK:
I don't think so - I think the Jenkins job will pick up and new merge to master that includes an updated version number and tags the current master commit.  But I'm gonna check it now you asked.

TheDoubleK:
Yeah this will work just fine.

PR #290, opened 2016-12-09
Fix logger failure mode

galund:
- if the log message cannot be used as a format string, that should
   not be fatal - the message must still be logged.
 - do not attempt to log the exception generated when trying to use
   the message as a format string, as that exception is frequently
   not usable as a format string either, and in that case you get
   an infinite recursion / stack overflow.
 - change the message about not being able to format the message
   into log-level INFO, as it's not really an error state in and
   of itself.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -115,8 +115,20 @@ class CustomLogFormatter(logging.Formatter):
 
         try:
             msg = msg.format(**record.__dict__)
-        except KeyError as e:
-            logger.exception("failed to format log message: {} not found".format(e))
+        except:
+            # We know that KeyError, ValueError and IndexError are all possible things that can go
+            # wrong here - there is no guarantee that the message passed into the logger is
+            # actually suitable to be used as a format string. This is particularly so where an
+            # we are logging arbitrary exception that may reference code.
+            #
+            # We catch all exceptions rather than just those three, because _any_ failure to format the
+            # message must not result in an error, otherwise the original log message will never be
+            # returned and written to the logs, and that might be important info such as an
+            # exception.
+            #
+            # NB do not attempt to log either the exception or `msg` here, or you will
+            # find that too fails and you end up with an infinite recursion / stack overflow.
+            logger.info("failed to format log message")
         return msg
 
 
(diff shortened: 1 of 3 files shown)


dmutils/logging.py
```diff
@@ -115,8 +115,20 @@ def format(self, record):
...
+            # we are logging arbitrary exception that may reference code.
+            #
+            # We catch all exceptions rather than just those three, because _any_ failure to format the
+            # message must not result in an error, otherwise the original log message will never be
+            # returned and written to the logs, and that might be important info such as an
+            # exception.
+            #
+            # NB do not attempt to log either the exception or `msg` here, or you will
+            # find that too fails and you end up with an infinite recursion / stack overflow.
+            logger.info("failed to format log message")
```
TheDoubleK: This might warrant being logged as an error rather than just info?  Either way, seems a legit fix :+1:

TheDoubleK:
:+1: for the fourth wall.

dmutils/logging.py
allait: So I understand where KeyError is coming from, and I'm assuming IndexError is similar. How do we end up with ValueError? And why do we suspect there are other possible exceptions?

galund: ValueError is what you get when there's a single brace in a string, e.g. the example in my test case.

Logging and exception handling code is the one place where I would break the rule about not having generic exception handlers. The point is, if *any* exception happens here, then it must not be fatal, otherwise our logging will fail. 

The risk introduced is that something consistently gets caught by this handler and ignored: I'd say that risk is minimal, because assuming someone is _already_ investigating some particular log message, if they expected it to have had `format` applied, and it didn't work, that will be obvious, and they can immediately investigate.

galund: @TheDoubleK reason for info not error, is that the actual problematic string is going to get logged right there and then (i.e. on the next line in the logs), so this is just informative. It's not really an error, because in the world of text strings that we might pass to the logger, infinitely many of them may be unusable as a format string. (Which makes this whole way of organising the logging a bit of an oddity, but hey :) )

PR #291, opened 2017-01-09
137170159-Notify-integration

benvand:
#### Directory reshuffle
* Move `dmutils/email.py` > `dmutils/email/dm_mandrill.py`
* Create `dm_utils/email/notify.py`
* Allow importing `dm_mandrill.send_email` from `dm_utils.email` for backwards compatibility

#### App fixes, reshuffles, refactors, tweaks
* Add `notifications-python-client.git` to `requirements.txt`
* Add `norecursedirs` to `setup.cfg` to stop py.test running tests for env packages
* Remove unnecessary method `token_created_before_password` and corresponding tests
* Move `hash_string` to new `dmutils/email/helpers.py` to use in Mandrill and Notify 
* New `stringify_dict` method in `dmutils/formats.py` to format dicts for error messages

#### Email stuff and Notify integration
* Normalise exceptions with new `dmutils/email/exceptions.py`
* Adds class `DMNotifyClient`
  * `get_all_notifications` will get a list of all the notifications from the api
  * `get_reference` hashes the email address, template id and template context to provide a reference
  * `send_email` is the method used for actually sending the api request
  * Caching sent references allows us to not send duplicate mails

#### Notify Integration Tests and Mandrill Tests
* Mirrored new directory structure in `tests` directory
* Normalise string declarations in `test_dm_mandrill.py`
* New fixture for a response from `notifications_python_client.NotificationsAPIClient.get_all_notifications`
* Fixtures for the client, an exception, a `send_email_notification` response
* Tests for functionality (Mocking methods on `notifications_python_client.NotificationsAPIClient`)

### DIFF
--- a/dmutils/email/__init__.py
+++ b/dmutils/email/__init__.py
@@ -1,0 +1,2 @@
+"""Allow accessing send email from email. Backwards compatibility."""
+from dmutils.email.dm_mandrill import send_email
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+"""Digital Marketplace Notify integration."""
+from collections import OrderedDict
+
+from flask import current_app
+from notifications_python_client import NotificationsAPIClient
+from notifications_python_client.errors import HTTPError
+
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+
+
+class DMNotifyClient(object):
+    """Digital Marketplace wrapper around the Notify python client."""
+
+    _client_class = NotificationsAPIClient
+    _sent_references_cache = None
+
+    def __init__(
+            self,
+            govuk_notify_api_key,
+            govuk_notify_base_url='https://api.notifications.service.gov.uk',
+            logger=None
+    ):
+        """Set up logging and mail client."""
+        self.logger = logger or current_app.logger
+        self.client = self._client_class(govuk_notify_api_key, govuk_notify_base_url)
+
+    def get_all_notifications(self, **kwargs):
+        """Wrapper for notifications_python_client.notifications.NotificationsAPIClient::get_all_notifications"""
+        return self.client.get_all_notifications(**kwargs)['notifications']
+
+    def get_delivered_notifications(self, **kwargs):
+        """Wrapper for notifications_python_client.notifications.NotificationsAPIClient::get_all_notifications"""
+        return self.get_all_notifications(status='delivered', **kwargs)
+
+    def get_delivered_references(self, invalidate_cache=False):
+        """Get the references of all notifications that have already been delivered."""
+        if invalidate_cache:
+            self._sent_references_cache = None
+        if self._sent_references_cache is None:
+            self._sent_references_cache = set(i['reference'] for i in self.get_delivered_notifications())
+        return self._sent_references_cache
+
+    def _update_cache(self, reference):
+        """If the cache has been instantiated then cache the new reference."""
+        if self._sent_references_cache is not None:
+            self._sent_references_cache.update([reference])
+
+    def has_been_sent(self, reference):
+        """Checks for a matching reference in our list of delivered references."""
+        return reference in self.get_delivered_references()
+
+    @staticmethod
+    def get_reference(email_address, template_id, personalisation=None, **kwargs):
+        """
+        Method to return the standard reference given the variables the email is sent with.
+
+        :param email_address: Emails recipient
+        :param template_id: Emails template ID on Notify
+        :param personalisation: Template parameters
+        :param kwargs: Extra data passed to the reference eg. {'notes': 'Manual resend'}
+        :return: Hashed string 'reference' to be passed to client.send_email_notification or self.send_email
+        """
+        personalisation_string = ','.join(list(map(str, personalisation.values()))) if personalisation else ''
+        details_string = '|'.join([email_address, template_id, personalisation_string])
+        return hash_string(details_string)
+
+    @staticmethod
+    def get_error_message(email_address, error):
+        """Format a logical error message from the error response."""
+        messages = []
+        message_prefix = 'Error sending message to {email_address}: '.format(email_address=email_address)
+        message_string = '{status_code} {error}: {message}'
+
+        for message in error.message:
+            format_kwargs = {
+                'status_code': error.status_code,
+                'error': message['error'],
+                'message': message['message'],
+            }
+            messages.append(message_string.format(**format_kwargs))
+        return message_prefix + ', '.join(messages)
+
+    def send_email(self, email_address, template_id, personalisation=None, allow_resend=True):
+        """
+        Method to send an email using the Notify api.
+
+        :param email_address: String email address for recipient
+        :param template_id: Template accessible on the Notify account whose  api_key you instantiated the class with
+        :param personalisation: The template variables, dict
+        :param allow_resend: if False instantiate the delivered reference cache and ensure we are not sending duplicates
+        :return: response from the api. For more information see https://github.com/alphagov/notifications-python-client
+        """
+        reference = self.get_reference(email_address, template_id, personalisation)
+        if not allow_resend and self.has_been_sent(reference):
+            self.logger.info("Email {template_id} has already been sent to {email_address} with Notify", extra=dict(
+                email_address=hash_string(email_address),
+                template_id=template_id
+            ))
+            return
+        try:
+            response = self.client.send_email_notification(
+                email_address,
+                template_id,
+                personalisation=personalisation,
+                reference=self.get_reference(email_address, template_id, personalisation)
+            )
+        except HTTPError as e:
+            self.logger.error(self.get_error_message(hash_string(email_address), e))
+            raise EmailError(str(e))
+        self._update_cache(reference)
+        self.logger.info("Sent {email_address} template email {template_id} with Notify", extra=dict(
+            email_address=hash_string(email_address),
+            template_id=template_id
+        ))
+        return response
--- a/dmutils/email/exceptions.py
+++ b/dmutils/email/exceptions.py
@@ -1,0 +1,6 @@
+# -*- coding: utf-8 -*-
+"""Digital Marketplace normalised exception classes."""
+
+
+class EmailError(Exception):
+    pass
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,4 @@ Flask-Script==2.0.5
 workdays==1.4
 unicodecsv==0.14.1
 cryptography==1.6
+git+https://github.com/alphagov/notifications-python-client.git@4.0.0#egg=notifications-python-client==4.0.0
--- a/dmutils/email/dm_mandrill.py
+++ b/dmutils/email/dm_mandrill.py
@@ -1,26 +1,23 @@
+# -*- coding: utf-8 -*-
+"""Digital Marketplace Mandrill integration."""
 import base64
-import hashlib
-import six
-import struct
 import json
+import struct
 from datetime import datetime
 
+import itsdangerous
+import six
+from cryptography import fernet
 from flask import current_app
 from flask._compat import string_types
 
 from mandrill import Mandrill, Error
-import itsdangerous
-from cryptography import fernet
-
-
-from .formats import DATETIME_FORMAT
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.formats import DATETIME_FORMAT
 
 ONE_DAY_IN_SECONDS = 86400
 SEVEN_DAYS_IN_SECONDS = 604800
-
-
-class MandrillException(Exception):
-    pass
 
 
 def send_email(to_email_addresses, email_body, api_key, subject, from_email, from_name, tags, reply_to=None,
@@ -59,7 +56,7 @@ def send_email(to_email_addresses, email_body, api_key, subject, from_email, fro
     except Error as e:
         # Mandrill errors are thrown as exceptions
         logger.error("Failed to send an email: {error}", extra={'error': e})
-        raise MandrillException(e)
+        raise EmailError(e)
 
     logger.info("Sent {tags} response: id={id}, email={email_hash}",
                 extra={'tags': tags, 'id': result[0]['_id'], 'email_hash': hash_string(result[0]['email'])})
@@ -178,14 +175,9 @@ def _parse_fernet_timestamp(ciphertext):
     return datetime.utcfromtimestamp(epoch_timestamp)
 
 
-def hash_string(string):
-    m = hashlib.sha256(string.encode('utf-8'))
-    return base64.urlsafe_b64encode(m.digest())
-
-
 def decode_password_reset_token(token, data_api_client):
     try:
-        decoded, timestamp = decode_token(
+        decoded, token_timestamp = decode_token(
             token,
             current_app.config["SECRET_KEY"],
             current_app.config["RESET_PASSWORD_SALT"],
@@ -201,10 +193,8 @@ def decode_password_reset_token(token, data_api_client):
         DATETIME_FORMAT
     )
 
-    if token_created_before_password_last_changed(
-            timestamp,
-            user_last_changed_password_at
-    ):
+    # If the token was created before the last password change
+    if token_timestamp < user_last_changed_password_at:
         current_app.logger.info("Error changing password: Token generated earlier than password was last changed.")
         return {'error': 'token_invalid'}
 
@@ -216,7 +206,7 @@ def decode_password_reset_token(token, data_api_client):
 
 def decode_invitation_token(encoded_token):
     try:
-        token, timestamp = decode_token(
+        token, _ = decode_token(
             encoded_token,
             current_app.config['SHARED_EMAIL_KEY'],
             current_app.config['INVITE_EMAIL_SALT'],
@@ -227,7 +217,3 @@ def decode_invitation_token(encoded_token):
         current_app.logger.info("Invitation reset attempt with expired token. error {error}",
                                 extra={'error': six.text_type(e)})
         return None
-
-
-def token_created_before_password_last_changed(token_timestamp, user_timestamp):
-    return token_timestamp < user_timestamp
--- a/dmutils/email/helpers.py
+++ b/dmutils/email/helpers.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+"""Email helpers."""
+import base64
+import hashlib
+
+
+def hash_string(string):
+    """Hash a given string."""
+    m = hashlib.sha256(string.encode('utf-8'))
+    return base64.urlsafe_b64encode(m.digest())
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [pep8]
 max-line-length = 120
+
+[pytest]
+norecursedirs = venv venv3 src
--- a/tests/email/fixtures/notifications__get_all_notifications__response.json
+++ b/tests/email/fixtures/notifications__get_all_notifications__response.json
@@ -1,0 +1,201 @@
+{
+  "links": {
+    "current": "https://api.notifications.service.gov.uk/v2/notifications",
+    "next": "https://api.notifications.service.gov.uk/v2/notifications?older_than=ac0973cd-3d1c-446d-bec5-665347beaf53"
+  },
+  "notifications": [
+    {
+      "body": "05/01/2017\n\nDear supplier, \n\nBen's Awesome Company\u2019s application to Digital Outcomes and Specialists 2 was successful.\n\nYour results were: \n\u2022 Digital outcomes - yes\nno\nmaybe\nI don't know\n\nWe\u2019ve notified all applicants about their results. A 10-day 'standstill period' will start on 17 January 2017 and end at 11.59pm (GMT) on 26 January 2017. We expect the framework to be awarded on 27 January 2017 and services to go live on the Digital Marketplace in February 2016. \n\nYou must sign and return your framework agreement before you can sell services on the Digital Marketplace. \n\nYour framework agreement is ready to sign at: \nhttps://www.digitalmarketplace.service.gov.uk/suppliers/frameworks/digital-outcomes-and-specialists-2/agreement\n\n\nWe\u2019ll notify you when the standstill period is over. Until then, you shouldn\u2019t make any public announcements about the result of your application.\n\nYours faithfully,\nAlison Lisett\nHead of Sourcing Delivery\nCrown Commercial Service",
+      "completed_at": "2017-01-05T16:47:27.278075Z",
+      "created_at": "2017-01-05T16:47:26.042704Z",
+      "email_address": "ben.vandersteen@digital.cabinet-office.gov.uk",
+      "id": "85475586-e90f-4064-8720-5c9cc5850589",
+      "line_1": "None",
+      "line_2": "None",
+      "line_3": "None",
+      "line_4": "None",
+      "line_5": "None",
+      "line_6": "None",
+      "phone_number": "None",
+      "postcode": "None",
+      "reference": "X-zrZv_IbzjZUnhsbWlsecLbwjndTpG0ZynXOif7V-k=",
+      "sent_at": "2017-01-05T16:47:26.474093Z",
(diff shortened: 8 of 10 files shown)


dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
 
 def token_created_before_password_last_changed(token_timestamp, user_timestamp):
     return token_timestamp < user_timestamp
+
+
+class DMNotifyClient(object):
```
TheDoubleK: Was there some discussion about the best place for this to live?  It seems a bit odd having this class in the `email.py` file alongside all the Mandrill stuff that isn't in its own `DMMandrill...` class, feels like this could be a separate thing altogether maybe.

tests/test_email.py
```diff
@@ -322,8 +355,123 @@ def test_decode_invitation_token_does_not_work_if_token_expired(email_app):
...
+    def setup(self):
+        """Set up some standard attributes for tests."""
+        self.email_address = 'example@example.com'
+        self.template_id = 'my-cool-example-template'
+        self.personalisation = OrderedDict([
+            ('foo', 'foo_p13n'),
+            ('bar', 'bar_p13n'),
+            ('baz', 'baz_p13n'),
+        ])
+        self.time = datetime.now()
```
TheDoubleK: I don't think this `self.time` is used anywhere.

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+        if time is None and set_time:
+            time = datetime.now().strftime(DATETIME_FORMAT)
+        details_dict = OrderedDict([
+            ('email_address', email_address),
+            ('template_id', template_id),
+            ('personalisation', personalisation),
+            ('time', time),
+        ])
+        details_dict.update([(k, v) for k, v in kwargs.items()])
+        return json.dumps(details_dict)
```
TheDoubleK: The whole dict seems like a big thing for a reference - will Notify save it all as a big multiline string? Do they have a limit on size of a reference?

tests/test_email.py
```diff
@@ -322,8 +355,123 @@ def test_decode_invitation_token_does_not_work_if_token_expired(email_app):
...
+                reference=''.join([
+                    '{"email_address": "example@example.com", "template_id": "my-cool-example-template", ',
+                    '"personalisation": null, "time": "2016-01-01T12:00:00.000000Z"}'
+                ]),
+            )
+
+    def test_custom_reference(self, dm_notify_client, dm_notify_example_success):
+        """Test we can override the auto reference."""
+        with mock.patch('notifications_python_client.NotificationsAPIClient.send_email_notification') as email_mock:
+            dm_notify_example_success.update(reference='')
```
TheDoubleK: I don't think we need to update the mock response at all here if we're not checking it anywhere?

dmutils/email.py
allait: Yeah, I'm not sure why we're storing this data in the reference at all. Some of it is personal information, and I don't really know how Notify handles references, but it seems to me that the primary use for them is to check whether a specific message was already sent. So unless I'm missing a use case where we need to restore information from the reference because it's not available anywhere else then I think we can concatenate and hash the fields instead of sending them as JSON. So something like:

```python
return sha('|'.join([email_address, template_id, ...]))`

File: dmutils/email.py [hunk 5]
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
 
 def token_created_before_password_last_changed(token_timestamp, user_timestamp):
     return token_timestamp < user_timestamp
+
+
+class DMNotifyClient(object):
+    """Digital Marketplace wrapper around the Notify python client."""
+
+    client_class = NotificationsAPIClient
```
allait: Why do we need to do this? It looks like we can use `NotificationsAPIClient` directly?

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+        try:
+            response = self.client.send_email_notification(
+                email_address,
+                template_id,
+                personalisation=personalisation,
+                reference=reference
+            )
+        except HTTPError as e:
+            self.logger.error(email_address, self.get_error_message(e))
+            raise e
```
allait: We have our own `HTTPError`, which is raised from API clients, so I think there's a risk of them getting confusing in the views. I'd suggest we make our own exception class, possibly grouping it together with the existing Mandrill exception.

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+        :param email_address: Emails recipient
+        :param template_id: Emails template ID on Notify
+        :param personalisation: Template parameters
+        :param time: Generally not wise to set this. Spoofs time that the email was sent.
+        :param set_time: Flag, whether or not to set the time in the reference.
+        :param kwargs: Extra data passed to the reference eg. {'notes': 'Manual resend'}
+        :return: String json 'reference' to be passed to client.send_email_notification or self.send_email
+        """
+        if time is None and set_time:
+            time = datetime.now().strftime(DATETIME_FORMAT)
```
allait: We should always use `.utcnow()` for things like this

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+        return message_prefix + ', '.join(messages)
+
+    def send_email(self, email_address, template_id, personalisation=None, reference=None):
+        """
+        Method to send an email using the Notify api.
+
+        :param email_address: String email address for recipient
+        :param template_id: Template accessible on the Notify account whose  api_key you instantiated the class with
+        :param personalisation: The template variables, dict
+        :param reference: If you want to manually specify a reference you may, not recommended.
```
allait: If we don't have a use-case and it's not recommended already, do we need this at all?

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+        for message in error.message:
+            format_kwargs = {
+                'status_code': error.status_code,
+                'error': message['error'],
+                'message': message['message'],
+            }
+            messages.append(message_string.format(**format_kwargs))
+        return message_prefix + ', '.join(messages)
+
+    def send_email(self, email_address, template_id, personalisation=None, reference=None):
```
allait: Should we have a similar method that would check if the message has been sent already? It feels like the reference methods are leading up to that.

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+
+        :param email_address: Emails recipient
+        :param template_id: Emails template ID on Notify
+        :param personalisation: Template parameters
+        :param time: Generally not wise to set this. Spoofs time that the email was sent.
+        :param set_time: Flag, whether or not to set the time in the reference.
+        :param kwargs: Extra data passed to the reference eg. {'notes': 'Manual resend'}
+        :return: String json 'reference' to be passed to client.send_email_notification or self.send_email
+        """
+        if time is None and set_time:
```
allait: Why do we want time in our reference at all? I'd expect them to be part of the Notify API response anyway, and it feels like they make it very hard to check if a particular message has already been sent, since you can't generate the same reference repeatedly for the same message.

dmutils/email.py
```diff
@@ -1,17 +1,18 @@
 import base64
 import hashlib
-import six
-import struct
 import json
+import struct
+from collections import OrderedDict
 from datetime import datetime
 
+import itsdangerous
```
benvand: I've taken to working off of http://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#Imports_formatting

Std lib
3rd party
Our libraries (I usually an extra section 'our' libraries here)
then local imports

All in depth then alphabetical order

```
import base64
import hashlib
import json
import struct
from collections import OrderedDict
from datetime import datetime

import itsdangerous
import six
from cryptography import fernet
from flask import current_app
from flask._compat import string_types
from mandrill import Mandrill, Error

from notifications_python_client import NotificationsAPIClient
from notifications_python_client.errors import HTTPError

from .formats import DATETIME_FORMAT
```

I'm easy if there's another convention, tbh I do this predominantly because my editor does it for me :)

benvand: I've also come across codebases that separate out the web framework specific imports into their own thing but this is probably overkill here

```
import base64
import hashlib
import json
import struct
from collections import OrderedDict
from datetime import datetime

import itsdangerous
import six
from cryptography import fernet

from flask import current_app
from flask._compat import string_types
from mandrill import Mandrill, Error

from notifications_python_client import NotificationsAPIClient
from notifications_python_client.errors import HTTPError

from .formats import DATETIME_FORMAT
```

dmutils/email.py
benvand: @allait I'd kind of prefer for this to have a separate `notify.py` file. What do you think?

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
+
+    @staticmethod
+    def get_reference(email_address, template_id, personalisation, time=None, set_time=True, **kwargs):
+        """
+        Method to return the standard reference given the variables the email is sent with.
+
+        :param email_address: Emails recipient
+        :param template_id: Emails template ID on Notify
+        :param personalisation: Template parameters
+        :param time: Generally not wise to set this. Spoofs time that the email was sent.
```
benvand: Very true, I was thinking if we want to do this for some sort of ordering or something but it's not necessary yet (or possibly ever). Removed.

As an aside it could still be overwritten by passing in a `time` kwarg

dmutils/email.py
benvand: So the problem with this is that we have no way of getting the `personalisation` context dict back from the notify api unless we push it into the reference. Therefore unless we store it we can't know what context the email was sent with :/

The `email_address`, `template_id` and `personalisation` combined denote a unique instance of an email.
We may want to send the same user the same template with 2 different contexts but we have no way of knowing which is which unless we store the personalisation in the reference.

For instance these are 2 different emails, with potentially vastly different content:
```
send_email('example@example.com', '1234-12341234-12341234-1234', {'foo': 'bar'})
send_email('example@example.com', '1234-12341234-12341234-1234', {'foo': 'baz'})
```
But we can't tell which is which without the personalisation.

#### Could probably be considered a premature optimisation

requirements.txt
```diff
@@ -14,3 +14,4 @@ Flask-Script==2.0.5
 workdays==1.4
 unicodecsv==0.14.1
 cryptography==1.6
+git+https://github.com/alphagov/notifications-python-client.git@4.0.0#egg=notifications-python-client==4.0.0
```
benvand: They don't support 2.* yet so it's not available on `pip2` :(
@leohemsted is looking into/ floating the idea of 2.7 support soon apparently

tests/test_email.py
benvand: :upside_down_face: 
Good catch

tests/test_email.py
benvand: True

dmutils/email.py
benvand: In my experience this paradigm makes reuse easier. What the `DMNotifyClient` class actually does is is instantiate a client and pass it arguments to send an email. 
I was going to just inherit the `NotificationsAPIClient` class but that'd be a bit fragile if there were unexpected updates.

It would be possible to do something like this:
```
class EmailerBase(object):

    client_class = None
    client_args = None


    def __init__(self):
        self.client = client_class(*client_args)
        self.logger = self.get_logger()

    def get_logger(self):
        raise NotImplementedError('Class requires get_logger')

    @staticmethod
    def decode_reference(ref, datetime_format=None):...

    @staticmethod
    def get_error_message(email_address, error):...

    def get_send_method(self):...

    def get_send_args(self):...

    def send_email(self, email_address):

        reference = self.get_reference(email_address, *self.get_send_args)
        try:
            response = self.get_send_method(
                self.get_send_args
            )
        except HTTPError as e:
            self.logger.error(self.get_error_message(email_address, e))
            raise e
        self.logger.info("Sent {email_address} template email {template_id} with Notify".format(
            email_address=email_address,
            template_id=template_id
        ))
        return response


class DMNotifyClient(EmailerBase):
    client_class = NotificationsAPIClient
    client_args = (
        govuk_notify_api_key,
        'https://api.notifications.service.gov.uk'
    )

    def get_send_method(self):
        return self.client.send_email_notification
    
    def get_send_args(self):
        return (template_id, personalisation=None, reference=None)


class MandrillClient(EmailerBase):
    client_class = Mandrill
    client_args = api_key

    def get_send_method(self):
        return self.client.messages.send

    def get_send_args(self): ...
        # get all the mandrill stuff

```

dmutils/email.py
benvand: To check for the same email you need to specify the email address, template id and the 'personalisation'. Adding time ensures that the reference is unique. In my view the reference should really represent a 'send' instance rather than the content of the email and the most useful way I could think of to make a unique reference was just dumping in the attributes we used to create the email.

dmutils/email.py
benvand: True

dmutils/email.py
```diff
@@ -231,3 +232,95 @@ def decode_invitation_token(encoded_token):
...
 def token_created_before_password_last_changed(token_timestamp, user_timestamp):
     return token_timestamp < user_timestamp
+
+
+class DMNotifyClient(object):
+    """Digital Marketplace wrapper around the Notify python client."""
+
+    client_class = NotificationsAPIClient
+
+    def __init__(self, api_key, base_url='https://api.notifications.service.gov.uk', logger=None):
```
benvand: Done

benvand:
Questions for tomorrow:
Hashing email. How to get this back?
`client_class` a paradigm worth considering?
`personalisation` and making sure we can find a unique instance of an email.
Really; what should the `reference` be?

dmutils/email.py
allait: I like `email.py` too. We could make it clear that this file contains two separate approaches, but I think it makes sense to keep them close.

requirements.txt
allait: Does it mean that it won't work on python2? Or is this only about official support?

dmutils/email.py
allait: Fair enough. I somehow missed the fact that it's a class attribute, I got the impression that this was a global variable. Let's make it `_client_class` though, to make it clear that it's not supposed to be used externally.

requirements.txt
benvand: Only official support. It works fine on py2. However I think what it also means we can't _assume_ py2 support going forward.
I can't see any point in implementing backwards incompatibility unless they absolutely have to but it means we need to be careful until the officially support py2.

dmutils/email.py
TheDoubleK: Ah sweet, that looks fine, I just didn't quite spot the pattern :)

benvand:
For posterity:
> Hashing email. How to get this back?
Dont need to
> client_class a paradigm worth considering?
Paradigm is fine but indicate the attribute is private with `_`
> Really; what should the reference be?
Ref is `email_address`, `template` and `context`
> notify.py or email.py?
Neither; subdir

Also: hash emails and reference.

requirements.txt
leohemsted: @allait @TheDoubleK @benvand we use it in python 2.7 in our load tests, which are fairly informal, but do work. the only steps we need to take to get it on pypi is run tests against it and then update the meta tags to support it. Zero code change at all is anticipated, and after we're released I don't plan on dropping support for [65% of the python userbase](https://twitter.com/vlasovskikh/status/801720613312364544?ref_src=twsrc%5Etfw) 😬

[the ticket's in the backlog](https://www.pivotaltracker.com/story/show/137108317) and i do plan on working on that Soon™

dmutils/email/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Allow accessing send email from email. Backwards compatibility."""
+from dmutils.email.dm_mandrill import send_email
```
allait: I'd do the same for the notify client, to keep imports the same if it ends up being the only email client in the future

dmutils/email/exceptions.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+"""Digital Marketplace normalised exception classes."""
+
+
+class DMEmailApiError(Exception):
```
allait: I'd call this `EmailError` and I don't think we need separate classes form Mandrill and Notify. I don't think they'll overlap in code and having the same exception makes it easier to switch.

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def get_delivered_notifications(self, **kwargs):
+        """Wrapper for notifications_python_client.notifications.NotificationsAPIClient::get_all_notifications"""
+        return self.get_all_notifications(status='delivered', **kwargs)
+
+    def get_delivered_references(self, invalidate_cache=False):
+        """Get the references of all notifications that have already been delivered."""
+        if invalidate_cache:
+            self._sent_references_cache = None
+        if self._sent_references_cache is None:
```
allait: I think this could be `if invalidate_cache:` and then we don't need the two lines above

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+            self._sent_references_cache = [i['reference'] for i in self.get_delivered_notifications()]
```
allait: I think we can make this a `set()`

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+        :param personalisation: Template parameters
+        :param kwargs: Extra data passed to the reference eg. {'notes': 'Manual resend'}
+        :return: Hashed string 'reference' to be passed to client.send_email_notification or self.send_email
+        """
+        details_dict = OrderedDict([
+            ('email_address', email_address),
+            ('template_id', template_id),
+            ('personalisation', ','.join(stringify_dict(personalisation))),
+        ])
+        details_string = '|'.join(stringify_dict(details_dict))
```
allait: I don't think we care about the keys, only the values, so we could  `'|'.join([email_adress, template_id, personalisation])` and get rid of the dictionary, OrderedDict import and the `stringify_dict` helper altogether.

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+        """Format a logical error message from the error response."""
+        messages = []
+        message_prefix = 'Error sending message to {email_address}: '.format(email_address=email_address)
+        message_string = '{status_code} {error}: {message}'
+
+        for message in error.message:
+            format_kwargs = {
+                'status_code': error.status_code,
+                'error': message['error'],
+                'message': message['message'],
```
allait: Are we sure notify error messages don't contain the destination email address?

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+            }
+            messages.append(message_string.format(**format_kwargs))
+        return message_prefix + ', '.join(messages)
+
+    def send_email(self, email_address, template_id, personalisation=None, no_resend=False):
```
allait: I'd go with `resend=True` instead of `no_resend=False`, I find it usually makes it easier to reason about conditionals.

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,123 @@
...
+
+        :param email_address: String email address for recipient
+        :param template_id: Template accessible on the Notify account whose  api_key you instantiated the class with
+        :param personalisation: The template variables, dict
+        :param no_resend: Bool, instantiate the delivered reference cache and ensure we are not sending duplicates
+        :return: response from the api. For more information see https://github.com/alphagov/notifications-python-client
+        """
+        reference = self.get_reference(email_address, template_id, personalisation)
+        if no_resend and self.has_been_sent(reference):
+            self.logger.warn("Email {template_id} has already been sent to {email_address} with Notify", extra=dict(
```
allait: This feels like an `.info` rather than a warning

dmutils/email/dm_notify.py
benvand: Yup :(
This is the printed context of an error raised by Notify:
```{
  "errors": [
    {
      "error": "ValidationError",
      "message": "email_address Not a valid email address"
    },
    {
      "error": "ValidationError",
      "message": "template_id is not a valid UUID"
    }
  ],
  "status_code": 400
}```

File: dmutils/email/dm_notify.py (hunk 19 above)
Comment (benvand) [inline dmutils/email/dm_notify.py]: Fair, we'll likely still need OrderedDict to ensure ordering for the tests and so that the ref is created the same but yeah.

File: dmutils/email/dm_notify.py (hunk 21 above)
Comment (benvand) [inline dmutils/email/dm_notify.py]: true

File: dmutils/email/__init__.py (hunk 15 above)
Comment (benvand) [inline dmutils/email/__init__.py]: The only reason I didn't bother was that it should make refactoring a bit easier if we have explicit import. This is only a backwards compatibility inclusion. Not _needed_ for new code. Also we'd have to import the client not the method then it might look a bit weird getting the client for Notify and the send method for Mandrill from the same place :/

File: dmutils/email/exceptions.py (hunk 16 above)
Comment (benvand) [inline dmutils/email/exceptions.py]: > makes it easier to switch
Yeah good point actually, hadn't thought of that

File: dmutils/email/dm_notify.py (hunk 17 above)
Comment (benvand) [inline dmutils/email/dm_notify.py]: We need the second line because `_sent_references_cache` can be set to `None` if the `invalidate_cache` flag is set and if the cache has never been instantiated.

It would be possible to use an `or` statement like so:
```
if self._sent_references_cache is None or invalidate_cache:
    self._sent_references_cache  = ...
```

I chose to write it like the above because I think it's more readable. 'We've told it to invalidate the cache, so we do that, then we reset the cache.' Rather than implicitly invalidating the cache by overwriting it.

File: dmutils/email/dm_notify.py (hunk 18 above)
Comment (benvand) [inline dmutils/email/dm_notify.py]: Yup

File: dmutils/email/dm_notify.py (hunk 22 above)
Comment (benvand) [inline dmutils/email/dm_notify.py]: Hmmm, not sure about this one. We've taken the time to set the `resend=False` flag because we don't want to send duplicates. The code then tries to send a duplicate. I'd prefer visibility on that.
I'll change it if you feel strongly about it.

File: dmutils/email/dm_notify.py (hunk 20 above)
Comment (allait) [inline dmutils/email/dm_notify.py]: So we either only log status and the error or we do a replace on the message with `message['message'].replace(email_address, '***')`

File: dmutils/email/__init__.py (hunk 15 above)
Comment (allait) [inline dmutils/email/__init__.py]: Fair enough. On a related note, `dm_` in `dm_mandrill` and `dm_notify` seems a bit redundant. It's already in `dmutils`, so I'm not sure if it's adding anything

File: dmutils/email/dm_notify.py [hunk 23]
```diff
@@ -0,0 +1,117 @@
...
+        """
+        Method to return the standard reference given the variables the email is sent with.
+
+        :param email_address: Emails recipient
+        :param template_id: Emails template ID on Notify
+        :param personalisation: Template parameters
+        :param kwargs: Extra data passed to the reference eg. {'notes': 'Manual resend'}
+        :return: Hashed string 'reference' to be passed to client.send_email_notification or self.send_email
+        """
+        personalisation_string = ','.join(list(map(str, personalisation.keys()))) if personalisation else ''
```
allait: Should this be `personalisation.values()`?

dmutils/email/dm_notify.py
```diff
@@ -0,0 +1,117 @@
...
+
+        :param email_address: String email address for recipient
+        :param template_id: Template accessible on the Notify account whose  api_key you instantiated the class with
+        :param personalisation: The template variables, dict
+        :param allow_resend: if False instantiate the delivered reference cache and ensure we are not sending duplicates
+        :return: response from the api. For more information see https://github.com/alphagov/notifications-python-client
+        """
+        reference = self.get_reference(email_address, template_id, personalisation)
+        if not allow_resend and self.has_been_sent(reference):
+            self.logger.warn("Email {template_id} has already been sent to {email_address} with Notify", extra=dict(
```
allait: I think `info` since this is still an expected behaviour. We explicitly ask it not to resend because we expect the message might've already been sent, so there's nothing to be warned about.

benvand:
Alright everyone, thanks for the comments, I've addressed everything and as discussed with @allait on Friday I'm going to go ahead and merge. 👍

PR #292, opened 2017-01-16
NO-TICKET-version-bump-23.1.2

benvand:


allait:
We're loosely following http://semver.org/ and since the change is adding a whole new notify client, I'd say it deserves a minor bump to `23.2.0`.

👍 right away to avoid delaying that.

PR #293, opened 2017-01-18
NO-TICKET-fix-dependencies

benvand:
#### Refactor from `requirements.txt` to `setup.py`

Discussion document:
    https://docs.google.com/document/d/1xvGUt3lfZAAh88Jh7meg6p4jEt2smUEKcp7Wm7lZrj0/edit


TheDoubleK:
This feels like a nicer solution than having to bubble the dependency up to the apps that use utils.  It does mean that dependencies work differently here than everywhere else though...  would be good to hear from @allait on which option he prefers.

TheDoubleK:
This will need a patch version bump too.

benvand:
I've dropped a ticket on the tech debt board to refactor this in all branches but I'm thinking doing it here first is a good idea. This is the package that is most imported by our apps.

allait:
I don't think the explanation in that document is correct. `setup.py` obviously knows about the `notifications-python-client` package since it's trying to install it to begin with. The error happens because `4.0.0` package version is not in PyPI, which is where `setup.py` would look since it only has the package name.

There's a long history behind pip's decision to disable `dependency_links`, but most of it boils down to the view that libraries shouldn't control where their dependencies are coming from, since this creates conflicts (eg your app depends on libraries `a` and `b`, both of which depend on `c`, but they pull it from different places, so it's not guaranteed to be the same). 

Now, that's not really an issue for us, since we control both the library and the app and we don't depend on conflicting things. However, since `pip` has dropped `dependency_links` support in the past, we had to adopt the approach of copying our dependencies to the app level, as that's the place those conflicts can be explicitly resolved.

We don't need to move everything to `setup.py` in order to get `dependency_links` - we can still parse them out of `requirements.txt` like we used to https://github.com/alphagov/digitalmarketplace-utils/commit/508f55e (this was later removed https://github.com/alphagov/digitalmarketplace-utils/commit/7efa01d after pip has deprecated the support). I'd say the reason to keep `requirements.txt` around is that that's what apps do and it makes it easier to keep Makefiles / travis config / build pipelines similar.

Another problem I see with this though is that while we can change how travis installs dependencies, we don't do the same during the actual deployment. Elastic Beanstalk runs `requirements.txt` on its own, and chances are it doesn't include this flag. This doesn't apply to scripts, but it does mean that this approach probably wouldn't work anywhere else. And it probably won't work if we ever decide to move scripts to run from PaaS using CloudFoundry tasks.

So it feels like we're creating a few problems while trying to remove a duplicate `requirements.txt` line. While I don't necessarily agree with pip's design choice, it does feel like fighting against it is more trouble than it's worth.

allait:
And it looks like Travis is running an older version of pip, which still marks the `process-dependency-links` flag as deprecated.

benvand:
> `setup.py` obviously knows about the `notifications-python-client` package since it's trying to install it to begin with

The only thing it _obviously_ has is a string that it passes to Pypi.
We can't get this library from Pypi because it currently doesn't support 2.x.

> we can still parse them out of requirements.txt like we used to 

Have a look at the demos I set up. There's one that shows how the old way fails. It fails even with `--process-dependency-links` all over the place.

```
git clone https://github.com/benvand/a_old_test.git
cd a_old_test
virtualenv venv
source venv/bin/activate
pip install --process-dependency-links -r requirements_for_dev.txt
```

> Another problem I see ... this approach probably wouldn't work anywhere else

It's not our job to write hacks to integrate with other platforms. We can change EB and we can choose not to migrate to PaaS/CloudFoundry if it isn't suitable for our needs.

> it looks like Travis is running an older version of pip, which still marks the process-dependency-links flag as deprecated

The deprecation warning is still in place for `--process-dependency-links` but since 2014 all the team have done is _add_ functionality to support it https://pip.pypa.io/en/stable/news/

> I don't think the explanation in that document is correct

I disagree. I've done a lot of research putting together all the tests, documentation and the PR

http://stackoverflow.com/questions/14399534/how-can-i-reference-requirements-txt-for-the-install-requires-kwarg-in-setuptool#comment34497476_16624700

benvand:
Not having a dig, I just didn't put it all together with the sole purpose of misleading people. :)

allait:
Sorry, I probably should've been more clear. This is the part of the doc that I disagree with:

> But anything in the requirements.txt which also had a git repo dependency would be handled with python setup.py install, parse_requirements wouldn’t work, str(r.req) would return ’None’ and finallythe No matching distribution error would be raised.
In short anything in a requirements file is downloaded and setup.py install is run on it.

For both the scripts repo and the examples you've set up running `pip install` would trigger `setup.py` for all dependencies. It never uses any of the nested `requirements.txt` files directly (only whatever was parsed into `setup.py`), and since the error explicitly mentions failing to install the package my conclusion is that it does indeed parse correctly. The problem is that the referenced packaged version is not available on PyPI.

Which is where parsing `dependency_links` comes in.

> Have a look at the demos I set up. There's one that shows how the old way fails. It fails even with --process-dependency-links all over the place.

I couldn't find this example in the doc, but for `--process-dependency-links` to work you'd need to populate `dependency_links` in setup.py, which we don't do at the moment. It's easy enough to add however (again, we've done this before https://github.com/alphagov/digitalmarketplace-utils/commit/7efa01d) without having to move everything to `setup.py`.

> I've done a lot of research putting together all the tests, documentation and the PR

I didn't mean to question this at all. But I'm not coming into this from the outside either. We've spent time investigating these things before, so I'm trying to bring context for why these decision have been made.

> It's not our job to write hacks to integrate with other platforms. We can change EB and we can choose not to migrate to PaaS/CloudFoundry if it isn't suitable for our needs.

While it's true that all of those things can be changed, I think we're losing perspective at this point. The problem we have right now is that the current approach requires us to explicitly list non-PyPI dependencies in the top-level (application or script) `requirements.txt`. I don't think that's a reason to throw away our entire current and future app and test infrastructure. And as it stands right now, the proposed solution would require us to hack our way on Travis, EB and PaaS.

For what it's worth – locking all dependencies (including versions and source location) on the app level is not a rare approach. That's what ruby does with `Gemfile.lock` and that's [essentially the reason requirements.txt exists](https://caremad.io/posts/2013/07/setup-vs-requirement/) (which also explains Donald's position on not parsing requirements.txt into setup.py). So the python approach would be to drop git URLs from `dmutils` entirely, except for testing purposes. But it adds a lot of overhead and doesn't give us much in this case, so we compromise.

benvand:
With your solution we're attempting to support (and have, by the looks of it, already [failed](https://github.com/alphagov/digitalmarketplace-utils/commit/7efa01d) multiple [times](https://docs.google.com/document/d/1xvGUt3lfZAAh88Jh7meg6p4jEt2smUEKcp7Wm7lZrj0/edit) at supporting) 2 different install methods. 
`pip install -r requirements.txt` and
`python setup.py install`

We're relying on a paradigm that we've been [warned](http://stackoverflow.com/questions/14399534/how-can-i-reference-requirements-txt-for-the-install-requires-kwarg-in-setuptool#comment34497476_16624700) not to use by a pip maintainer. 

We're leaving obsolete requirements files in place just to support multiple dependency gathering methods.

Or we're going to ship a module that cannot fulfil its own dependencies?


For the sake of changing a few install scripts.

I just cannot see any benefit over continuing to support requirements.txt by further hacking the setup.py when we can just drop the dependencies in setup.py and be done with it. It's already resulted in unexpected behaviour more than once. 

Doing this:
```
dependency_links = filter(		
     lambda x: x != 'None',		
     [str(r.link if hasattr(r, 'link') else r.url) for r in requirements]		
 )
```

Just further shows that we should be putting stuff in the `setup.py`. It also obsoletes one of these lines:
https://github.com/alphagov/digitalmarketplace-utils/pull/293/files#diff-354f30a63fb0907d4ad57269548329e3L6


It's just as much effort to put the `dependency_links` hack at the top of the `setup.py` and remove whichever of these lines becomes redundant:
```
 .travis.yml
  python:
    - "2.7"		    - "2.7"
    - "3.4"		    - "3.4"
  install:		  install:
 -  - pip install -r requirements_for_test.txt
 -  - pip install -e .
```

As it is to move the requirements to the `setup.py` except we improve readability, simplicity, code-cleanliness, ease of support and reduce tech-debt all at the same time.

If pip is going to use `setup.py` anyway why don't we? We don't need to support 2 ways of installing dependencies.

benvand:
> I don't think that's a reason to throw away our entire current and future app and test infrastructure

I don't get this. We're not throwing away anything.

PR #296, opened 2017-01-20
Refactor from requirements.txt to setup.py

benvand:
The purpose of this change is to ensure that this module works wheninstalled by `setup.py` and to remove redundencies in the requirements/ dependencies installation.
This module is likely to be installed by `setup.py` _within_ applications and therefore should be able to handle everything via the `setup.py` 

* Refactor from requirements.txt to setup.py
* Alphabetical order requirements
* Version bump
* Unicode sandwich `dmutils.email.helpers.hash_string` function (stops bytes leaking in py3)
* Reorder imports
* Reorder functions to not be referenced before they're declared

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -1,4 +1,4 @@
-from . import logging, config, proxy_fix, formats, request_id
+from . import config, proxy_fix, formats, request_id
 from .flask_init import init_app, init_manager
 
 import flask_featureflags
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags
 
-__version__ = '24.0.0'
+__version__ = '24.0.1'
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -1,6 +1,4 @@
-from . import config, proxy_fix, formats, request_id
+from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
-
-import flask_featureflags
 
 __version__ = '24.0.1'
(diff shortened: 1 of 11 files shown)


allait:
Looks good 👍

TheDoubleK:
This will need a version number bump too.

dmutils/__init__.py
```diff
@@ -1,6 +1,6 @@
-from . import logging, config, proxy_fix, formats, request_id
```
TheDoubleK: Dropping this import makes me nervous but I've had a quick look around and can't immediately see anywhere else referencing it, though these things can be tricky to track down.

Have you run at least one of the frontend apps with this version of utils pulled in?  Would be good to be 100% sure that nothing is expecting to find `logging` here.

dmutils/email/dm_mandrill.py
```diff
@@ -158,21 +157,21 @@ def decrypt_data(encrypted_data, secret_key, namespace, max_age_in_seconds):
...
 
-    Decryption should be attempted before using this function, as that does cryptographically strong tests on the
-    validity of the ciphertext.
-    """
-    decoded = base64.urlsafe_b64decode(ciphertext)
+    The token may have been created by a previous version of dmutils, which only supported signed unencrypted tokens.
+    To maintain backwards compatibility during rollouts, we try and decode using the old format, and if that fails
+    assume it is is a new encrypted token.
 
+    This functionality can be removed once all rollouts have completed and the longest time-to-live of the old signed
```
TheDoubleK: The old tokens are long gone, so maybe now is the time to remove this functionality, rather than move it round?

tests/conftest.py
```diff
@@ -35,3 +35,21 @@ def os_environ(request):
     request.addfinalizer(env_patch.stop)
 
     return env_patch.start()
+
+
+@pytest.fixture
```
TheDoubleK: Nice to get this into the right place :clap:

dmutils/__init__.py
benvand: Not sure how this got removed but good catch, it shouldn't be

PR #298, opened 2017-02-08
Reintroduce flask_featureflags in __init__, required by other apps

benvand:


TheDoubleK:
:+1:

TheDoubleK:
Actually, hold off!

TheDoubleK:
You could add a `#noqa` after the import - we do this elsewhere to flag to developers that the import is there for a reason, even if it looks like it isn't used.

PR #300, opened 2017-02-10
BUGFIX-notifications-added-to-pip-upgrade

benvand:
Notifications Python Client has been added to pip2 so we no longer need to fetch it from github 🎉 
* Get notifications from pip not github


leohemsted:
💯 👌

PR #301, opened 2017-02-28
140127375-nbsp-filter

benvand:
Required for https://www.pivotaltracker.com/story/show/140127375

* Add new template filter to convert spaces to nbsp
* Make available to all templates
* Add test
* Version bump

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -3,4 +3,4 @@ from .flask_init import init_app, init_manager
 
 import flask_featureflags  # noqa
 
-__version__ = '24.0.4'
+__version__ = '24.1.0'
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 import re
-from flask import Markup
+from flask import Markup, escape
 
 
 def smartjoin(input):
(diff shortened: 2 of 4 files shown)


benvand:
screen shot 2017-02-28 at 11 22 25
screen shot 2017-02-28 at 11 20 59

benvand:
screen shot 2017-02-28 at 11 20 59

dmutils/filters.py
```diff
@@ -40,3 +40,8 @@ def format_links(text):
         return formatted_text
     else:
         return text
+
+
+def nbsp(text):
+    """Replace spaces with nbsp."""
+    return Markup('&nbsp;'.join(text.split(' ')))
```
allait: We need to escape the original text before marking it as safe. This means we can't use this filter with anything that could contain HTML, but those parts should be `Markup` objects at this point anyway.

tests/test_filters.py
```diff
@@ -66,3 +66,10 @@ def test_multiple_urls():
 def test_no_links_no_change():
     text = 'There are no Greek Γ Δ Ε Ζ Η Θ Ι Κ Λ links.'
     assert format_links(text) == text
+
+
+def test_nbsp():
```
allait: Let's add a test for escaping the original text too

dmutils/__init__.py
```diff
@@ -3,4 +3,4 @@
 
 import flask_featureflags  # noqa
 
-__version__ = '24.0.4'
+__version__ = '24.0.5'
```
galund: being pedantic, I think this is supposed to be a 24.1.0 as it's a new feature?

benvand: Done 👌

PR #302, opened 2017-03-15
Don't autogenerate bucket_short_name from bucket name

allait:
Our S3 keys always start with <framework_slug>/<upload_type>/...,
where <upload_type> (eg 'documents', 'submissions', 'agreements')
maps 1:1 to the buckets in live environments.

S3 uploader used to parse upload type directly from the bucket name,
using our bucket name pattern (digitalmarketplace-<type>-<stage>-<env>).

This approach allows us to avoid passing in the directory name
explicitly, but it creates 2 issues:

1. We need to use separate buckets for development environments,
   following the same naming convention we use for preview, staging,
   and production (eg digitalmarketplace-documents-dev-dev and
   digitalmarketplace-submissions-dev-dev). Using separate buckets
   means we can't use the S3 bucket domain name as our assets host,
   since it's different for each bucket. So in order to make all
   uploaded file links work locally we'd have to set up an nginx
   config to proxy the requests to the correct bucket, which would
   complicate our local setup and make spinning up one-off environments
   on PaaS harder.

2. When mocking the S3 objects we need to set the `.bucket_short_name`
   value on the mock. This is done 26 times in the supplier-frontend.

It's possible to fix both of those problems, but setting the upload_type
manually seems like an easier approach. It separates the bucket
selection (which is primarily driven by permissions and lifecycle
concerns) from the document upload process and allows us to use a single
bucket for development, which makes it different from live environments,
but in return simplifies the local environment setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,38 @@
 
 Records breaking changes from major version bumps
 
+## 25.0.0
+
+PR: [#302](https://github.com/alphagov/digitalmarketplace-utils/pull/302)
+
+### What changed
+
+`upload_document` and `upload_service_documents` now require an explicit `upload_type`
+argument (eg 'documents' or 'submissions' for document uploads).
+
+`S3.short_bucket_name` property is removed, so there's no need to set the attribute
+on the mocks.
+
+###Example app change
+
+Old:
+```
+upload_service_documents(
+    uploader, documents_url, draft,
+    request.files, section, public=False
+)
+
+```
+
+New:
+```
+upload_service_documents(
+    uploader, 'documents', documents_url, draft,
+    request.files, section, public=False
+)
+```
+
+
 ## 24.0.0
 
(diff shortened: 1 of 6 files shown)


risicle:
This sounds sensible in general, but I can't speak of odd corners of the code that this may have odd ramifications for.

TheDoubleK:
I think this looks OK and makes sense to do it - it always feels like something is wrong when I have to change the DM_ASSETS_URL in my local config depending what feature I'm working on.

Does this also help towards https://trello.com/c/vihwQZfo/44-staging-doesn-t-have-its-own-assets-url-this-means-that-testing-some-features-on-staging-is-impossible-e-g-testing-admin-replaci or is that a completely separate issue?

allait:
@galund yeah, I don't think we need to mock the bucket name anymore.
@TheDoubleK I left a note on the trello card. We're going to keep separate buckets in staging and we do have a staging assets domain, we need to make sure the apps are configured to use it though.

If everyone's happy with this, I'm going to merge it.

PR #306, opened 2017-04-20
[#142936995]S3 boto3 port

risicle:
~~~Don't try to use this yet it won't work, I'm just posting this to get feedback on interface changes (see `CHANGELOG.md`).~~~

I don't know if I should bother replicating `CHANGELOG.md` here. Read that.

Another thing I did was start to use the `moto` library for the test mocking, which I have more confidence in than little makeshift i-assume-i-fully-understand-this-interface mocks - it's mocking aws' behaviour at an http level. Though it is of course far from perfect itself.

~~~There's still a requirements dependency on `boto` (2) - there was a resason for that - let me see if i can remember it...~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ PR: [#306](https://github.com/alphagov/digitalmarketplace-utils/pull/306)
 - `S3.save()` no longer accepts a `move_prefix` argument
 - `S3.list()`'s returned "key dict"s won't include a `last_modified` parameter at all if `load_timestamps=False` (instead of including a potentially misleading value)
 - `S3.list()` called with `load_timestamps=False` will also return its results in an arbitrary order (instead of a potentially misleading one)
-- The `s3` module *does* import & expose `ClientError`, which *can* be used as the equivalent of boto2's `S3ResponseError`, but be aware that in using it you give up any pretense at boto version independence.
+- The `s3` module *does* still expose a `S3ResponseError`, but it is a relabelled boto3 `ClientError`, in a slightly odd gesture to backwards compatibility with consumers that were using that. `ClientError` is raised by boto3 in broadly similar situations to those where boto2 would raise `S3ResponseError`.
 
 (not that I could find any external code that used it) `get_file_size_up_to_maximum` is now `get_file_size`, which is a far more sensible way of presenting the interface given the calling code is going to have to compare the result against `FILE_SIZE_LIMIT` anyway
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Records breaking changes from major version bumps
 
+
+## 27.0.0
+
+PR: [#306](https://github.com/alphagov/digitalmarketplace-utils/pull/306)
+
+### What changed
+
+(not that I could find any external code that used it) `get_file_size_up_to_maximum` is now `get_file_size`, which is a far more sensible way of presenting the interface given the calling code is going to have to compare the result against `FILE_SIZE_LIMIT` anyway
+
+
 ## 26.0.0
 
 PR: [#307](https://github.com/alphagov/digitalmarketplace-utils/pull/307)
@@ -20,6 +30,7 @@ framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud
 for lot in framework['lots']:
     ...
 ```
+
 
 ## 25.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,15 +12,70 @@ PR: [#306](https://github.com/alphagov/digitalmarketplace-utils/pull/306)
 `S3` was ported to use `boto3` and in the process changed muchly.
 
 - The constructor takes a `region` kwarg (expecting an aws region name) instead of an explicit `host` kwarg.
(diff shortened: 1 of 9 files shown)


risicle:
Curious thing: `moto` itself has a dependency on `boto` (2), but I think it just uses it to import some aws-related constants.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 
 Records breaking changes from major version bumps
 
+## 26.0.0
+
+PR: [#306](https://github.com/alphagov/digitalmarketplace-utils/pull/306)
+
+### What changed
```
risicle: Yes not having examples is just my laziness

dmutils/s3.py
```diff
@@ -1,74 +1,87 @@
...
-import boto.exception
+import boto3
 import datetime
 import mimetypes
 import logging
 from dateutil.parser import parse as parse_time
+from six import text_type
 
-from boto.exception import S3ResponseError  # noqa
+# import this and catch it if you like but doing so you give up any pretense at boto version independence.
```
risicle: `from botocore.exceptions import ClientError as S3ResponseError` is a bit of a weird lie though isn't it? It's the sort of thing that implies one way of working right up to the point that someone goes "oh it's a boto error, ok - `from boto.exception import S3ResponseError ... except S3ResponseError`, yeah I expect that works..." but I don't exactly have a better solution. I guess it's also the issue that it's not (necessarily) just `S3ResponseError` or `ClientError`, but other exceptions that leak out and apps might want to be able to catch...

dmutils/s3.py
```diff
@@ -1,74 +1,87 @@
...
+        self._resource = boto3.resource("s3", region_name=region_name)
+        self._bucket = self._resource.Bucket(bucket_name)
 
-        self.bucket_name = bucket_name
-        self.bucket = conn.get_bucket(bucket_name)
+    @property
+    def bucket_name(self):
+        return self._bucket.name
 
-    def save(self, path, file, acl='public-read', move_prefix=None, timestamp=None, download_filename=None,
```
risicle: Yup.

dmutils/s3.py
```diff
@@ -1,74 +1,87 @@
...
         if download_filename:
-            headers['Content-Disposition'] = '{}; filename="{}"'.format(
-                disposition_type, download_filename
-            ).encode('utf-8')
-        key.set_contents_from_file(
-            file,
-            headers=headers
+            extra_kwargs["ContentDisposition"] = u'{}; filename="{}"'.format(
+                disposition_type,
+                # boto/aws can't cope with unicode here, but wants the ultimate result as a `str` in py3, so doing this
```
risicle: The whole pipeline of `Content-Disposition` is a bit of a minefield from a unicode pov, right down to the user's filesystem coping with unicode... all makes me wonder if we shouldn't just `unidecode` all our suggested filenames.

TheDoubleK: Yep there are extended unicode characters in some* of our supplier names and it's caused various things to break in the past.  I think just dropping the troublesome chars at this point (i.e. for Content-Disposition headers) is totally fine.  In the past we "fixed" the data (coming from a CSV file) for our scripts instead :/

* At least one

risicle:
> "you're a bit weird"

Damn. I was using this to test whether anyone actually read it.

risicle:
When it comes to "oddly specific", the singapore government are supposed to think that it's a codebase that has flaws that the team are aware of, rather than "um, these people think sensible api design involves introducing functions like..."

```
def file_is_not_empty(file_contents):
     return not file_is_empty(file_contents)
```

But all code I write contains comments that could be interpreted as being a bit snippy, most often aimed at things I've written myself but have some measured flaws. I have far more confidence in code which has comments like that over code that stone-facedly implies that everything is perfect as far as the authors are aware.

And "just fixing" these api interfaces is way out of the scope of this story and will inevitably involve yet more... discussions.

allait:
@risicle intent isn't magic and unfortunately we don't get to choose what people reading the comments are supposed to think, so the onus is on us to make our communication as clear as possible.

I think there are a few approaches we could take in this case:
* If you feel like there's a common need to understand why these functions are here you could write a comment to explain why they came to exist and how they're used at the moment to make it easier for the next person who comes across them to make changes to them quicker or make a decision on whether they're still needed. (My 2 cents - I don't feel like this is a common need.)
* If you want to make a note on a the reasons behind a design decision you're not sure about the best place for this is a commit message attached to the diff. It's permanently attached to a specific  piece of code at a specific point in time, so it doesn't get out of sync when the code changes and it doesn't get in the way when it's not relevant to what you're trying to accomplish.
* If you want to flag this is a potential improvement then you could make a ticket for it either on the tech debt board or your team's Trello/Pivotal. As you've mentioned, fixing this interface is a separate piece of work, so it's unlikely that the next person seeing the comment is going to do it as part of an unrelated change.

risicle:
This is a blanket statement I'm going to make here and so inherently absurd, but _nobody ever reads commit comments_. They certainly don't get read when a piece of code is being read as a means to understanding how a larger system fits together. They are also not even particularly easy to _locate_ once someone has come along later and made some minor/stylistic change to the lines. If the commit in question only happened to remove some lines, _it won't even show up_ in `git annotate`. Additionally, comments present in code prompt later authors to update them/add to them to reflect the new reality in a way that they don't get prompted to explain a specific piece of their thinking when making a commit.

risicle:
Anyway, comments softened.

risicle:
I've made the PR editable, so you can always change it...

PR #308, opened 2017-04-24
Add new DMSandboxEnvironment subclassing jinja2.SandboxEnvironment

benvand:
Add new class, subclassing Sandbox and add our custom filters


idavidmcdonald:
👍

PR #309, opened 2017-05-04
Log to stdout

Wynndow:
This is for running the apps in containers on the PaaS.

If in development, write plain text logs to stdout so they can be read
and are useful.

If in a live environment, write json logs to stdout. These will be
picked up by supervisord and redirected to file, then streamed to
cloudwatch. The files are logrotated.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -19,7 +19,6 @@ logger = logging.getLogger(__name__)
 def init_app(app):
     app.config.setdefault('DM_LOG_LEVEL', 'INFO')
     app.config.setdefault('DM_APP_NAME', 'none')
-    app.config.setdefault('DM_LOG_PATH', None)
 
     @app.after_request
     def after_request(response):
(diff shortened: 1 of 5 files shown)


dmutils/logging.py
```diff
@@ -61,16 +60,12 @@ def get_handlers(app):
     standard_formatter = CustomLogFormatter(LOG_FORMAT, TIME_FORMAT)
     json_formatter = JSONFormatter(LOG_FORMAT, TIME_FORMAT)
 
-    # Log to files if the path is set, otherwise log to stderr
-    if app.config['DM_LOG_PATH']:
```
allait: Oh, I think we need to keep this for now since we still want to write two files in the existing EB environments. So before the switch we'll have DM_LOG_PATH set to empty in the PaaS environment while the EB environments use the path from the config files. 

After we shutdown EB apps we'll be able to remove this logic from utils and drop the config variable from application configs and PaaS environments.

dmutils/logging.py
```diff
@@ -61,16 +60,12 @@ def get_handlers(app):
...
-        handler = logging.FileHandler(app.config['DM_LOG_PATH'])
+    if app.config['DM_ENVIRONMENT'] in ['development', 'test']:
```
allait: I wonder if having a separate config variable in the apps for this would be more straightforward. Eg `DM_LOG_TYPE = "json"`. We can default to JSON here and then override it in development and test configs in the apps.

PR #310, opened 2017-05-16
Add `DMMailChimpClient`

idavidmcdonald:
Trello: https://trello.com/c/XOCeuOYW/254-update-mailchimp-user-list-everyday

Adds `DMMailChimpClient` class which is then used our script to update mailchimp user lists for DOS opportunities

We previously had some of this functionality in the scripts repo but we have decided to move it into utils in a more class based approach (like we do for our notify client wrapper) and added the additional functionality to subscribe emails to a mailchimp list.

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -76,3 +76,8 @@ class DMMailChimpClient(object):
                 extra={"error": str(e)}
             )
         return False
+
+    def subscribe_new_emails_to_list(self, list_id, email_addresses):
+        for email_address in email_addresses:
+            self.subscribe_email_to_list(list_id, email_address)
+        return True
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -55,7 +55,7 @@ class DMMailChimpClient(object):
             )
         return False
 
-    def subscribe_email_to_list(self, list_id, email_address):
+    def subscribe_new_email_to_list(self, list_id, email_address):
         """ Will subscribe email address to list if they do not already exist in that list else do nothing"""
         hashed_email = self.get_email_hash(email_address)
         try:
@@ -79,5 +79,5 @@ class DMMailChimpClient(object):
 
     def subscribe_new_emails_to_list(self, list_id, email_addresses):
         for email_address in email_addresses:
-            self.subscribe_email_to_list(list_id, email_address)
+            self.subscribe_new_email_to_list(list_id, email_address)
         return True
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -69,7 +69,10 @@ class DMMailChimpClient(object):
             )
         except RequestException as e:
             self.logger.error(
-                "Mailchimp failed to send campaign id '{0}'".format(campaign_id),
+                "Mailchimp failed to add user ({}) to list ({})".format(
+                    hashed_email,
+                    list_id
+                ),
                 extra={"error": str(e)}
             )
         return False
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -4,6 +4,7 @@
 from mailchimp3 import MailChimp
 from requests.exceptions import RequestException
 from hashlib import md5
+import six
 
 
 class DMMailChimpClient(object):
@@ -19,7 +20,9 @@ class DMMailChimpClient(object):
 
     @staticmethod
     def get_email_hash(email_address):
-        return md5(email_address.lower()).hexdigest()
+        """md5 hashing of lower cased emails has been chosen by mailchimp to identify email addresses"""
+        formatted_email_address = six.text_type(email_address.lower()).encode('utf-8')
+        return md5(formatted_email_address).hexdigest()
 
     def create_campaign(self, campaign_data):
         try:
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -15,7 +15,7 @@ class DMMailChimpClient(object):
         mailchimp_api_key,
         logger
     ):
-        self.client = MailChimp(mailchimp_username, mailchimp_api_key)
+        self._client = MailChimp(mailchimp_username, mailchimp_api_key)
         self.logger = logger
 
     @staticmethod
@@ -26,7 +26,7 @@ class DMMailChimpClient(object):
 
     def create_campaign(self, campaign_data):
         try:
-            campaign = self.client.campaigns.create(campaign_data)
+            campaign = self._client.campaigns.create(campaign_data)
             return campaign['id']
         except RequestException as e:
             self.logger.error(
@@ -39,7 +39,7 @@ class DMMailChimpClient(object):
 
     def set_campaign_content(self, campaign_id, content_data):
         try:
-            return self.client.campaigns.content.update(campaign_id, content_data)
+            return self._client.campaigns.content.update(campaign_id, content_data)
         except RequestException as e:
             self.logger.error(
                 "Mailchimp failed to set content for campaign id '{0}'".format(campaign_id),
@@ -49,7 +49,7 @@ class DMMailChimpClient(object):
 
     def send_campaign(self, campaign_id):
         try:
-            self.client.campaigns.actions.send(campaign_id)
+            self._client.campaigns.actions.send(campaign_id)
             return True
         except RequestException as e:
             self.logger.error(
@@ -62,7 +62,7 @@ class DMMailChimpClient(object):
         """ Will subscribe email address to list if they do not already exist in that list else do nothing"""
         hashed_email = self.get_email_hash(email_address)
         try:
-            return self.client.lists.members.create_or_update(
+            return self._client.lists.members.create_or_update(
                 list_id,
                 hashed_email,
                 {
@@ -101,5 +101,5 @@ class DMMailChimpClient(object):
         return success
 
     def get_email_addresses_from_list(self, list_id):
-        member_data = self.client.lists.members.all(list_id, get_all=True)
+        member_data = self._client.lists.members.all(list_id, get_all=True)
         return [member["email_address"] for member in member_data["members"]]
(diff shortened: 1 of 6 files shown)


dmutils/email/dm_mailchimp.py
```diff
@@ -0,0 +1,101 @@
...
+    def subscribe_new_emails_to_list(self, list_id, email_addresses):
+        success = True
+        for email_address in email_addresses:
+            if not self.subscribe_new_email_to_list(list_id, email_address):
+                success = False
+        return success
+
+    def get_email_addresses_from_list(self, list_id):
+        member_data = self.client.lists.members.all(list_id, get_all=True)
+        return [member["email_address"] for member in member_data["members"]]
```
risicle: Sooo... why make this a wrapper of the mailchimp `client` instead of a subclass? The latter has the effect of taking the MailChimp client and augmenting it. The latter adds a micromanagement layer between the code that wants to use mailchimp and the client.

dmutils/email/dm_mailchimp.py
```diff
@@ -0,0 +1,101 @@
...
+                extra={"error": str(e)}
+            )
+            return False
+
+    def subscribe_new_emails_to_list(self, list_id, email_addresses):
+        success = True
+        for email_address in email_addresses:
+            if not self.subscribe_new_email_to_list(list_id, email_address):
+                success = False
+        return success
```
risicle: `return all(self.subscribe_new_email_to_list(list_id, email_address) for email_address in email_addresses)`

dmutils/email/dm_mailchimp.py
```diff
@@ -0,0 +1,101 @@
...
+        mailchimp_api_key,
+        logger
+    ):
+        self.client = MailChimp(mailchimp_username, mailchimp_api_key)
+        self.logger = logger
+
+    @staticmethod
+    def get_email_hash(email_address):
+        email_address = six.text_type(email_address.lower()).encode('utf-8')
+        return md5(email_address).hexdigest()
```
Wynndow: md5 isn't very secure, we use sha256 here: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/email/helpers.py#L10

tests/email/test_dm_mailchimp.py
```diff
@@ -0,0 +1,191 @@
...
+            dm_mailchimp_client.client.lists.members, 'create_or_update',  autospec=True) as create_or_update:
+        response = mock.Mock()
+        response.json.return_value = {"detail": "Unexpected error."}
+        create_or_update.side_effect = RequestException("error sending", response=response)
+
+        res = dm_mailchimp_client.subscribe_new_email_to_list('list_id', 'example@example.com')
+
+        assert res is False
+        logger.error.assert_called_once_with(
+            "Mailchimp failed to add user ({}) to list ({})".format(
```
Wynndow: Can you just inline this?

dmutils/email/dm_mailchimp.py
idavidmcdonald: Mailchimp specify the md5 as that is what they use for identifying - http://developer.mailchimp.com/documentation/mailchimp/guides/manage-subscribers-with-the-mailchimp-api/#identify-subscribers

Wynndow: Fair do's!

dmutils/email/dm_mailchimp.py
```diff
@@ -0,0 +1,101 @@
...
+                list_id,
+                hashed_email,
+                {
+                    "email_address": email_address,
+                    "status_if_new": "subscribed"
+                }
+            )
+        except RequestException as e:
+            # expected error due to user email being on Mailchimp blacklist. Don't fail but log.
+            if "looks fake or invalid, please enter a real email address." in e.response.json()["detail"]:
```
idavidmcdonald: It is a bit crap that we have to do this. If we don't then our script will fail everyday (we have a production user who is blacklisted on mailchimp).

It is a bit fragile if they do change their message but that feels OK because then we will get a different type of error and be alerted by a failing script and we can investigate and up date if needbe. I assume they change their error messages infrequently.

The bad scenario would be if they started showing that message for more errors unrelated to the one we want to allow, then we would have legit errors being hidden. This again feels unlikely.

Mailchimp error docs if you hadn't seen them (in the related commit message) - http://kb.mailchimp.com/accounts/management/about-error-messages

dmutils/email/dm_mailchimp.py
risicle: This is exactly the sort of place a comment to that effect belongs then.

(In my books, every use of md5 should be accompanied with a justification)

risicle: though to be honest, the attack model we're trying to protect against here, preimage resistance, only has a theoretical attack for md5.

dmutils/email/dm_mailchimp.py
```diff
@@ -0,0 +1,101 @@
...
+        mailchimp_username,
+        mailchimp_api_key,
+        logger
+    ):
+        self.client = MailChimp(mailchimp_username, mailchimp_api_key)
+        self.logger = logger
+
+    @staticmethod
+    def get_email_hash(email_address):
+        email_address = six.text_type(email_address.lower()).encode('utf-8')
```
idavidmcdonald: Sorry, could you just clarify what you mean by 'shadowing'?

dmutils/email/dm_mailchimp.py
risicle: Again, this is what comments are for.

dmutils/email/dm_mailchimp.py
Wynndow: I just mean passing in `email_address` and then reassigning the variable immediately. Might be better using a different variable name. `encoded_email_address` or....something better.

risicle: he means that (if you make a habit of this) later on in functions if you wonder what this `email_address` thing is and look upwards to find out, on first glance it looks like it's the plain call argument, not realizing that it's been overwritten by this function call.

dmutils/email/dm_mailchimp.py
idavidmcdonald: I feel we've written a reasonably verbose commit message describing what is happening and our decision for doing so - https://github.com/alphagov/digitalmarketplace-utils/pull/310/commits/9828b20827cd8f354ba5d06c600301c54dfe9c9b

Obviously a commit message isn't inline (which has it's pro's and con's, maybe a discussion for another day) but I'd be apprehensive about adding too much more to our inline comment. Saying that, I'm open to small suggestions that won't extend it by too much more but could convey extra useful information.

dmutils/email/dm_mailchimp.py
risicle: damn why didn't I see his comment? github lag.

dmutils/email/dm_mailchimp.py
risicle: Commit messages are *so* much harder to track down. I'm concerned about the next person who needs to understand what this code's doing and that the original author has considered alternatives. Digging out all commit messages affecting the area in question is quite a task.

dmutils/email/dm_mailchimp.py
idavidmcdonald: `return all([self.subscribe_new_email_to_list(list_id, email_address) for email_address in email_addresses])` but your suggestion is nice, will implement

benvand: This wouldn't work. Because you've used a generator expression `all` will yield at the first `False` causing only part of the emails to be added.
```
foo = 0

def bar1():
    global foo
    foo +=1

def bar2():
    global foo
    foo +=1
    return False

def bar3():
    global foo
    foo +=1

all(i() for i in [bar1, bar2, bar3])
print foo
>>> 1

foo = 0
[i() for i in [bar1, bar2, bar3]]
print foo
>>> 3
```

This would work:
```
return all([self.subscribe_new_email_to_list(list_id, email_address) for email_address in email_addresses])
```
But it's inadvisable to perform side effects like this in a comprehension anyway.
If this had been written like so:
```
for email_address in email_addresses:
     if not self.subscribe_new_email_to_list(list_id, email_address):
         yield False
```
this error would have been much easier to catch. As it stands it took a bit of pseudo code before I was fully confident of the behaviour here.

benvand: Sorry we crossed on the wires while you were working on this @idavidmcdonald 
I disagree with doing this in a comprehension.
Comprehensions are meant to be used as constructors not as a complete replacement for for loops.

risicle: Damn you're right - it's side-effecty code

dmutils/email/dm_mailchimp.py
benvand: This is inline with the `DMNotify` client class and the `dm_mandrill` module. 
The class was designed as an interface between Marketplace and Mailchimp, we don;t need to expose MailChimps lower lever stuff to our devs.

dmutils/email/dm_mailchimp.py
risicle: (and this is why not to ever drop someone else's suggestion in without checking it I guess)

idavidmcdonald: Cool, will revert back to how it was.

I also found this with the top comment also supporting using a for loop and now I learnt something new about when to use/not use comprehensions. http://stackoverflow.com/questions/2849645/in-python-is-it-better-to-use-list-comprehensions-or-for-each-loops

benvand: I wish I could be that succinct when I write about code 😂

dmutils/email/dm_mailchimp.py
risicle: If so, then let's underscore `client` to `self._client` to denote that.

dmutils/email/dm_mailchimp.py
risicle: Interestingly, did any of your tests catch this when you changed it?

idavidmcdonald: Yeah tests caught it 💥

dmutils/email/dm_mailchimp.py
idavidmcdonald: Good idea @risicle

dmutils/email/dm_mailchimp.py
risicle: Yeah, so, *clearly* what I _meant_ here was: whatever you do _don't do_ `return all(self.subscribe_new_email_to_list(list_id, email_address) for email_address in email_addresses)`

I don't know how that wasn't obvious...................

PR #313, opened 2017-06-07
Remove unused imports from `dm_mandrill.py` and tests

Wynndow:
itsdangerous was still being imported even though no longer used. It was
used for encoding/decoding tokens for emails. Due to an issue with these
appearing in the logs and being easily decoded, we now encrpyt the
tokens using the cryptography library.


Wynndow:
[Remove signed token parsing code from dmutils - S](https://trello.com/c/ehARQPvD/106-remove-signed-token-parsing-code-from-dmutils---s)

PR #314, opened 2017-06-28
Add filter to preserve line breaks in HTML output

MuriloDalRi:
Only works on textbox_large elements and only keeps a maximum of two line breaks.

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -74,9 +74,7 @@ _single_newline_re = re.compile(r'(\r\n)')
 
 
 @evalcontextfilter
-def preserve_line_breaks(eval_ctx, value, question_type):
-    if question_type and question_type != 'textbox_large':
-        return value
+def preserve_line_breaks(eval_ctx, value):
 
     # Turn potential markdown objects into string
     value = u'{}'.format(escape(value))
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -68,7 +68,7 @@ def capitalize_first(maybe_text):
     return maybe_text
 
 # find repeated sequences of '\r\n\', optionally separated by a space character
-_multiple_newlines_re = re.compile(r'(\r\n *){2,}')
+_multiple_newlines_re = re.compile(r'(\r\n[ \t\f\v]*){2,}')
 # find \r\n sequences
 _single_newline_re = re.compile(r'(\r\n)')
 
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -67,7 +67,7 @@ def capitalize_first(maybe_text):
 
     return maybe_text
 
-# find repeated sequences of '\r\n\', optionally separated by a space character
+# find repeated sequences of '\r\n\', optionally separated by other non-newline whitespace space characters
 _multiple_newlines_re = re.compile(r'(\r\n[ \t\f\v]*){2,}')
 # find \r\n sequences
 _single_newline_re = re.compile(r'(\r\n)')
@@ -76,7 +76,8 @@ _single_newline_re = re.compile(r'(\r\n)')
 @evalcontextfilter
 def preserve_line_breaks(eval_ctx, value):
 
-    # Turn potential markdown objects into string
+    # `escape()` returns Markdown objects in python2
+    # We want to cast the output value back into unicode strings
     value = u'{}'.format(escape(value))
 
     # limit sequences of "\r\n\r\n ..."s to two
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -79,7 +79,7 @@ def preserve_line_breaks(eval_ctx, value, question_type):
         return value
 
     # Turn potential markdown objects into string
-    value = '{}'.format(escape(value))
+    value = u'{}'.format(escape(value))
 
     # limit sequences of "\r\n\r\n ..."s to two
     value = _multiple_newlines_re.sub(u'\r\n\r\n', value)
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -1,6 +1,5 @@
 from __future__ import unicode_literals
 import re
-from flask import Markup, escape
 from six import string_types
 from jinja2 import evalcontextfilter, Markup, escape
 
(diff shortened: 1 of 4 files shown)


dmutils/filters.py
```diff
@@ -66,3 +66,26 @@ def capitalize_first(maybe_text):
...
+# find \r\n sequences
+_single_newline_re = re.compile(r'(\r\n)')
+
+
+@evalcontextfilter
+def preserve_line_breaks(eval_ctx, value, question_type=None):
+    if question_type and question_type != 'textbox_large':
+        return value
+
+    value = '{}'.format(escape(value))
```
pcraig3: Probably good to leave a comment about what this is for.  It's not really obvious that it's turning (potential) Markdown objects back into strings.

tests/test_filters.py
```diff
@@ -118,3 +120,13 @@ def test_capitalize_first_for_non_strings():
     assert capitalize_first([{'list': 'of'}, 'things']) == [{'list': 'of'}, 'Things']
     assert capitalize_first({'this': 'thing'}) == {'this': 'thing'}
     assert capitalize_first('https://www.example.com') == 'https://www.example.com'
+
+
+def test_preserve_line_breaks():
```
pcraig3: You _could_ use `pytest.mark.parametrize` to pass in different values for `autoescape`, just so we can see that the output is the same.

tests/test_filters.py
```diff
@@ -118,3 +120,13 @@ def test_capitalize_first_for_non_strings():
...
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '\r\n') == '<br>'
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '\r\n\r\n') == '<br><br>'
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '\r\n \r\n \r\n') == '<br><br>'
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '\r\n\r\n\r\n\r\n\r\n\r\n') == '<br><br>'
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '') == ''
+    assert preserve_line_breaks(mock.Mock(autoescape=False), '\r\n<h2>') == '<br>&lt;h2&gt;'
+    assert preserve_line_breaks(mock.Mock(autoescape=False), 'You’ll be working') == 'You’ll be working'
```
pcraig3: It would be good to check a couple of other things that won't work, like 
- `"\n"` (should be `"\n"`)
- `"\r\n\r\n  \r\n"`  (should be `"<br><br>  <br>"`)

dmutils/filters.py
```diff
@@ -66,3 +66,26 @@ def capitalize_first(maybe_text):
...
     return maybe_text
+
+# find repeated sequences of '\r\n\', optionally separated by a space character
+_multiple_newlines_re = re.compile(r'(\r\n ?){2,}')
+# find \r\n sequences
+_single_newline_re = re.compile(r'(\r\n)')
+
+
+@evalcontextfilter
+def preserve_line_breaks(eval_ctx, value, question_type=None):
```
pcraig3: Final point: maybe we should take away `=None` here to stop people from being able to use this too easily.

Not sure where it's going to end up being used, but for both individual brief and individual service pages, we always have a `.type` hanging around.

dmutils/filters.py
MuriloDalRi: 👍

tests/test_filters.py
MuriloDalRi: Done.

tests/test_filters.py
MuriloDalRi: Shouldn't "\r\n\r\n \r\n" give us "<br><br>"?

If we don't account for spaces it will be possible to add much more than two line breaks continuously.

dmutils/filters.py
MuriloDalRi: 👍

tests/test_filters.py
pcraig3: Yeah, sorry, it looks like backtick formatting wasn't preserving all the spaces.  If you have two spaces between `\r\n`s, it's going to treat them as separate groups. 

So, 
<pre>"\r\n\r\n  \r\n\r\n  \r\n\r\n"</pre>
will become
```"<br><br>  <br><br>  <br><br>"```

Comment (MuriloDalRi) [inline tests/test_filters.py]: The new regex will match any number of spaces. I've also added a test for it.

**Comment by pcraig3** [comment]
_I_ think it's good, but let's wait for Alexey to give it a once-over.  In the meanwhile, we can start pulling it into the buyer/briefs/supplier apps and testing it out.

File: dmutils/filters.py [hunk 5]
```diff
@@ -66,3 +66,27 @@ def capitalize_first(maybe_text):
         return [capitalize_first(item) for item in maybe_text]
 
     return maybe_text
+
+# find repeated sequences of '\r\n\', optionally separated by a space character
+_multiple_newlines_re = re.compile(r'(\r\n *){2,}')
```
allait: `r'(\r\n[ \t\f\v]*)'` would also capture tabs and other kinds of white space

dmutils/filters.py
```diff
@@ -66,3 +66,27 @@ def capitalize_first(maybe_text):
...
+_single_newline_re = re.compile(r'(\r\n)')
+
+
+@evalcontextfilter
+def preserve_line_breaks(eval_ctx, value, question_type):
+    if question_type and question_type != 'textbox_large':
+        return value
+
+    # Turn potential markdown objects into string
+    value = '{}'.format(escape(value))
```
allait: I think this needs `u'{}'`, since original value is still unicode. It's also worth writing a test with unicode characters for this

tests/test_filters.py
```diff
@@ -118,3 +121,18 @@ def test_capitalize_first_for_non_strings():
     assert capitalize_first([{'list': 'of'}, 'things']) == [{'list': 'of'}, 'Things']
     assert capitalize_first({'this': 'thing'}) == {'this': 'thing'}
     assert capitalize_first('https://www.example.com') == 'https://www.example.com'
+
+
+@pytest.mark.parametrize("bool", (False, True,))  # Shouldn't matter
+def test_preserve_line_breaks(bool):
```
allait: `bool` is a builtin, which makes the calls a bit confusing without seeing the test declaration. Could we call it something like `autoescape_flag` to disambiguate?

Also, why are we testing this for both cases if we're not checking the Markup return type?

pcraig3: I think the idea is to test functional equivalency rather than having any idea about treating Markup differently. 

If that's already a given, then, yeah, we don't need both cases.

pcraig3:
I think let's go for it 👍

PR #317, opened 2017-07-12
Date/time formats adhere to style guide

katstevens:
Date/time string formats should adhere to the style guide (https://docs.google.com/document/d/1Buo3WGkN2YVUBtlWe7Qvc6bgAcpmKW_GwjHv5Hxs-iQ/edit?ts=595e430f#):
- Use the 12 hr clock (e.g. 11.59pm)
- Remove any leading zeroes from the hour (e.g. 1.59pm instead of 01.59pm)
- Use lower case am/pm

Also added edge case tests for daylight savings/midnight timestamps.

This will support the display of date/time for supplier application deadlines (see ticket: https://trello.com/c/8CJAv25W/714-show-time-opportunity-closes)

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '27.1.2'
+__version__ = '27.2.0'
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Oh you could also squash the commits so the version bump goes with the code in one commit

katstevens:
Done - also fixed a case where the date function returns `None`.

katstevens:
@idavidmcdonald I've had to add a new filter to handle the forced UTC+00 format. Does this still count as a patch version number or does it count as a minor change (I'm happy to keep as the former)?

TheDoubleK:
I'd say new filter = new feature = middle number version increase (27.2.0)

PR #319, opened 2017-07-14
Append timezone to UTC-forced dates

katstevens:
Appends `GMT` to all date strings that have been forced to UTC+00.

We also now force GMT for any time that is not 11.59pm, not just for 12.59am times. This should probably be reviewed if we start storing non-naive datestamps in the DB.

Ticket: https://trello.com/c/8CJAv25W/714-show-time-opportunity-closes

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '27.2.0'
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Approved but would be nice to add a comment to the `utc` filter as discussed.

risicle:
Hmm think this should have been a major version bump - it broke the supplier frontend's tests and I could only find out how/where by doing a bisect.

PR #321, opened 2017-08-15
Fix Python3 format_links filter

katstevens:
Bug: The `format_links` template filter was converting Markup content to strings and 'helpfully' escaping it for us. 

To fix: we explicitly cast the value to its original type, rather than returning the string output by re. (Thanks @allait for this idea).

Once this is merged we can bump the version number on the buyer-frontend and briefs-frontend, where this filter is used.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '27.2.1'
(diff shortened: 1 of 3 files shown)


CrystalPea:
Looks good to me 👍

TheDoubleK:
Looks like a win to me :+1:

idavidmcdonald:
Will need version bump!

katstevens:
Oops - now bumped 😄

PR #325, opened 2017-09-01
forms: add EmailField

risicle:
Vaguely connected story: https://trello.com/c/V8bUSjq8/77-implement-an-email-sign-up-capability-for-suppliers-2

We have many uses of the following code copypasta'd around the dm:

```
email = StripWhitespaceStringField('Contact email', validators=[
    DataRequired(message="You must provide an email address"),
    Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$", message="Please enter a valid email address")
])
```

Ignoring the fact that the regex isn't quite right (e.g. not a raw string literal!) it's not very nice to have this in bits around the place. So I've added an `EmailField` here centrally which we can try and get right - in that vein I've added a bunch of tests. Running these tests you'll note causes us to now depend on `Flask` and `Flask-WTF`.

Arguably we should be depending on at least wtforms anyway (and not just as a dev dependency) because it is *required* for some of the functionality of the package, we just happen to be using it from projects that already pull those deps in anyway. Anyway, probably not too important.

Note also I didn't include the `DataRequired` validation for this field as we want to allow maximum reuse (with optional fields too).

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '28.0.3'
+__version__ = '28.1.0'
--- a/dmutils/forms.py
+++ b/dmutils/forms.py
@@ -1,4 +1,8 @@
+from itertools import chain
+import re
+
 from wtforms import StringField
+from wtforms.validators import Regexp
 
 
 class StripWhitespaceStringField(StringField):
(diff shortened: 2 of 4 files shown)


risicle:
I've also now exposed the underlying `EmailValidator` to allow more flexible use.

TheDoubleK:
This looks good and useful to me :+1:

risicle:
After a bit of thought I've decided to keep the default message as "Please enter a valid email address" because even though "You must provide a valid email address." is used in more places from what I've seen so would require more overriding. Thing is, the former is more generic and the latter rather implies it's a required field, and I'm trying to make that as un-assumed as possible.

PR #328, opened 2017-09-08
DMMailChimpClient: make subscribe_new_email_to_list more robust to exceptions

risicle:
Not all `RequestException`s have a usable `response` attribute (e.g. if they never managed to make their connection) so this much be checked before trying to access `.json()`


TheDoubleK:
Makes sense.

:+1:

PR #332, opened 2017-09-15
Move ODS module from briefs-frontend and add general DownloadFileView

samuelhwilliams:
## Summary
The briefs-frontend has a module to help construct ODS files. We now need to make ODSs in the buyer-frontend for service shortlist downloads, so let's move it to utils.

I've also added a generic `DownloadFileView` that is subclassed in the buyer-frontend (and should be reasonably compatible with the briefs-frontend download view, which I will put up in a separate PR later). This draws out the duplication from the view and lets the subclass only implement the logic and structure required for the csv/ods files, while the parent handles receiving the request and returning the response. Hooks are included for the start and end of the request to allow further logic to be executed at these stages (the shortlist download needs to record the action if the response succeeds).

Bumps version to 28.6.0.

## Blocked by
* [x] Need to write tests for the DownloadFileView.

## Required by
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/615
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/32

## Other associated PRs
https://github.com/alphagov/digitalmarketplace-api/pull/656
~~https://github.com/alphagov/digitalmarketplace-apiclient/pull/98~~
~~https://github.com/alphagov/digitalmarketplace-frameworks/pull/473~~

## Ticket
https://trello.com/c/igRhA3Bt/711-shortlist-file-download

### DIFF
--- a/dmutils/views.py
+++ b/dmutils/views.py
@@ -1,0 +1,225 @@
+from abc import ABCMeta, abstractmethod
+import csv
+import enum
+from flask import abort, request, Response
+from flask.views import View
+from io import BytesIO
+from odf.style import TextProperties, TableRowProperties, TableColumnProperties, TableCellProperties, FontFace
+import six
+
+from dmutils import csv_generator
+from dmutils import ods
+
+
+@six.add_metaclass(ABCMeta)
+class DownloadFileView(View):
+    """An abstract base class appropriate for subclassing in the frontend apps when the user needs to be able to
+    download some data as a CSV or ODS file. All abstract methods must be implemented on the subclass (although example
+    implementations are included here with the kind of return value expected); all other methods should be able to
+    be left alone to support handling and dispatching the request."""
+
+    FILETYPES = enum.Enum('Filetypes', ['CSV', 'ODS'])
+
+    def __init__(self, **kwargs):
+        self.request = request
+
+        self.data_api_client = None
+        self.search_api_client = None
+        self.content_loader = None
+
+        super(View, self).__init__(**kwargs)
+
+    @abstractmethod
+    def _init_hook(self, **kwargs):
+        """Hook into the start of the request to associate required clients/loaders with the instance and perform any
+        other logic required for the start of the request."""
+        self.data_api_client = None
+        self.search_api_client = None
+        self.content_loader = None
+
+    def _post_request_hook(self, response=None, file_context=None, **kwargs):
+        """A hook that can be used to implement any logic that should be run after the main download file process,
+        and immediately before sending the response to the user."""
+        pass
+
+    @abstractmethod
+    def get_file_context(self, **kwargs):
+        """Must return a dictionary containing any data required by the generation routines.
+
+        Required keys:
+         * `filename`: without a file extension
+         * `sheetname`: the name of the first sheet in the ODS file (assumption: single sheeted spreadsheet)"""
+        return {'filename': 'my-download', 'sheetname': 'Sheet 1'}
+
+    @abstractmethod
+    def determine_filetype(self, file_context=None, **kwargs):
+        """Logic to tell the View which filetype to generate; must return a choice from DownloadFileView.FILETYPES"""
+        return DownloadFileView.FILETYPES.ODS
+
+    @abstractmethod
+    def generate_csv_rows(self, file_context):
+        """Must return a nested iterable of rows+cells, eg [[heading1, heading2], [data1, data2], ...]"""
+        return [['Heading 1', 'Heading 2'], ['Row 1, Column 1', 'Row 1, Column 2']]
+
+    @abstractmethod
+    def populate_styled_ods_with_data(self, spreadsheet, file_context):
+        """Takes an empty dmutils.ods.SpreadSheet (with required fonts/styles already added) and populates the required
+        data into it."""
+        sheet = spreadsheet.sheet(file_context['sheetname'])
+        sheet.write_row(name='header', cells=['Heading 1', 'Heading 2'])
+        sheet.write_row(name='row1', cells=['Row 1, Column 1', 'Row 1, Column 2'])
+
+    @staticmethod
+    def create_blank_ods_with_styles():
+        """Create a dmutils.ods.SpreadSheet pre-configured with some default styles, ready for population with data
+        appropriate for the subclass View. Modifications here (except adding styles) are likely breaking changes."""
+        spreadsheet = ods.SpreadSheet()
+
+        # Add the font we will use for the entire spreadsheet.
+        spreadsheet.add_font(FontFace(name="Arial", fontfamily="Arial"))
+
+        # Add some default styles for columns.
+        spreadsheet.add_style("col-default", "table-column", (
+            TableColumnProperties(breakbefore="auto"),
+        ), parentstylename="Default")
+
+        spreadsheet.add_style("col-wide", "table-column", (
+            TableColumnProperties(columnwidth="150pt", breakbefore="auto"),
+        ), parentstylename="Default")
+
+        spreadsheet.add_style("col-extra-wide", "table-column", (
+            TableColumnProperties(columnwidth="300pt", breakbefore="auto"),
+        ), parentstylename="Default")
+
+        # Add some default styles for rows.
+        spreadsheet.add_style("row-default", "table-row", (
+            TableRowProperties(breakbefore="auto", useoptimalrowheight="false"),
+        ), parentstylename="Default")
+
+        spreadsheet.add_style("row-tall", "table-row", (
+            TableRowProperties(breakbefore="auto", rowheight="30pt", useoptimalrowheight="false"),
+        ), parentstylename="Default")
+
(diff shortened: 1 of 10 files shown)


dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+from flask.views import View
+from io import BytesIO
+from odf.style import TextProperties, TableRowProperties, TableColumnProperties, TableCellProperties, FontFace
+import six
+
+from dmutils import csv_generator
+from dmutils import ods
+
+
+@six.add_metaclass(ABCMeta)
```
galund: I don't entirely follow why formally making this an ABC here makes any difference. (But then, I started reading PEP 3119, and lost the will to live, so meh.) The pattern I've seen before is just to name the thing `_Whatever_Base` and just throw some NotImplemented exceptions for the abstract methods.

galund: (oh I see, abstract methods aren't abstract at all, by my understanding of the term up until now - they are more like 'must override'. Which feels very weird - if you want to provide some helper functions to assist implementation of specific functions, then those could be separate anyway - possibly in another class altogether.)

galund: scratch that, those are examples? I find this confusing

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+
+        response = self.create_response(file_context, file_type)
+
+        self._post_request_hook(response, **kwargs)
+        return response
+
+
+@six.add_metaclass(ABCMeta)
+class SimpleDownloadFileView(DownloadFileView):
+    """A slightly more generic version of the DownloadFileView where it is possible to have a single source of data
```
galund: I think you mean "less generic" :)

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+    for all download filetypes. If you want a fairly simple structure to your spreadsheet (in effect, a single sheet,
+    with self-contained rows where cells don't span rows or columns), you can implement only `get_file_data_and_styles`
+    to work with all supported filetypes."""
+    @abstractmethod
+    def get_file_data_and_styles(self, file_context):
```
galund: Rename to `get_file_data_and_column_styles` perhaps?

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+        """Logic to tell the View which filetype to generate; must return a choice from DownloadFileView.FILETYPES"""
+        return DownloadFileView.FILETYPES.ODS
+
+    @abstractmethod
+    def generate_csv_rows(self, file_context):
+        """Must return a nested list of rows+cells, [[heading1, heading2], [data1, data2], ...]"""
+        return [['Heading 1', 'Heading 2'], ['Row 1, Column 1', 'Row 1, Column 2']]
+
+    @abstractmethod
+    def generate_ods(self, spreadsheet, file_context):
```
galund: Feels like the interface here is confusingly different to the csv equivalent. It might be fixable simply by renaming this to `append_table_to_sheet`. (The names here are a bit confusing, am I right in saying that Excel and Google call the tabs "sheets", but ODF calls the tabs "tables" and the documents themselves are "sheets"?)

In which case I'm not sure that the function should return anything. It's also confusing that the (example?) method calls create_default_ods, overriding what was passed in?

tests/test_ods.py
```diff
@@ -0,0 +1,214 @@
...
+
+        cell = CoveredTableCell.return_value
+
+        instance._row.addElement.assert_called_once_with(cell)
+
+
+class TestSheet(object):
+    @given(st.text())
+    def test___init__(self, name):
+        with po(ods, 'Table') as Table:
```
galund: I'm a bit uncertain about unit testing that mocks implementation details. Sometimes you have to (e.g. API calls), but otherwise you're basically testing the implementation of the function rather than the end result, which makes for very fragile tests, that have to be rewritten if the implementation changes.

Not claiming I know how to do this better, but it's surely ok for a unit test to end up running code that's effectively at a lower layer than the code you're testing.

tests/test_views.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @pytest.mark.parametrize('file_type, call_method',
+                             (
+                                 ('CSV', 'generate_csv_rows'),
+                                 ('ODS', 'generate_ods'),
+                             ))
+    def test_create_response(self, file_type, call_method):
+        kwargs = {'filename': 'test'}
+        self._patch_create_response.stop()
+
```
galund: Again I don't think that testing the implementation detail is really the right thing to do here. It leads to a very large amount of test code, anyway. We should be testing what the response looks like (potentially even just comparing to some fixture-blob) rather than worrying about how many times BytesIO had to get called?!

tests/test_ods.py
samuelhwilliams: test_ods was copied over from briefs, so I'm not really going to touch the tests here.

galund: ah fair enough then

dmutils/views.py
samuelhwilliams: Are the comments insufficient? I guess I don't find it confusing but obviously I wouldn't since I've been so close to it for so long. There are example implementations so that it's easy to look at the class and know how to implement the subclass and what the expected returns should be.

dmutils/views.py
samuelhwilliams: I agree it's different, but it was the simplest for each that I thought of. Though I may have missed a trick.

Yeah the terminology is really weird and conflicting. You've got it right though.

The override was a mistake - taken that out.

samuelhwilliams: I agree it's different, but it was the simplest for each that I thought of. Though I may have missed a trick.

Yeah the terminology is really weird and conflicting. You've got it right though.

The override was a mistake - taken that out.

tests/test_views.py
samuelhwilliams: I've stripped out some of the more useless bits of the implementation testing, but maybe we can sit together for a bit to go over what else is wrong, as I'm drawing a bit of a blank about where to draw the line for implementation vs result.

dmutils/views.py
risicle: They *are* examples, but they *are* also invokable through `super`, I just assume the intention here is not to.

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @abstractmethod
+    def determine_filetype(self, file_context=None, **kwargs):
+        """Logic to tell the View which filetype to generate; must return a choice from DownloadFileView.FILETYPES"""
+        return DownloadFileView.FILETYPES.ODS
+
+    @abstractmethod
+    def generate_csv_rows(self, file_context):
+        """Must return a nested list of rows+cells, [[heading1, heading2], [data1, data2], ...]"""
+        return [['Heading 1', 'Heading 2'], ['Row 1, Column 1', 'Row 1, Column 2']]
```
risicle: Do they *have to* be lists or just sequences?

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+        self.search_api_client = None
+        self.content_loader = None
+
+    def _post_request_hook(self, response=None, file_context=None, **kwargs):
+        """A hook that can be used to implement any logic that should be run after the main download file process,
+        and immediately before sending the response to the user."""
+        pass
+
+    @abstractmethod
+    def get_file_context(self, **kwargs):
```
risicle: I can see you were using the term "context" as used in template rendering, but this confused me - I thought you might be passing around a file context manager.

dmutils/views.py
```diff
@@ -0,0 +1,223 @@
...
+        """Hook into the start of the request to associate required clients/loaders with the instance and perform any
+        other logic required for the start of the request."""
+        self.data_api_client = None
+        self.search_api_client = None
+        self.content_loader = None
+
+    def _post_request_hook(self, response=None, file_context=None, **kwargs):
+        """A hook that can be used to implement any logic that should be run after the main download file process,
+        and immediately before sending the response to the user."""
+        pass
```
risicle: Are you speculatively implementing things that aren't used assuming they *will* be useful?

risicle:
I think the abstract class implementation is objectively *fine*, but it seems an awful lot of hard work to do something relatively straightforward.

dmutils/views.py
risicle: (hmm actually the documentation seems to be contradictory on whether they can or not, whatever...)

dmutils/views.py
samuelhwilliams: No, this _is_ useful for the buyer-frontend implementation.

dmutils/views.py
samuelhwilliams: Eh, I didn't pick the term - this was the method name used in the old briefs-frontend implementation of the download view.

dmutils/views.py
samuelhwilliams: You're right - just sequences.

samuelhwilliams:
Ready for final sign-off @galund

PR #335, opened 2017-09-21
Create user account email class

Wynndow:
The code for emailing a creation or invite email to a buyer or supplier
is really similar. We're also moving these emails to Notify which also
simplifies the process (no need for templates).

This class can be imported by the frontend apps and used to generate and
send the tokens and emails for user creation/invitiation.

### DIFF
--- a/dmutils/email/user_account_email.py
+++ b/dmutils/email/user_account_email.py
@@ -1,0 +1,40 @@
+from flask import current_app, session, abort
+from . import DMNotifyClient, generate_token, EmailError
+from .helpers import hash_string
+
+
+class UserAccountEmail():
+
+    def __init__(self, token_data, template_id):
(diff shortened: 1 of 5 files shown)


dmutils/email/user_account_email.py
```diff
@@ -0,0 +1,40 @@
+from flask import current_app, session, abort
+from . import DMNotifyClient, generate_token, EmailError
+from .helpers import hash_string
+
+
+class UserAccountEmail():
```
allait: This feels like this could be a function. Class attributes are essentially used as arguments for the single method.
Do you think this will get extended in the future in a way that would be easier to do with a class?

Wynndow: Have updated to not be a class, and have imported the external blueprint from the supplier frontend.

PR #337, opened 2017-09-25
Add retry to mandrill requests

benvand:


dmutils/email/dm_mailchimp.py
```diff
@@ -58,6 +64,7 @@ def send_campaign(self, campaign_id):
             )
         return False
 
+    @timeout_retry()
     def subscribe_new_email_to_list(self, list_id, email_address):
```
allait: This method is called  from a supplier-frontend view. I don't think we should have in-request retries as they can potentially tie up application server process for a long time if eg mailchimp is temporary unavailable. Can we have a flag to only enable retries from scripts?

idavidmcdonald: Good spot

Could be something set at the `DMMailChimpClient` level rather than each method in our client needing to know about whether we are going to call it from a script or from a FE app.

PR #341, opened 2017-10-31
Remove old user creation tokens

MuriloDalRi:
https://trello.com/c/Op3wb69r/31-remove-code-in-utils-to-support-old-user-creation-tokens

### DIFF
--- a/dmutils/email/tokens.py
+++ b/dmutils/email/tokens.py
@@ -133,10 +133,6 @@ def decode_invitation_token(encoded_token):
             current_app.config['INVITE_EMAIL_SALT'],
             SEVEN_DAYS_IN_SECONDS
         )
-        if 'role' not in token:
-            token.update({
-                'role': 'supplier' if token.get('supplier_id') else 'buyer'
-            })
         return token
 
     except fernet.InvalidToken as error:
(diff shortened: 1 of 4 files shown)


dmutils/email/tokens.py
```diff
@@ -153,7 +149,7 @@ def decode_invitation_token(encoded_token):
 
             return {
                 'error': 'token_expired',
-                'role': token.get('role') or ('supplier' if token.get('supplier_id') else 'buyer')
```
Wynndow: Thinking about it, I think this line may need to stay as is. It's possible that someone still has an email with a link/token in it that was generated before we switched to the new token styles. 

If they click that link, the old token will throw an InvalidToken error when decrypted. We'll then try and decrypt it again with no max age, which will work, but we'll set the role key here to `None`. Which will then get passed to the user-frontend and the template will get rendered without user role specific information, which would be bad.

idavidmcdonald:
Good spot on breaking change changelog stuff @Wynndow

dmutils/email/tokens.py
MuriloDalRi: Good point, I'll change it back.

MuriloDalRi: Do you reckon we need a new test for that scenario?

Wynndow: I think we already have them, they're the second two tests you're removing in this PR, so should probably keep them.

PR #346, opened 2018-01-10
add support page route

DilwoarH:
Ticket: https://trello.com/c/Ss5PxdxQ/77-contact-filtering-for-buyer-and-supplier-contact-details-lists

Required PRs: 
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/691

screenshot-localhost-2018-01-04-10-30-52-159

### DIFF
--- a/dmutils/external.py
+++ b/dmutils/external.py
@@ -9,6 +9,11 @@ external = Blueprint('external', __name__)
 def get_brief_by_id(framework_framework, brief_id):
     raise NotImplementedError()
 
+
+@external.route('/help')
+def help():
+    raise NotImplementedError()
+
 # Supplier frontend
 
 


samuelhwilliams:
Can you include the ticket in the PR summary please? I'd be interested to read up on the background here. Do we know what app this is going in? It feels probably most suitable to users, in which case the URL would _ideally_ be `/user/support`.

If we keep it as just `/support` we'll need to make other changes alongside this, i.e. update the nginx config to route it to the correct app.

DilwoarH:
@samuelhwilliams - Please see the comment below for the decision

https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/691#issuecomment-357666494

samuelhwilliams:
Yep that's fine then. 👍

PR #348, opened 2018-01-19
If no "timestamp" metadata fall back to AWS last_modified

TheDoubleK:
Addresses this bug: https://trello.com/c/qBiGn3bP/9-legal-documents-files-not-appearing-on-framework-application-dashboard

Some pages in our apps (eg framework dashboard) look for the existence of "last_modified" on an S3 key as an indicator for if the file exists or not, and only show a link to files with a "last_modified" date.

This meant that if a file was uploaded to S3 using the AWS console and didn't have "timestamp" set then the link to the file would not display, even though the file is there.

This also fixes all warnings generated during test runs.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '31.1.0'
+__version__ = '31.1.1'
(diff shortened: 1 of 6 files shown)


dmutils/__init__.py
```diff
@@ -4,4 +4,4 @@
 import flask_featureflags  # noqa
 
 
-__version__ = '31.1.0'
+__version__ = '31.1.1'
```
samuelhwilliams: Would this maybe be considered a breaking/major change? Previously this would ignore/bypass files that didn't have a custom timestamp; now it will return any file present in the bucket.

PR #349, opened 2018-01-22
Adds missing external routes used by other applications

DilwoarH:
adds in:

- `/`
- `/g-cloud/suppliers`
- `/suppliers`
- `/buyers/create`

Bumps version to `31.2.0`

### DIFF
--- a/dmutils/external.py
+++ b/dmutils/external.py
@@ -5,8 +5,18 @@ external = Blueprint('external', __name__)
 # Buyer frontend
 
 
+@external.route('/')
+def index():
+    raise NotImplementedError()
+
+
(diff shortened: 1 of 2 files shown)


dmutils/external.py
```diff
@@ -5,17 +5,32 @@
...
+
 @external.route('/help')
 def help():
     raise NotImplementedError()
 
+
 # Supplier frontend
 
+@external.route('/suppliers')
+def dashboard():
```
samuelhwilliams: We should probably call this `supplier_dashboard`.

DilwoarH: done @samuelhwilliams

PR #350, opened 2018-01-22
Add localiZed time zone format

brenetic:
Adds date time formatting for BST

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '31.2.0'
+__version__ = '31.2.1'
--- a/dmutils/formats.py
+++ b/dmutils/formats.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
 import pytz
+import re
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 DATE_FORMAT = "%Y-%m-%d"
 DISPLAY_SHORT_DATE_FORMAT = '%-d %B'
 DISPLAY_DATE_FORMAT = '%A %-d %B %Y'
 DISPLAY_TIME_FORMAT = '%H:%M:%S'
-DISPLAY_DATETIME_FORMAT = '%A %-d %B %Y at %I:%M%p'
+DISPLAY_DATETIME_FORMAT = '%A %-d %B %Y at %I:%M%p %Z'
 
 
 def timeformat(value, default_value=None):
@@ -40,8 +41,8 @@ def utcdatetimeformat(value, default_value=None):
     if local_format:
         if "11:59pm" not in local_format:
             # Force UTC+00 if the date has rolled over to the next day and append timezone
-            return "{} GMT".format(datetimeformat(value, default_value, localize=False))
-        return "{} GMT".format(local_format)
+            return _use_gmt_timezone(datetimeformat(value, default_value, localize=False))
+        return _use_gmt_timezone(local_format)
     return None
 
 
@@ -54,6 +55,10 @@ def datetodatetimeformat(value):
 
 
 EUROPE_LONDON = pytz.timezone("Europe/London")
+
+
+def _use_gmt_timezone(date_string):
+    return re.sub(r"UTC|BST", "GMT", date_string)
 
 
 def _format_date(value, default_value, fmt, localize=True):
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -49,16 +49,16 @@ def test_dateformat(dt, formatted_date):
 
 
 @pytest.mark.parametrize("dt, formatted_datetime", (
-    (datetime(2012, 11, 10, 9, 8, 7, 6), "Saturday 10 November 2012 at 9:08am"),
-    ("2012-11-10T09:08:07.0Z", "Saturday 10 November 2012 at 9:08am"),
-    (datetime(2012, 8, 10, 9, 8, 7, 6), "Friday 10 August 2012 at 10:08am"),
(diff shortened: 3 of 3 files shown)


dmutils/formats.py
```diff
@@ -22,14 +23,18 @@ def dateformat(value, default_value=None):
...
+def datetimeformat(value, default_value=None, localize=True, format_type=DISPLAY_DATETIME_FORMAT):
     # en_GB locale uses uppercase AM/PM which contravenes our style guide
-    formatted_date = _format_date(value, default_value, DISPLAY_DATETIME_FORMAT, localize=localize)
+    formatted_date = _format_date(value, default_value, format_type, localize=localize)
     if formatted_date:
         return formatted_date.replace('AM', 'am').replace('PM', 'pm').replace(" 0", " ")
     return formatted_date
 
 
+def localizeddatetimeformat(value, default_value=None, localize=True):
```
galund: This is all fine, we end up with questions of naming things, which isn't very interesting so feel free to ignore, but here goes anyway...

ISTM that the existing function is already supposed to be a "localized" function, at least by default, seeing as it has that parameter `localize=True` by default.

I'm not sure defaulting to the existing behaviour is right, and we've ended up in a position where _by default_ dates come out in UTC, but everything else comes out in London time - but depending on which function you use, it might not have a time zone.

None of this is going to breed any kind of consistency in our applications. Potentially we want two things:

1) A way to get UTC date-only thing - this function should be named in a particular way so that nobody uses it by mistake. We can bump versions and go backward-incompatible if necessary, I think.

2) A way to get a normal date-only string that converts to London TZ time by default.

3) A way to get a full date-time, that also converts to London time by default, and always includes the time zone indicator.

katstevens: I guess this is a symptom of the wider problem of 'date formats in the Digital Marketplace are  already wildly inconsistent'. For example, we import the raw `DATE_FORMAT` and `DISPLAY_DATE_FORMAT` (etc) strings in a bunch of places (content loader, scripts, api validation) without passing them through these `localize` functions in the utils. And in a bunch of other places we don't use the utils patterns at all! Some of these dates are displayed to users (eg in script generated emails), some are for filenames, it's a bit of a mess. I sense a tech debt ticket coming on 😸 

However, in general I agree that **localised should be the default for displaying dates to users** unless it's the 23:59:59 deadline case (where daylight savings could tip it over into the next day, causing confusion for people applying). 

Just to sanity check - the third case you list is what we've just implemented in this PR (`Monday 26 March 2012 at 12:59am BST`)?

dmutils/flask_init.py
```diff
@@ -56,6 +56,7 @@ def add_header(response):
     application.add_template_filter(formats.dateformat)
     application.add_template_filter(formats.datetimeformat)
     application.add_template_filter(formats.datetodatetimeformat)
+    application.add_template_filter(formats.localizeddatetimeformat)
```
samuelhwilliams: Let's add a tech debt ticket to review all of our date time filters. Having six is mildly terrifying 😨 (although having this one feels like a _good_ thing regardless).

samuelhwilliams: https://trello.com/c/jyQZCEB7/303-review-datetime-filters

tests/test_formats.py
```diff
@@ -66,6 +66,15 @@ def test_datetimeformat(dt, formatted_datetime):
     assert datetimeformat(dt) == formatted_datetime
 
 
+@pytest.mark.parametrize("dt, formatted_datetime", (
+    (datetime(2012, 3, 24, 23, 59, 7, 6, tzinfo=pytz.utc), "Saturday 24 March 2012 at 11:59pm GMT"),
+    (datetime(2012, 3, 25, 23, 59, 7, 6, tzinfo=pytz.utc), "Monday 26 March 2012 at 12:59am BST"),
```
samuelhwilliams: I had to double take at this one, although now I have it's clear that this is fine. But could we maybe swing these dates to _obvious_ extremes? My first thought when I saw two dates so close together is that there was some other interface to the function that would let you toggle the GMT/BST thing. I think if the dates were in January/July it'd be more obvious that the date is controlling the timezone rather than anything else.

dmutils/formats.py
galund: Sanity check - yes (my complaint there is just the name is not the obvious one you would choose, if you didn't look carefully at the code).

katstevens: Fair enough. I think we went for `localiseddateformat` as the counterpart to `utcdateformat`, but we can rename to something like... `datetimezoneformat` (ugh, must be something better!)

tests/test_formats.py
katstevens: It's using the same examples as the test below, so we can compare the behaviour of the different filters. Perhaps an explanatory comment would be useful after all.

dmutils/__init__.py
```diff
@@ -4,4 +4,4 @@
 import flask_featureflags  # noqa
 
 
-__version__ = '31.2.0'
+__version__ = '32.0.0'
```
samuelhwilliams: If we're just adding a feature, this isn't a major (breaking) change, so could this be `31.3.0`?

katstevens: We're changing an existing filter now to add ` BST` or ` GMT` on the end. It's probably borderline whether this is a breaking change - anyone else have a strong opinion?

TheDoubleK: It's a change of design, but won't break any existing code as far as I can see - you will be able to pull this change in to apps without needing to change any application code (tests that check date strings don't count), so definitely not a breaking change to me.

TheDoubleK: In fact, I'd probably argue that this is only a patch change - it's not a new feature, it's an adjustment of an existing feature, so I'd probably go `31.2.1`.

katstevens: Count me convinced 😸 @brenetic do you want to do the honours?

PR #351, opened 2018-01-24
Adds missing external routes used by buyer application

DilwoarH:


dmutils/external.py
```diff
@@ -32,6 +32,11 @@ def supplier_dashboard():
     raise NotImplementedError()
 
 
+@external.route('/suppliers/opportunities/<brief_id>/responses/start')
```
andreagrandi: A newbie question from a Django developer (so, please ignore me if I'm saying something stupid): shouldn't be `<int:brief_id>` instead of `<brief_id>` ?

DilwoarH: I'm a newbie to python too, @samuelhwilliams ?

samuelhwilliams: Either will work, although if you know the type then it can be good to be explicit by including it, as in @andreagrandi's example.

andreagrandi: Oh cool, I guessed it correctly then :) I spotted it because I was looking at Django 2.0 new urls just yesterday and they basically adopted the same syntax of Flask (which I'm less familiar with)

PR #352, opened 2018-01-29
Document the usage of our various date and datetime formats

TheDoubleK:
For this tech-debt ticket: https://trello.com/c/jyQZCEB7/303-review-datetime-filters

I've grepped through all our repos and I think everything here is true.  Honest.

### DIFF
--- a/dmutils/formats.py
+++ b/dmutils/formats.py
@@ -19,7 +19,7 @@ def timeformat(value, default_value=None):
     "timeformat" is used in summary tables where space is tight and times are shown on their own line.
     Currently (Jan 2018) only used in admin app tables.
     """
-    return _format_date(value, default_value, DISPLAY_TIME_FORMAT)
+    return _format_date(value, default_value, DISPLAY_TIME_FORMAT, localize=False)
 
 
 def shortdateformat(value, default_value=None):
--- a/dmutils/formats.py
+++ b/dmutils/formats.py
@@ -13,6 +13,9 @@ DISPLAY_DATETIME_FORMAT = '%A %-d %B %Y at %I:%M%p %Z'
 
 def timeformat(value, default_value=None):
     """
+    Example value: datetime.strptime("2018-07-25 10:15:00", "%Y-%m-%d %H:%M:%S")
+    Example output: '11:15:00'
+
     "timeformat" is used in summary tables where space is tight and times are shown on their own line.
     Currently (Jan 2018) only used in admin app tables.
     """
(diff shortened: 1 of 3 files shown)


dmutils/formats.py
```diff
@@ -12,30 +12,59 @@
...
 
 
 def utcdatetimeformat(value, default_value=None):
     """
-    Use this filter to format deadline timestamps that are stored in the database as
+    Only use this filter to format deadline timestamps that are stored in the database as
     23:59:59 (UTC+00), when the localisation would otherwise display the date as 12.59am
-     on the next day (due to daylight savings), potentially causing confusion for buyers/suppliers.
+    on the next day (due to daylight savings), potentially causing confusion for buyers/suppliers.
+    This is currently (Jan 2018) used wherever we show the closing date for a brief.
```
katstevens: It's also used for the Clarification Questions deadline I think? (At least it should be!)

TheDoubleK: Yep, I can add that.

TheDoubleK:
I'm not sure it is... the formats themselves are defined just above the formatter functions in the same file, but in theory these formats could be changed at any time and the formatters shouldn't care about it.

I guess the formats could have an example next to them, and then the examples would be more likely to be updated if the formats change.  I can add tomorrow.

CrystalPea:
Good idea with adding the examples near the formats themselves, makes sense.

dmutils/formats.py
```diff
@@ -47,6 +94,17 @@ def utcdatetimeformat(value, default_value=None):
 
 
 def datetodatetimeformat(value):
```
samuelhwilliams: Interesting. This method feels like it has the wrong name. It doesn't take a date (`date to`) and turn it into the same format as `datetimeformat`. It takes a date and turns it into the same format as comes out of our `dateformat`. Breaking change, but it feels like this should be called `datetodateformat`...

katstevens: We should also probably rename `utcdatetimeformat` to make it clear that we're *forcing* to UTC, but probably not in this PR?

TheDoubleK:
This PR and associated tech debt ticket is all about **adding some comments** and definitely not about **breaking all the things**.

TheDoubleK:
@CrystalPea I decided against having examples next to the formats themselves - I tried it out but I think it seems better with the formatter descriptions, and let's just hope people keep comments up to date if they change things in future.

dmutils/formats.py
samuelhwilliams: Yeah, seems reasonable. But not for this PR as you say. 👍

PR #354, opened 2018-02-07
Use `SHARED_EMAIL_KEY` not `SECRET_KEY` for reset password tokens

idavidmcdonald:
https://trello.com/c/WTbemGdn/219-reset-password-emails-are-using-secretkey-instead-of-sharedemailkey-to-encrypt-tokens

We will be changing our tokens for reset password emails to be
generated using the `SHARED_EMAIL_KEY` rather than the `SECRET_KEY`
in an applications config. This is because
- this will be consistent with the other email tokens we generate
- although there is technically no harm in using `SECRET_KEY`,
the actual purpose of the `SECRET_KEY` is to be used for encrypting
Flask sessions. This would mean if we needed to rotate a key for
our Flask sessions it would also break out email tokens and vice
versa. By using different keys we should reduce the risk of a
token rotation having an impact on too large a part of the site

We maintain backwards compatability so tokens generated within the
past day using the old token will not be broken by this change.
The next step will be to start encrypting tokens with
`SHARED_EMAIL_KEY` for a full day at which point we can remove
the backwards compatability support for `SECRET_KEY`.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '31.3.2'
+__version__ = '31.4.0'
--- a/dmutils/email/tokens.py
+++ b/dmutils/email/tokens.py
@@ -100,13 +100,24 @@ def decode_password_reset_token(token, data_api_client):
     try:
         decoded, token_timestamp = decode_token(
             token,
-            current_app.config["SECRET_KEY"],
+            current_app.config["SHARED_EMAIL_KEY"],
             current_app.config["RESET_PASSWORD_SALT"],
             ONE_DAY_IN_SECONDS,
         )
-    except fernet.InvalidToken as e:
-        current_app.logger.info("Error changing password: {error}", extra={'error': six.text_type(e)})
-        return {'error': 'token_invalid'}
+    except fernet.InvalidToken:
+        try:
+            # We used to issue tokens with `current_app.config["SECRET_KEY"]` as our key, so we maintain backwards
+            # compatability with them for a short period (1 day) to allow for both type of tokens to be valid until the
+            # older ones have expired and can remove the support for tokens using `current_app.config["SECRET_KEY"]`
+            decoded, token_timestamp = decode_token(
+                token,
+                current_app.config["SECRET_KEY"],
+                current_app.config["RESET_PASSWORD_SALT"],
+                ONE_DAY_IN_SECONDS,
+            )
+        except fernet.InvalidToken as e:
+            current_app.logger.info("Error changing password: {error}", extra={'error': six.text_type(e)})
+            return {'error': 'token_invalid'}
 
     user = data_api_client.get_user(decoded["user"])
     user_last_changed_password_at = datetime.strptime(
--- a/tests/email/test_tokens.py
+++ b/tests/email/test_tokens.py
@@ -112,41 +112,54 @@ def test_decrypt_token_ok_for_known_good_token():
     assert data == ({'email': 'test@example.com', 'user': 123}, datetime(2016, 1, 1, 12, 0, 0))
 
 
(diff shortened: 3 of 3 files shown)


dmutils/__init__.py
```diff
@@ -4,4 +4,4 @@
 import flask_featureflags  # noqa
 
 
-__version__ = '31.3.2'
+__version__ = '31.4.0'
```
Wynndow: Is this actually a breaking change?

dmutils/email/tokens.py
```diff
@@ -100,13 +100,24 @@ def decode_password_reset_token(token, data_api_client):
...
+            # compatability with them for a short period (1 day) to allow for both type of tokens to be valid until the
+            # older ones have expired and can remove the support for tokens using `current_app.config["SECRET_KEY"]`
+            decoded, token_timestamp = decode_token(
+                token,
+                current_app.config["SECRET_KEY"],
+                current_app.config["RESET_PASSWORD_SALT"],
+                ONE_DAY_IN_SECONDS,
+            )
+        except fernet.InvalidToken as e2:
+            current_app.logger.info("Error changing password: {error}", extra={'error': six.text_type(e2)})
```
Wynndow: Is it worth getting rid of this use of `six`?

dmutils/email/tokens.py
```diff
@@ -100,13 +100,24 @@ def decode_password_reset_token(token, data_api_client):
...
+            # We used to issue tokens with `current_app.config["SECRET_KEY"]` as our key, so we maintain backwards
+            # compatability with them for a short period (1 day) to allow for both type of tokens to be valid until the
+            # older ones have expired and can remove the support for tokens using `current_app.config["SECRET_KEY"]`
+            decoded, token_timestamp = decode_token(
+                token,
+                current_app.config["SECRET_KEY"],
+                current_app.config["RESET_PASSWORD_SALT"],
+                ONE_DAY_IN_SECONDS,
+            )
+        except fernet.InvalidToken as e2:
```
Wynndow: Could drop the `as e` from line 107, then this only needs to be `as e`. Although could leave till this gets cleand up after the transition.

dmutils/email/tokens.py
idavidmcdonald: I initially did. However it turns out we are supporting both python 2 and python 3 for utils (I guess it's because they are still being used by the scripts repo which still runs a few things in 2).

dmutils/email/tokens.py
idavidmcdonald: Just dropped it as it broke flake8 (strangely only in python 2 and not in python 3)

dmutils/__init__.py
idavidmcdonald: I wasn't sure about this.

Reason it is a breaking change - someone who uses this function could see different behaviour from before
Reason it is not a breaking change - someone who uses this function will not need to make any code changes if they upgrade to this version as it somewhat adds functionality (support from another key)

https://semver.org/ refers to a breaking change as breaking the `public API` but that didn't help me decide either. My plan was to make it a breaking change when I withdraw support for the old token in the next PR.

I have this same debate with myself when things change in the FE toolkit. Let's say you changed a button from being small and blue to being big and green but no code changes are required for the person pulling in the repo, is that a breaking or non breaking change? The visual interface has debatably 'broken' but the interface might not have.

Would be good to hear your opinion?

Wynndow: I would say non breaking change. All things being equal if I bumped the version of utils to this latests one, nothing will break. There's increased functionality (I can encode my tokens with two different keys now), but I don't need to change anything.

The example with the button is a bit different I think, as you now can't do thing you used to (click a small blue button).

I think major bump when you drop support for the old token

dmutils/email/tokens.py
Wynndow: Fair dos.

dmutils/email/tokens.py
Wynndow: nice

PR #356, opened 2018-02-13
Change back to eea/odfpy

samuelhwilliams:
## Summary
Stop using our alphagov fork of odfpy, which was created to allow us to patch a bug until it went into the main library. Our patch was merged into release 1.3.6 of odfpy, so now we can just use that.

Reverts https://github.com/alphagov/digitalmarketplace-utils/pull/343

## Ticket
https://trello.com/c/3bDwjW6V/278-odfpy-remove-dependency-on-alphagov-fork

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Records breaking changes from major version bumps
 
+## 33.0.0
+
+Reverting our usage of an interim alphagov fork of odfpy now that our patch has been merged into master and released.
+1
+ACTION: update your `requirements-app.txt`, removing the alphagov odfpy
+github URI (if present).
+
 ## 32.0.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 Records breaking changes from major version bumps
 
+## 33.0.0
+
+Reverting our usage of an interim alphagov fork of odfpy now that our patch has been merged into master and released.
+1
+ACTION: update your `requirements-app.txt`, removing the alphagov odfpy
+github URI.
```
risicle: Might like to add "...if it exists", because apps that don't care about ods probably never added it and this may leave some people head-scratching.

samuelhwilliams:
Updated.

PR #357, opened 2018-02-13
adds in logout external link

DilwoarH:


dmutils/external.py
```diff
@@ -59,6 +59,11 @@ def render_login():
     raise NotImplementedError()
 
 
+@external.route('/user/logout')
```
samuelhwilliams: I don't know if it makes much of a difference, but this is a POST-only route (`route('/logout', methods=["POST"])`). Can we add this so that it's the same?

DilwoarH: done

DilwoarH:
@samuelhwilliams updated

PR #360, opened 2018-02-23
 Remove backwards compatibility for email importing

benvand:
We no longer need this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Records breaking changes from major version bumps
 
+## 34.0.0
+
+PR: [#360](https://github.com/alphagov/digitalmarketplace-utils/pull/360)
+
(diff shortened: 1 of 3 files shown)


risicle:
So what should https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/blob/729b4b1a6bd0beb0ced75a3e2a662e3d66880f17/app/main/helpers/briefs.py#L41 do then?

benvand:
Weird, how did that sneak past my greps.
It should be
`from dmutils.email.dm_mandrill import send_email`

benvand:
https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/35/files

risicle:
Thanks. I think you can imagine what my comment will be about this shouldn't be a problem we're having and architecture etc. so I'll spare you it.

benvand:
:)

PR #361, opened 2018-02-28
request id handling: back-populate the first value of DM_TRACE_ID_HEADERS back to the DM_REQUEST_ID_HEADER setting

risicle:
...because it turns out that some components (notably the apiclient) depend on that setting - we can get rid of this once that/those are fixed. this also has the nice property that `DM_REQUEST_ID_HEADER` should now always equal the first entry of `DM_TRACE_ID_HEADERS`, even if it was `DM_TRACE_ID_HEADERS` that was overridden.

Trello https://trello.com/c/dlbLmRDB/343-make-use-of-zipkin-headers-in-preference-to-request-id-pt-1

### DIFF
--- a/dmutils/request_id.py
+++ b/dmutils/request_id.py
@@ -101,6 +101,12 @@ def init_app(app):
     app.config.setdefault("DM_SPAN_ID_HEADERS", ("X-B3-SpanId",))
     app.config.setdefault("DM_PARENT_SPAN_ID_HEADERS", ("X-B3-ParentSpan",))
 
+    # we do something a little odd here now - back-populate the first value of DM_TRACE_ID_HEADERS back to the
+    # DM_REQUEST_ID_HEADER setting, because it turns out that some components (notably the apiclient) depend on that
+    # setting
+    if app.config.get("DM_TRACE_ID_HEADERS"):
+        app.config["DM_REQUEST_ID_HEADER"] = app.config["DM_TRACE_ID_HEADERS"][0]
+
     # dynamically define this class as we don't necessarily know how request_class may have already been modified by
     # another init_app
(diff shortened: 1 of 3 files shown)


galund:
When you say "some apps"... *all* apps use the API client, don't they? Isn't fixing the API client just part of this whole thing?

risicle:
I said "some apps" I more accurately meant "some repos", meaning the apiclient as a single entity. Also for some reason I've left the full comment in the source where I meant to yank it out into a comment and leave a truncated version.

Yes, fixing the apiclient is part of fixing the whole thing, but I'd like to be able to deploy these changes independently instead of having a "flag day".

samuelhwilliams:
The summary doesn't make sense to me. Seems like the first sentence isn't finished and you started a side comment in parenthesis that never end? 🤔

dmutils/request_id.py
```diff
@@ -101,6 +101,11 @@ def init_app(app):
     app.config.setdefault("DM_SPAN_ID_HEADERS", ("X-B3-SpanId",))
     app.config.setdefault("DM_PARENT_SPAN_ID_HEADERS", ("X-B3-ParentSpan",))
 
+    # we do something a little odd here now - back-populate the first value of DM_TRACE_ID_HEADERS back to the
+    # DM_REQUEST_ID_HEADER setting, because it turns out that some apps (notably the apiclient depend on that setting
```
samuelhwilliams: Same half sentence here that I can't work out what it means.

risicle: I didn't close my bracket.

risicle:
Ok I've tested this locally integrated with the admin add and it appears to work correctly.

PR #364, opened 2018-03-05
Add supplier organisation size to User object

katstevens:
Trello: https://trello.com/c/PZyZdMtG/337-custom-dimension-to-capture-size-of-supplier

For our analytics we want to capture logged-in suppliers' organisation size. This needs to be exposed via an attribute of the Flask user object, so we can set the custom dimension in the template, for example:

```
{% set custom_dimensions = [
      {"data_id": 11, "data_value": current_user.supplier_organisation_size }
   ] 
%}
```

This version of utils should be backwards compatible even if the relevant API changes (see https://github.com/alphagov/digitalmarketplace-api/pull/754) haven't been released yet.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '34.1.1'
+__version__ = '34.2.0'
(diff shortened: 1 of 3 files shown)


brenetic:
👍 Addition to serialization looks good

PR #367, opened 2018-03-12
Bump version to 34.5.0

DilwoarH:
Adds in user research flag

### DIFF
--- a/dmutils/user.py
+++ b/dmutils/user.py
@@ -7,7 +7,7 @@ def user_has_role(user, role):
 
 class User():
     def __init__(self, user_id, email_address, supplier_id, supplier_name, supplier_organisation_size,
-                 locked, active, name, role):
+                 locked, active, name, role, user_research_opted_in):
         self.id = user_id
         self.email_address = email_address
         self.name = name
@@ -17,6 +17,7 @@ class User():
         self.supplier_organisation_size = supplier_organisation_size
         self.locked = locked
         self.active = active
+        self.user_research_opted_in = user_research_opted_in
 
     def is_authenticated(self):
         return self.is_active()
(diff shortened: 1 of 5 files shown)


dmutils/user.py
```diff
@@ -72,7 +75,8 @@ def from_json(user_json):
             locked=user.get('locked', False),
             active=user.get('active', True),
             name=user['name'],
-            role=user['role']
+            role=user['role'],
+            user_research_opted_in=user['userResearchOptedIn']
```
katstevens: Is there a possibility that an API User response won't contain this key (tests, maybe)? It might be a good idea to initialise it to `None` as per the `supplier_foo` fields so that this is backwards-compatible.

DilwoarH: Hey Kat,

Thanks for the review, I have updated.

dmutils/user.py
```diff
@@ -59,10 +61,13 @@ def from_json(user_json):
         supplier_id = None
         supplier_name = None
         supplier_organisation_size = None
+        userResearchOptedIn = user['userResearchOptedIn'] if user['userResearchOptedIn'] else None
```
katstevens: I think this could be simplified further to `user_research_opted_in = user.get('userResearchOptedIn')`? (Note the snake case for the Python var...)

DilwoarH: thanks (didn't know it was a thing in python) :) - updated

katstevens: No worries 😸 In fact we could probably get rid of the var altogether and do the `user.get`  when creating the User object on line 81.

DilwoarH: @katstevens updated

samuelhwilliams:
FYI adding a required parameter to the `User` class is a breaking change. https://github.com/alphagov/digitalmarketplace-test-utils/blob/master/dmtestutils/login.py constructs `User` objects from some JSON blobs and this construction now fails without the `role`/`user_research_opted_in` parameters.

Will add a PR to bump and document this.

PR #368, opened 2018-03-13
Add disk space check for healthcheck

samuelhwilliams:
## Summary
We want to be able to check free disk space on instances as part of
their healthcheck endpoint. Here's where we are going to do it.

`get_disk_space_status` will report 'OK' if at least 5% disk space is
free (by default) and 'LOW' if less than 5% is free.

## Ticket
https://trello.com/c/BBvu6eUk/383-ensure-healthcheck-endpoint-fails-if-low-disk-space

### DIFF
--- a/dmutils/status.py
+++ b/dmutils/status.py
@@ -1,5 +1,6 @@
-import os
 import datetime
+import math
+import os
 from .formats import DATE_FORMAT
 from flask_featureflags import FEATURE_FLAGS_CONFIG
 
@@ -24,6 +25,19 @@ def get_flags(current_app):
                 flags[config_var] = current_app.config[config_var]
 
     return flags
+
+
+def get_disk_space_status(low_disk_percent_threshold=5):
+    """Accepts a single parameter that indicates the minimum percentage of disk space which should be free for the
+    instance to be considered healthy.
+
+    Returns a tuple containing two items: a status (OK or LOW) indicating whether the disk space remaining on the
+    instance is below the threshold and the integer percentage remaining disk space."""
+    disk_stats = os.statvfs('/')
+
+    disk_free_percent = 100 - int(math.ceil(((disk_stats.f_bfree * 1.0) / disk_stats.f_blocks) * 100))
+
+    return 'OK' if disk_free_percent >= low_disk_percent_threshold else 'LOW', disk_free_percent
 
 
 def enabled_since(date_string):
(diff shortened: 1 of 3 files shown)


dmutils/status.py
```diff
@@ -26,6 +28,19 @@ def get_flags(current_app):
     return flags
 
 
+def get_disk_space_status(low_disk_threshold=5):
```
galund: Variable name needs an indication of units / scale.

dmutils/status.py
```diff
@@ -26,6 +28,19 @@ def get_flags(current_app):
...
+    """Accepts a single parameter that indicates the minimum percentage of disk space which should be free for the
+    instance to be considered healthy.
+
+    Returns a tuple containing two items: a status (OK or LOW) indicating whether the disk space remaining on the
+    instance is below the threshold and the integer percentage remaining disk space."""
+    disk_usage = shutil.disk_usage('/')
+
+    disk_free_percent = math.floor((disk_usage.free / disk_usage.total) * 100)
+
+    return 'OK' if disk_free_percent >= low_disk_threshold else 'LOW', disk_free_percent
```
galund: Not quite sure I understand how "LOW" will be used - either a healthcheck passes or it fails? Can reserve judgment until I've looked at a PR that brings this into an app :-)

samuelhwilliams: I imagine the healthcheck endpoints basically just seeing if this returns 'OK', and if not, failing the healthcheck.

galund: Oh I see, LOW means fail, duh :)

dmutils/status.py
samuelhwilliams: Updated.

samuelhwilliams:
I originally did this as working with raw values but it's a fair bit uglier, and frankly we don't write anything except logs to disk so they should never fill up anywhere near full, so the difference between a static MB free and % free shouldn't matter so much.

Have just noticed that this fails on Py2 because shutil's disk_usage isn't available there. Can fall back to something else but it would be Unix-only. This is probably fine but, do we still need to support Python2 here?

galund:
FWIW I'd be fine with removing the requirement that this lib still passes tests on Python 2.

samuelhwilliams:
I've made it Py2 compatible but we've had to drop Windows support (:trollface:).

PR #369, opened 2018-03-14
Centralised app _status endpoint

samuelhwilliams:
## Summary
Brings the core logic for serving an app's `_status` endpoint into utils
to share code and reduce replication. Each app should now need to only
call `get_app_status` and pass in the appropriate arguments. The default
status check will validate remaining disk space on the instance as well
as connectivity with the data/search API if applicable. If extra checks
need to be performed, these can be provided using `additional_checks`,
which is a sequence of callables that return dicts with the data that
should be included in the JSON response.

We also add a check for disk usage into the centralised status endpoint.
If disk space is below 5% (by default), we will return a 500 HTTP status
code causing Cloud Foundry to mark the instance as unhealthy and recycle
it.

The test data for this is pretty blegh, but it does provide good coverage and the test _itself_ isn't very hard to follow, hopefully...

## Implementations
* https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/369
* https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/89
* https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/39
* https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/731
* https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/836
* https://github.com/alphagov/digitalmarketplace-user-frontend/pull/46
* https://github.com/alphagov/digitalmarketplace-api/pull/763
* https://github.com/alphagov/digitalmarketplace-search-api/pull/145

## Ticket
https://trello.com/c/BBvu6eUk/383-ensure-healthcheck-endpoint-fails-if-low-disk-space

### DIFF
--- a/dmutils/status.py
+++ b/dmutils/status.py
@@ -2,6 +2,7 @@ import datetime
 import math
 import os
 from .formats import DATE_FORMAT
+from flask import jsonify, current_app
 from flask_featureflags import FEATURE_FLAGS_CONFIG
 
 
@@ -47,3 +48,68 @@ def enabled_since(date_string):
         return date_string
 
     return False
+
+
+class StatusError(Exception):
+    """A stub class to use when implementing additional checks for an app's _status endpoint. See API for example.
+    When raising a StatusError, make sure that the message passed in uniquely identifies the additional check you are
+    performing so that any errors can more easily be tied back to a specific dependency that has failed."""
+    message = ''
+
+    def __init__(self, message):
+        super().__init__(message)
+        self.message = message
+
+
+def get_app_status(data_api_client=None,
+                   search_api_client=None,
+                   ignore_dependencies=False,
+                   additional_checks=None):
+    """Generates output for `_status` endpoints on apps
+
+    :param current_app: The flask `current_app` global.
+    :param data_api_client: The app's data_api_client, if used.
+    :param search_api_client: The app's search_api_client, if used.
+    :param ignore_dependencies: Minimal endpoint checks only (i.e. the app is routable and disk space is fine).
+    :param additional_checks: A sequence of callables that return dicts of data to be injected into the final JSON
+                              response or raise StatusErrors if they need to log an error that should fail the
+                              check (this will cause the endpoint to return a 500).
+    :return: A dictionary describing the current state of the app with, at least, a 'status' key with a value of 'ok'
+             or 'error'.
+    """
+    error_messages = []
+    response_data = {'status': 'ok'}
+
+    disk_status = get_disk_space_status()
+    response_data['disk'] = f'{disk_status[0]} ({disk_status[1]}% free)'
+    if disk_status[0].lower() != 'ok':
+        error_messages.append(f'Disk space low: {disk_status[1]}% remaining.')
+
+    if not ignore_dependencies:
+        response_data['version'] = current_app.config['VERSION']
+        response_data['flags'] = get_flags(current_app)
+
+        if data_api_client:
+            data_api_status = data_api_client.get_status() or {'status': 'n/a'}
+            response_data['api_status'] = data_api_status
+            if data_api_status['status'].lower() != 'ok':
+                error_messages.append('Error connecting to the Data API.')
+
+        if search_api_client:
+            search_api_status = search_api_client.get_status() or {'status': 'n/a'}
+            response_data['search_api_status'] = search_api_status
+            if search_api_status['status'].lower() != 'ok':
+                error_messages.append('Error connecting to the Search API.')
+
+        for additional_check in (additional_checks or []):
+            try:
+                response_data.update(additional_check())
+
+            except StatusError as e:
+                error_messages.append(e.message)
+
+    if error_messages:
+        response_data['status'] = 'error'
+        response_data['message'] = error_messages
+
+    return jsonify(**response_data), 200 if not error_messages else 500
--- a/tests/test_status.py
+++ b/tests/test_status.py
@@ -1,9 +1,10 @@
 from collections import namedtuple
 
+import json
 import mock
 import pytest
 
-from dmutils.status import get_disk_space_status
+from dmutils.status import get_disk_space_status, get_app_status
 
 
 def statvfs_fixture(total_blocks, free_blocks):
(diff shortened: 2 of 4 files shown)


tests/test_status.py
```diff
@@ -27,3 +29,176 @@ def test_disk_space_status(disk_usage, kwargs, total_blocks, free_blocks, expect
...
+                                 ('LOW', 1),
+                                 'error',
+                                 'error',
+                                 True,
+                                 [],
+                                 {'status': 'error', 'disk': 'LOW (1% free)',
+                                  'message': ['Disk space low: 1% remaining.']},
+                                 500,
+                             ),
+                         ))
```
risicle: :grin:

dmutils/status.py
```diff
@@ -47,3 +48,69 @@ def enabled_since(date_string):
...
+             or 'error'.
+    """
+    error_messages = []
+    core_data = {'status': 'ok'}  # Core checks (endpoint resolves, disk is OK)
+    extra_data = {}  # Data from additional checks to run (if ignore_dependencies is False), e.g. api connectivity
+
+    disk_status = get_disk_space_status()
+    core_data['disk'] = '{} ({}% free)'.format(*disk_status)
+    if disk_status[0].lower() != 'ok':
+        error_messages.append(f'Disk space low: {disk_status[1]}% remaining.')
```
risicle: :man_astronaut:

tests/test_status.py
```diff
@@ -27,3 +29,176 @@ def test_disk_space_status(disk_usage, kwargs, total_blocks, free_blocks, expect
...
+    return m
+
+
+@pytest.mark.parametrize('disk_status, '
+                         'data_api_status, '
+                         'search_api_status, '
+                         'ignore_dependencies, '
+                         'additional_checks, '
+                         'expected_json, '
+                         'expected_http_status',
```
risicle: You are actually able to just specify this as a sequence rather than a weird concatenated string (I found out recently)

dmutils/status.py
```diff
@@ -47,3 +48,69 @@ def enabled_since(date_string):
...
 
     return False
+
+
+class StatusError(Exception):
+    """A stub class to use when implementing additional checks for an app's _status endpoint. See API for example."""
+    message = ''
+
+    def __init__(self, message):
+        super().__init__()
```
galund: You should probably pass on `message` to the base constructor, otherwise `str()` conversion doesn't work as expected for an `Exception`.

samuelhwilliams: Done 👍

dmutils/status.py
```diff
@@ -47,3 +48,67 @@ def enabled_since(date_string):
...
+        error_messages.append(f'Disk space low: {disk_status[1]}% remaining.')
+
+    if not ignore_dependencies:
+        extra_data['version'] = current_app.config['VERSION']
+        extra_data['flags'] = get_flags(current_app)
+
+        if data_api_client:
+            data_api_status = data_api_client.get_status() or {}
+            extra_data['api_status'] = data_api_status
+            if data_api_status.get('status').lower() != 'ok':
```
galund: Get will return `None` if the thing isn't there, and `.lower` on `None` won't work, so there's an untested edge-case here and the logic isn't quite right.

tests/test_status.py
```diff
@@ -27,3 +28,175 @@ def test_disk_space_status(disk_usage, kwargs, total_blocks, free_blocks, expect
...
+                                 {'flags': {},
+                                  'api_status': {'status': 'error'},
+                                  'search_api_status': {'status': 'ok'},
+                                  'status': 'error',
+                                  'version': 'release-0',
+                                  'disk': 'OK (90% free)',
+                                  'message': ['Error connecting to the Data API.']},
+                                 500,
+                             ),
+                             (  # Test case #7 - data+search api client ERROR, ignore_deps = False, result result: 500
```
galund: " result result"

dmutils/status.py
samuelhwilliams: Yeah I've just been fixing this up. You snuck in over lunch 🙈

samuelhwilliams: The weirdness is that api client's `get_status` can return None in some edge case (where the apiclient is disabled, all requests return None). The API was testing this specifically so I've come back here to fix it up and you caught me in the middle of that process. I've updated it now to provide a default dict if `get_status` does happen to return None.

dmutils/status.py
```diff
@@ -47,3 +48,67 @@ def enabled_since(date_string):
...
+                extra_data.update(additional_check())
+
+            except StatusError as e:
+                error_messages.append(e.message)
+
+    if error_messages:
+        core_data['status'] = 'error'
+        core_data['message'] = error_messages
+
+    return jsonify(**{**core_data, **extra_data}), 200 if not error_messages else 500
```
galund: If there's a way of doing this that doesn't involve quite so many `**` things then I'd personally find that more readable, but whatevs :)

samuelhwilliams: Ha yeah, there is, sure.

samuelhwilliams: Done. Actually removed a bit of a weird abstraction by just removing `extra_data` altogether.

dmutils/status.py
```diff
@@ -47,3 +48,66 @@ def enabled_since(date_string):
...
+            response_data['search_api_status'] = search_api_status
+            if search_api_status['status'].lower() != 'ok':
+                error_messages.append('Error connecting to the Search API.')
+
+        for additional_check in (additional_checks or []):
+            try:
+                response_data.update(additional_check())
+
+            except StatusError as e:
+                error_messages.append(e.message)
```
galund: (This is all great btw.) So the question remaining in my head is whether we should be making sure an indication of _what_ additional check was being performed, enforced as being in the json output, somehow.

And a follow-on question is about what happens if additional_check causes some other error to occur, but I suppose as long as the end result in that case is a 500, then it's probably fine to not be catching those.

samuelhwilliams: Yeah, at the moment all of the additional checks just output 500s. If we need extra granularity in the future we can update this to take a message and a status code, but for now it's not needed, so I left it out.

samuelhwilliams: Updated the docstring on `StatusError`.

dmutils/status.py
```diff
@@ -47,3 +48,68 @@ def enabled_since(date_string):
...
+
+    if error_messages:
+        response_data['status'] = 'error'
+        response_data['message'] = error_messages
+
+    return jsonify(**response_data), 200 if not error_messages else 500
```
risicle: I realize it's too late but I *much* preferred the `**{**{...}, ...}` method. Much clearer than having side-effect-y branchy code all up the wang up there. Sure, if you don't know python fully, you don't know what it means the first time you see it. But then you look it up. And then you know *totally unambiguously* what pattern it implements and how it's being used there.

samuelhwilliams: It's not too late. 🤔

galund: Do we have to accept that every PR will now be a battle ground between ppl that prefer functional style over OO. Woop, Python.

risicle: The other way is no less oo than otherwise. It's imperative vs functional.

samuelhwilliams: Egh I'm not sure either way. I'm going to leave it as it is as I've removed the extra dict now, so there's nothing to unpack together. I think the extra dict was probably an unneeded bit of fluff here so I'm not sure there's value adding it back.

galund: If you don't allow mutable objects, then it's barely OO in my opinion ;-)

risicle: It's not disallowing mutation, just not encouraging its use where it's not necessary, like doing something that should be straightforward like assembling a dictionary. Mutation should be used where it is part & parcel of what the app is trying to _do_.

Wynndow: Late to the party here, but what was the `**{**{...}, ...}` method?

samuelhwilliams:
```
def echo(**kwargs):
  pass

a = {'a': 1}
a2 = {'a': 2}
b = {'b': 1}

{**a, **b} == {'a': 1, 'b': 1}
{**a, **a2} == {'a': 2}
{**a2, **a} == {'a': 1}

echo(**{**a, **a2}) == echo(a=2)
```

risicle: @Wynndow it's basically a side-effect-free, in-place dictionary merge.

PR #370, opened 2018-03-14
DMNotifyClient: add replacement_email_address mechanism

risicle:
In response to https://trello.com/c/YQB2HkiD/225-functional-tests-send-emails-to-invalid-email-addresses

This should allow a DMNotifyClient to be set up which will make its final call to notify with a fixed destination address - presumably one of notify's special integration test addresses, which we would add to the app config for staging/preview apps.

A few potentially controversial things here: 

 - The sentinel mechanism for the default kwarg
 - The address replacement happens "inside" the resend-detection code, and also has a "reference" based on the *original* address. I chose this because I thought it best to alter the code path as little as possible when we have this feature enabled.
 - Naming of the setting/feature. I'm reasonably flexible on this, but I did consider a number of alternatives before I decided `replacement_email_address` was the most appropriate.

### DIFF
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -18,11 +18,25 @@ class DMNotifyClient(object):
             self,
             govuk_notify_api_key,
             govuk_notify_base_url='https://api.notifications.service.gov.uk',
-            logger=None
+            logger=None,
+            redirect_domains_to_address=None,
     ):
-        """Set up logging and mail client."""
+        """
+            :param govuk_notify_api_key:
+            :param govuk_notify_base_url:
+            :param logger: logger to log progress to, taken from current_app if Falsey
+            :param redirect_domains_to_address: dictionary mapping email domain to redirected email address - emails
+                sent to a email with a domain in this mapping will instead be sent to the corresponding value set here.
+                if `redirect_domains_to_address` is `None` will fall back to looking for a
+                `DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS` setting in the current flask app's config (if available).
+        """
         self.logger = logger or current_app.logger
         self.client = self._client_class(govuk_notify_api_key, govuk_notify_base_url)
+        self._redirect_domains_to_address = (
+            current_app.config.get("DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS")
+            if current_app and redirect_domains_to_address is None else
+            redirect_domains_to_address
+        )
 
     def get_all_notifications(self, **kwargs):
         """Wrapper for notifications_python_client.notifications.NotificationsAPIClient::get_all_notifications"""
@@ -103,8 +117,17 @@ class DMNotifyClient(object):
             )
             return
         try:
+            # NOTE how the potential replacement of the email address happens *after* the has_been_sent check and
+            # reference generation
+            final_email_address = (
+                self._redirect_domains_to_address and self._redirect_domains_to_address.get(
+                    # splitting at rightmost @ should reliably give us the domain
+                    email_address.rsplit("@", 1)[-1].lower()
+                )
+            ) or email_address
+
             response = self.client.send_email_notification(
-                email_address,
+                final_email_address,
                 template_id,
                 personalisation=personalisation,
                 reference=reference,
--- a/tests/email/test_dm_notify.py
+++ b/tests/email/test_dm_notify.py
@@ -5,6 +5,8 @@ import json
 import mock
 import os
 from collections import OrderedDict
+from itertools import product
+
 import pytest
 
 from dmutils.email.dm_notify import DMNotifyClient
@@ -13,12 +15,14 @@ from dmutils.email.dm_notify import DMNotifyClient
 FIXTURES_DIR = os.path.join(os.path.dirname(__file__), 'fixtures')
 
 
+_test_api_key = '1111111111' * 8
+
+
 @pytest.fixture
 def dm_notify_client(app):
     """Supply initialized client."""
     with app.app_context():
-        test_api_key = '1111111111' * 8
-        return DMNotifyClient(test_api_key)
+        return DMNotifyClient(_test_api_key)
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


dmutils/email/dm_notify.py
```diff
@@ -18,11 +22,18 @@ def __init__(
             self,
             govuk_notify_api_key,
             govuk_notify_base_url='https://api.notifications.service.gov.uk',
-            logger=None
+            logger=None,
+            replacement_email_address=FROM_CURRENT_APP,
     ):
-        """Set up logging and mail client."""
+        # Set up logging and mail client.
```
samuelhwilliams: Method docstrings should be in triple quotes even if they're just a one-liner.

tests/email/test_dm_notify.py
```diff
@@ -13,12 +13,14 @@
 FIXTURES_DIR = os.path.join(os.path.dirname(__file__), 'fixtures')
 
 
+_test_api_key = '1111111111' * 8
```
samuelhwilliams: `'1' * 80` might be a tad simpler?

dmutils/email/dm_notify.py
```diff
@@ -18,11 +22,18 @@ def __init__(
             self,
             govuk_notify_api_key,
             govuk_notify_base_url='https://api.notifications.service.gov.uk',
-            logger=None
+            logger=None,
+            replacement_email_address=FROM_CURRENT_APP,
```
samuelhwilliams: This feels clearer if we name it something like `redirect_all_emails_to`.

I think the `FROM_CURRENT_APP` thing need a new name as it really doesn't make sense until you read the condition later on (`if current_app and replacement_email_address is FROM_CURRENT_APP else`). Not sure what a better name is though.

risicle: Yes agreed on both counts

tests/email/test_dm_notify.py
risicle: *shrug* wasn't my line, just transplanted it

dmutils/email/dm_notify.py
risicle: I turned this from a docstring to a comment because it's not a docstring. Docstrings are for external callers to know more about the function. The information here is information the caller doesn't *really* need to know. Or at least if we get into the habit of doing this, we're going to have to start adding docstring descriptions of all methods, describing *exactly* what they do, line by line like this one does.

risicle:
Agreed this does feel weird and it would indeed stop us from getting emails in normal circumstances. I think we need to agree what bits of what instances we want to have a "real" notify account for and which a sandboxed account is ok for.

Again, this is a symptom of using our preview & staging instances for both CI *and* QA. If these instances were CI-only, there would be no issue in enabling this.

dmutils/email/dm_notify.py
samuelhwilliams: Sorry, yes, you're right. I didn't pay enough attention.

dmutils/email/dm_notify.py
```diff
@@ -18,11 +22,18 @@ def __init__(
...
     ):
-        """Set up logging and mail client."""
+        # Set up logging and mail client.
         self.logger = logger or current_app.logger
         self.client = self._client_class(govuk_notify_api_key, govuk_notify_base_url)
+        # determine replacement_email_address
```
samuelhwilliams: This is a bit of a non-comment :trollface:

dmutils/email/dm_notify.py
samuelhwilliams: I think the `FROM_CURRENT_APP` thing can just be None really. This sentinel is kinda confusing to me. I understand what it's there for, but I think with the renamed variable, `None` is a reasonable value to use and simpler to understand in the logic later on.

samuelhwilliams:
Good point re: the double use of our environments. Makes me briefly wonder if we should add a QA environment between preview/integration and staging. Haven't spent enough time on that to consider the real ramifications though.

But yeah, my main issue with this is that _because_ it's baked into the client itself, we have no way to override it for when we are doing QA/manual work, and that's likely to be a not insignificant problem. It also means this code ships out to production and suddenly we have a very easy (albeit unlikely) way of silently disabling emails from being sent to our real users. This makes me super uncomfortable, even though I recognise the fact that the chance of it happening is low.

dmutils/email/dm_notify.py
risicle: Agreed, but was just trying to fit in with the above non-comment.

risicle:
I had an idea about this last night. Instead of doing a single blanket re-addressing, make the setting a mapping of domain -> replacement address. Then we just set it up as

```
{
    "example.com": "simulate-delivered@notifications.service.gov.uk",
    "example.gov.uk": "simulate-delivered@notifications.service.gov.uk"
}
```

because there's no way we would ever expect to be receiving an email sent to either of those domains. And then we just make sure we use either of those above domains in our FTs.

samuelhwilliams:
Nice, I think that feels a lot nicer to me. We still need to remember to only use these domains in the FTs, but that doesn't feel like too big of an issue.

TheDoubleK:
This seems a reasonable solution, and I can't immediately think of any way for it to massively bite us in the ass.  It does feel like maybe "yet another oddity"...

Are we 100% sure this will play nicely with the functional tests that actually check the email that has been sent?  Will they still be able to find what they need to confirm stuff is working?

risicle:
> the functional tests that actually check the email that has been sent?

Oh? Which ones are these?

galund:
Wouldn't it be easier to just change the email address of the accounts we use in our test suite?

risicle:
@galund we were talking about how the various parts of the test suite use emails in different ways and realized that we already have enough constraints on the email addresses we can use in FTs. Email addresses may need to:

 - Look like a gov domain
 - *Not* look like a gov domain
 - Not collide with previous accounts the FTs may have set up (there is at least one place where <pseudorandom>@example.com is used, presumably for this reason)

And keeping track of exactly which uses of the email addresses are likely to end up having notify requests is something that's liable to be error-prone given we've got a dozen people piling on top of this test suite and the code which it is exercising. It also is a notify-specific fix which we'd be applying potentially across our tests, and suppose we found out we also had some other restrictions of emails we could use for mandrill - then we'd have to sit down and separate our uses of addresses into notify-destined ones and mandrill-destined ones, and hope we don't end up with any situations where one address needs to work with both... all seemed like too much of a headache.

This way, all we have to do is remember to just use `@example.com` or `@example.gov.uk` for *all* email addresses in the FTs and we know we're safe.

TheDoubleK:
Sorry for the delay replying... there's a step in functional tests `then I receive a ... email to that address` or something like that.  It looks up Notify to check the email receipt.

eg: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/master/features/buyer/buyer_creation.feature#L20

TheDoubleK:
Using this step definition: https://github.com/alphagov/digitalmarketplace-functional-tests/blob/4dda47c133bfa4dcfdfd914d775ce3739e9ec468/features/step_definitions/emails.rb#L8

risicle:
@TheDoubleK great thanks - think that should work as before - the lie happens after the ref generation code.

samuelhwilliams:
I'd be tempted to do this as a major version bump, even though it isn't a breaking change, so that this new env variable gets documented in the changelog and we can recommend that it's explicitly set in the app config (to None). Otherwise it feels like a mystery env var that could be forgotten about fairly quickly. It would be nice to have it declared in all our apps' config so it's consistent.

samuelhwilliams:
After that I think we should get this in. Ran functional tests with it locally and they seem fine.

risicle:
You mean like a "you probably want to set this to what the other apps have it set to" notice? TBH I was planning on doing a active rollout of this after it's merged rather than passively letting it trickle out.

samuelhwilliams:
Cool, if you're going to force it out to all apps, I'm happy with that then. :+1:

PR #371, opened 2018-03-15
Remove Python2 compatibility

samuelhwilliams:
## Summary
We've done a lot of work migrating our apps and libs to Python 3, so we
are no longer actively looking to support Python 2. The only known
things left running on Python 2 are a couple of scripts in dm-scripts,
but we have tickets to update these to Python 3 in the immediate future
and they are unlikely to be run before that happens.

Our other libs with Python 2 tests running are dm-apiclient, dm-aws, and dm-content-loader.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.7"
   - "3.6"
 install:
   - pip install -r requirements-dev.txt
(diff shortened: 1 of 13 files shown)


samuelhwilliams:
@idavidmcdonald that makes a lot of sense. Done.

PR #372, opened 2018-03-16
Add length to email validator

Wynndow:
For this tech debt ticket: https://trello.com/c/StDVGuoI
See related PR here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/835

### Add default max length validator to EmailField
We weren't validating the length of emails.

Theoretically email should be limited to 254 chars
(https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address),
however using the `+...@` pattern I've sent an email with an address
over 320 chars. Setting a max of 511 should be plenty.

### DIFF
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -1,7 +1,7 @@
 from itertools import product
 
 from flask.ext.wtf import Form
-from wtforms.validators import DataRequired, Optional
+from wtforms.validators import DataRequired, Length, Optional
 from werkzeug.datastructures import ImmutableMultiDict
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/test_forms.py
```diff
@@ -59,6 +59,38 @@ def test_valid_emails(self, app, email_address):
             assert form.validate()
             assert form.data["test_email"] == email_address.strip()
 
+    def test_default_email_length(self, app):
+        email_address = 're' + 498 * 'a' + 'lly@long.com'  # 512 chars long
```
samuelhwilliams: Totally broken. The 'e' should be extended, not the 'a'.

Wynndow: You're totally right. I can only apologise.

Have remedied.

samuelhwilliams: 😂 👍

PR #373, opened 2018-03-16
Fix disk space metric

samuelhwilliams:
## Summary
I somehow completed mangled the disk space by switching around free/used
at the last moment. The method _was_ reporting how much space was _used_
rather than _free_. It now correctly reports how much space is
free/remaining.


samuelhwilliams:
Now 😛

PR #374, opened 2018-03-19
Mitigating "missing key" log message formatting failures

risicle:
This is a rough sketch for mitigating not-very-good log output as described in https://trello.com/c/qLOqQYy0/226-unable-to-format-log-message-log-messages-not-particularly-useful

No tests - I'm not sure it's worth going ahead with - will see what others think.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -123,9 +123,8 @@ class CustomLogFormatter(logging.Formatter):
             logger.info("failed to format log message")
         return msg
 
-class JSONFormatter(BaseJSONFormatter):
-    _format_message_missing_key_attempts = 4
 
+class JSONFormatter(BaseJSONFormatter):
     def process_log_record(self, log_record):
         rename_map = {
             "asctime": "time",
@@ -138,15 +137,18 @@ class JSONFormatter(BaseJSONFormatter):
         log_record['logType'] = "application"
 
         missing_keys = []
-        for attempt in range(self._format_message_missing_key_attempts):
+        for attempt in range(5):
             try:
                 log_record['message'] = log_record['message'].format(**log_record)
+
+                if missing_keys:
+                    logger.warning("Missing keys when formatting log message: {}".format(missing_keys))
+
                 break
+
             except KeyError as e:
                 missing_keys.append(e.args[0])
-                log_record[e.args[0]] = "<missing key>"
-            else:
-                logger.exception("Missing keys when formatting log mesage: {}".format(missing_keys))
+                log_record[e.args[0]] = f"{{{e.args[0]}: missing key}}"
         else:
             logger.exception("Too many missing keys when attempting to format log message: gave up")
 
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -123,8 +123,9 @@ class CustomLogFormatter(logging.Formatter):
             logger.info("failed to format log message")
         return msg
 
-
 class JSONFormatter(BaseJSONFormatter):
+    _format_message_missing_key_attempts = 4
+
     def process_log_record(self, log_record):
         rename_map = {
             "asctime": "time",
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
@risicle have made some tweaks to this and added tests. Would be good to get this in if we can.

tests/test_logging.py
```diff
@@ -135,14 +135,22 @@ def test_log_message_is_unchanged_if_fields_are_not_found(self):
         self.logger.info("hello {bar}")
         result = json.loads(self.buffer.getvalue())
 
-        assert result['message'] == "hello {bar}"
+        assert result['message'] == "hello {bar: missing key}"
```
risicle: Should probably change the name of this test if we're going to do this.

dmutils/logging.py
```diff
@@ -138,15 +137,18 @@ def process_log_record(self, log_record):
         log_record['logType'] = "application"
 
         missing_keys = []
-        for attempt in range(self._format_message_missing_key_attempts):
+        for attempt in range(5):
             try:
                 log_record['message'] = log_record['message'].format(**log_record)
+
+                if missing_keys:
+                    logger.warning("Missing keys when formatting log message: {}".format(missing_keys))
```
risicle: The reason I yanked the equivalent of this line down into an `else` clause is because I wanted to be absolutely certain that any `KeyError` we caught came from the original formatting attempt.

samuelhwilliams: It can't be in an else after you break out of the try, as that else will never happen.

tests/test_logging.py
samuelhwilliams: 👍

dmutils/logging.py
risicle: ahhhhhrrrrr... humf

dmutils/logging.py
```diff
@@ -138,15 +137,18 @@ def process_log_record(self, log_record):
         log_record['logType'] = "application"
 
         missing_keys = []
-        for attempt in range(self._format_message_missing_key_attempts):
+        for attempt in range(5):
```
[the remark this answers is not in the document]
samuelhwilliams: I don't really imagine we will need to change it, but you're right and I didn't really think as I was taking it out. 🙈

dmutils/logging.py
```diff
@@ -125,6 +125,10 @@ def format(self, record):
 
 
 class JSONFormatter(BaseJSONFormatter):
+    def __init__(self, *args, missing_key_attempts=5, **kwargs):
```
idavidmcdonald: Might be nice to call this `max_missing_key_attempts` so it reads a bit clearer later on in the code that it's a boundary

tests/test_logging.py
```diff
@@ -131,18 +131,26 @@ def test_log_message_gets_formatted(self):
...
 
-    def test_failed_log_message_formatting_logs_an_error(self):
+    def test_missing_key_when_formatting_logs_a_warning(self):
         self.logger.info("hello {barry}")
         raw_result = self.dmbuffer.getvalue()
         result = json.loads(raw_result)
 
-        assert result['message'].startswith("failed to format log message")
+        assert result['message'].startswith("Missing keys when formatting log message: ['barry']")
```
idavidmcdonald: Could we get a test for when there are two missing keys please so it's explicit what we expect when more than 1 missing key

tests/test_logging.py
idavidmcdonald: is the test name supposed to be `missing_key1` rather than `missing_key`?

samuelhwilliams:
@idavidmcdonald comments addressed

dmutils/logging.py
```diff
@@ -125,6 +125,10 @@ def format(self, record):
 
 
 class JSONFormatter(BaseJSONFormatter):
+    def __init__(self, *args, max_missing_key_attempts=5, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._max_missing_key_attempts = max_missing_key_attempts
```
risicle: blimey this was just supposed to be an "emergencies only" knob to twiddle, didn't mean for it to spawn a constructor

samuelhwilliams: If we want to support this thing, we might as well do it properly? And pep/flake8 complains if you declare a variable not in the constructor, so 🤷‍♂️

risicle: pfff `# qa`

risicle:
Bah. Didn't bump the version. Nevermind, it'll trickle out.

dmutils/logging.py
idavidmcdonald: 😱Ben will be horrified

PR #376, opened 2018-03-26
Retrieve emails as a generator

samuelhwilliams:
## Summary
We have been getting intermittent 504 timeouts on our DOS email upload
job and they trace back to DMMailChimpClient's
get_email_addresses_from_list job, which tries to get emails back in
batches of 1000. This seems to take too long sometimes, so have reduced
it to a page size of 100 and converted the method to a generator to
reduce overhead.

 ## Ticket
https://trello.com/c/4cK1wTvq/236

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -120,8 +120,7 @@ class DMMailChimpClient(object):
                 success = False
         return success
 
-    def get_email_addresses_from_list(self, list_id, pagination_size=1000):
-        email_addresses = []
+    def get_email_addresses_from_list(self, list_id, pagination_size=100):
         offset = 0
         while True:
             member_data = self.timeout_retry(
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
Hopefully not...

risicle:
We've typically given some kind of hint in the method name when returning an iterator,

samuelhwilliams:
This is fair enough, although I think reading the method name you'd still expect to be given an iterable back - just from the fact it's plural on `email_addresses`. Happy enough to rename it if we think this is for the best, but I don't think it'll make the library more consistent with anything. Might need to be done as a "one fell swoop" sort of thing. 🤔

risicle:
:man_shrugging:

dmutils/email/dm_mailchimp.py
```diff
@@ -130,6 +129,6 @@ def get_email_addresses_from_list(self, list_id, pagination_size=1000):
             if not member_data.get("members", None):
                 break
             offset += pagination_size
-            email_addresses.extend(member["email_address"] for member in member_data["members"])
 
-        return email_addresses
+            for member in member_data["members"]:
+                yield member['email_address']
```
risicle: FWIW if you wanted to show off you could do this as a `yield from (member['email_address'] for member in member_data["members"])`

samuelhwilliams: Aha. I had a feeling something like this existed but didn't follow through on that feeling to actually look it up.

Thanks

PR #377, opened 2018-04-05
tracing: output spanId and parentSpanId in json logs

risicle:
Trello https://trello.com/c/b2pEjCkV/358-zipkin-headers-pt3-logging and https://trello.com/c/UOXf4vZZ/308-make-use-of-zipkin-headers-in-preference-to-request-id

This genericizes the ad-hoc request id injection, pulling arbitrary key/values from the request's extensible `get_extra_log_context()` method.

Also included in here is the addition of the span id header to the response (in the same, slightly non-standard way we include the trace id and previously the request id) to make it easier to piece together corresponding log events from each end of a request.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -11,6 +11,13 @@ from pythonjsonlogger.jsonlogger import JsonFormatter as BaseJSONFormatter
 LOG_FORMAT = '%(asctime)s %(app_name)s %(name)s %(levelname)s ' \
              '%(trace_id)s "%(message)s" [in %(pathname)s:%(lineno)d]'
 
+
+LOG_FORMAT_EXTRA_JSON_KEYS = (
+    "span_id",
+    "parent_span_id",
+)
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -53,11 +60,15 @@ def configure_handler(handler, app, formatter):
     return handler
 
 
+def get_json_log_format():
+    return LOG_FORMAT + "".join(f" %({key})s" for key in LOG_FORMAT_EXTRA_JSON_KEYS)
+
+
 def get_handler(app):
     if app.config.get('DM_PLAIN_TEXT_LOGS'):
         formatter = CustomLogFormatter(LOG_FORMAT)
     else:
-        formatter = JSONFormatter(LOG_FORMAT)
+        formatter = JSONFormatter(get_json_log_format())
 
     if app.config.get('DM_LOG_PATH'):
         handler = logging.FileHandler(app.config['DM_LOG_PATH'])
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -104,15 +104,17 @@ class RequestExtraContextFilter(logging.Filter):
 class CustomLogFormatter(logging.Formatter):
     """Accepts a format string for the message and formats it with the extra fields"""
 
-    FORMAT_STRING_FIELDS_PATTERN = re.compile(r'\((.+?)\)', re.IGNORECASE)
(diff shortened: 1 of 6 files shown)


risicle:
I've got another change to stick on top of this PR so it would be nice to get it merged...

dmutils/logging.py
```diff
@@ -9,7 +9,14 @@
 from pythonjsonlogger.jsonlogger import JsonFormatter as BaseJSONFormatter
 
 LOG_FORMAT = '%(asctime)s %(app_name)s %(name)s %(levelname)s ' \
-             '%(request_id)s "%(message)s" [in %(pathname)s:%(lineno)d]'
+             '%(trace_id)s "%(message)s" [in %(pathname)s:%(lineno)d]'
+
+
+LOG_FORMAT_EXTRA_JSON_KEYS = (
```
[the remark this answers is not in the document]
risicle: It's liable to start getting long, also the included params here aren't likely to be relevant for local deployments, which is our only plain text use case at the moment.

dmutils/logging.py
```diff
@@ -77,16 +88,15 @@ def filter(self, record):
...
+class RequestExtraContextFilter(logging.Filter):
+    """
+        Filter which will pull extra context from the current request's `get_extra_log_context` method (if present)
+        and make this available on log records
+    """
     def filter(self, record):
-        record.request_id = self.request_id
+        if has_request_context() and callable(getattr(request, "get_extra_log_context", None)):
+            for key, value in request.get_extra_log_context().items():
+                setattr(record, key, value or "-")
```
katstevens: Just checking my understanding - if the context value is `None` this will log as `-`? Would `'None'` or `'null'` be more useful for a dev looking through the logs?

😹 I just googled for other examples and someone literally wrote an academic paper on logging nulls: http://thesai.org/Downloads/Volume1No1/Paper_1-A_New_Approach_for_Handling_Null_Values_in_Web_Server_Log.pdf

risicle: Well the reason I chose it is because it's the "standard" way of doing it for line-based logging, and wasn't really going to start hacking either of the formatters to have per-formatter null representation. But I'm not particularly tied to `-`, just wanted to terminate the `None`s *somewhere* lest something downstream isn't robust against it. Thoughts anyone?

risicle: Actually, y'know, I *could* put the substitution in `CustomLogFormatter` and just let the nulls come through to the json output so that we had an easy way of seeing that there's no value. That might be better.

dmutils/logging.py
```diff
@@ -98,11 +108,10 @@ class CustomLogFormatter(logging.Formatter):
 
     def add_fields(self, record):
         for field in self.FORMAT_STRING_FIELDS_PATTERN.findall(self._fmt):
-            record.__dict__[field] = record.__dict__.get(field)
-        return record
+            record.__dict__[field] = record.__dict__.get(field) or "-"
```
galund: I this is maybe obvious but this going to catch `False` and `0` and `0.00` as well as `None` now, the zeroes in particular it might be a bit weird to convert to `-` if they were a real value or something.

risicle: ah balls - zeros.

risicle: See, I hummed and hahed over which of these to use for about a minute but was thinking mostly about wanting to catch explicitly-set `None`s. Forget that.

risicle:
Ok I do actually have approval to merge this so I'm going to in a few minutes unless there are any objections.

PR #378, opened 2018-04-06
Tracing: add handling of zipkin "sampling decision" and "debug flag" headers

risicle:
This sits on top of https://github.com/alphagov/digitalmarketplace-utils/pull/377 and so the diff is spammy, however this exposes and propagates (according to the zipkin spec https://github.com/openzipkin/b3-propagation) the debug and "sample decision" headers.

The two of these have similar intended purposes from what I can tell, with "debug" being a more severe form of "sample decision". It's really up to us what behaviour our apps use this information to trigger, but I suspect what we will end up doing is outputting a bit more log information if "sample decision" is set - basically doing things that won't have a significant effect on performance, and "debug" to output information which would take a bit of extra work to compute. Essentially, using "sample decision" to measure performance and "debug" to debug wrong behaviour. Though, really, as I say, it's up to us to decide.


dmutils/request_id.py
```diff
@@ -46,6 +46,20 @@ def parent_span_id(self):
             self._parent_span_id = self._get_first_header(current_app.config['DM_PARENT_SPAN_ID_HEADERS'])
         return self._parent_span_id
 
+    @property
+    def sampling_decision(self):
```
galund: FWIW I think `is_sampled` (or perhaps `should_sample`) might make more sense, either of these would certainly convey boolean-ness better than `sampling_decision`?

risicle: I don't disagree, but don't really want to go and start inventing my own terms for things.

galund: Hmm, well the English in that zipkin spec seems pretty flakey so I'm not sure I want to go any of its suggestions on language!

The point AIUI is that the sampling _decision_ is made once and then passed on. It's worth noting that the header itself doesn't use the word "decision" so I don't really see why we should be bound to stick with it either.

risicle: Right but I'm also not crazy about the term used for the `User-Agent` header but jumping into a new app/framework I wasn't used to I would expect to find it named "user agent", not for the developer to have made a questionable decision inventing some new & arbitrary name for it.

galund: `sampling_active`
`do_sampling`

I can go all day at this, but `sampling_decision` is still wrong :-)

risicle: ok - going `sampled` because it's the header name.

risicle: Actually, what does everyone else think? Because changing it to `sampled` would mean making the config option `DM_SAMPLED_HEADER` which sounds like it does something completely different.

risicle: ok actually to end it i'm going with `is_sampled`

PR #379, opened 2018-04-12
Clean up utils dependencies

katstevens:
- Upgrades `Flask-WTF` to 0.14.2 (addressing CSRF issues: https://flask-wtf.readthedocs.io/en/latest/changelog.html). This introduces some deprecation warnings (I've fixed them for this repo's tests) but no breaking changes as far as I can see.
- Upgrades `Flask` to 0.10.1 to match all the FE apps. 
- Removes duplicated `Flask` and `Flask-WTF` from `requirements-dev.txt`.

The FE apps and the APIs all pin their `Flask` (and `Flask-WTF`) dependencies in their `requirements-app.txt`, even though the same versions are pulled in via dm-utils. Was there a reason for stating them explicitly in the apps, or can we use dm-utils as the single source of truth instead? Thoughts welcome!

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '36.0.0'
+__version__ = '36.0.1'
--- a/dmutils/forms.py
+++ b/dmutils/forms.py
@@ -39,4 +39,4 @@ class EmailValidator(Regexp):
 
     def __init__(self, **kwargs):
         kwargs.setdefault("message", "Please enter a valid email address.")
-        return super(EmailValidator, self).__init__(self._email_re, **kwargs)
+        super(EmailValidator, self).__init__(self._email_re, **kwargs)
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,6 +11,6 @@ pytest==3.2.3
 pytest-cov==2.5.1
 python-coveralls==2.5.0
 Flask==0.10.1
-Flask-WTF==0.12
+Flask-WTF==0.14.2
 
 git+https://github.com/alphagov/digitalmarketplace-test-utils.git@1.2.0#egg=digitalmarketplace-test-utils==1.2.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,7 +10,5 @@ moto==0.4.31
 pytest==3.2.3
 pytest-cov==2.5.1
 python-coveralls==2.5.0
-Flask==0.10.1
-Flask-WTF==0.14.2
 
 git+https://github.com/alphagov/digitalmarketplace-test-utils.git@1.2.0#egg=digitalmarketplace-test-utils==1.2.0
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,10 @@ setup(
     install_requires=[
          'Flask-FeatureFlags==0.6',
          'Flask-Script==2.0.5',
-         'Flask-WTF==0.12',
-         'Flask>=0.10',
+         'Flask-WTF==0.14.2',
+         'Flask>=0.10.1',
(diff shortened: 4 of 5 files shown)


risicle:
I guess... not *all* wtforms use is done through/using dmutils, and err... maybe (?) we could have a different need for up-to-date-ness in different FEs? But tbh, that would leave us far more vulnerable to forgetting to upgrade one or two FEs dependencies and leaving them behind.

Tis interesting how we're turning `-utils` into essentially `-centralized`.

risicle:
And of course this is where things like `pip` fall apart, because in an ideal world, having a FE-specified version of wtforms that was incompatible with `-utils` form components would be picked up at build-time when the tests failed. But `pip` don't care bout that.

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
 Records breaking changes from major version bumps
 
+## 37.0.0
+
+PR [#379](https://github.com/alphagov/digitalmarketplace-utils/pull/379)
+
+The FE apps should now import Flask dependencies via dmutils. For example:
+
+```python
+from dmutils import flask, flask_featureflags, flask_login, flask_wtf
```
samuelhwilliams: Have just been playing around with this and it feels like it's going to be a pretty big change. Anywhere we do something like `from flask import Markup` will break and we'll have to do `from dmutils import flask` and then reference it as `flask.Markup` everywhere. Either that, or we keep importing directly from `flask` and ignore this (which is what I suspect will happen...), which for us doesn't technically matter as they're the same thing - but semantically it feels like we're making a change that makes things more difficult and will introduce confusion. I'm torn on this one now... 🤔

samuelhwilliams: Maybe we make a new repository called `digitalmarketplace-core` for all of these 'centralised' things :trollface:

benvand: digitalmarketplace-core
_and_
digitalmarketplace-frontend-core :trollface: 💥

benvand: We should be able to do `from dmutils.flask import Markup` though

samuelhwilliams: Hmm, I tried it and that didn't work. 🤷‍♂️ 

I vaguely recall having issues with importing deep dependencies, but I can't really recall the mechanics behind it.

katstevens: How about I split the `__init__` stuff into a separate PR and just have the dependency update in this one?

samuelhwilliams: Yeah I think that's a good idea before this one gets bogged-down.

risicle: I don't *really* understand why we need to import this *through* `dmutils`, just because the dependency is fulfilled by dmutils requirements. There are a good number of places we directly import packages that are second-tier dependencies, I think this sets a weird precedent.

benvand: Where and why do we do that?
Say you have 
Package A that depends on Module B and Module C
Module B depends on Module C

So we remove Module C from Package A's requirements...

What then happens if Module B drops Module C as a dependency?

risicle: Indeed hypothetically it's a wrongness, but it's done a lot because the reality is that most packages don't expose *their* dependencies in a reliable way. An example - we only depend on `Flask-WTF` yet we freely import from `wtforms` - and I'm not totally sure `flask_wtf` actually exposes all of what's accessible through `wtforms`.

benvand: We should import Flask through `dmutils` because that best reflects how we're collecting the package. _Through_ the requirements of `dmutils`.

img_20180413_095657

risicle: Right, but this isn't a pattern that can be followed in the general case, because as I mentioned above, the vast majority of packages in the python world don't follow this strictly enough, so it would be a "rule" we're breaking almost everywhere else.

katstevens:
For now I'm going to merge this with just the dependency update, and leave the `__init__` imports for... another time.

PR #380, opened 2018-04-16
Update dependencies

samuelhwilliams:
## Summary
Some of our libraries now have dependencies on different versions of the
same lib. This commit, and some corresponding commits on our other
repos, will bring them all in line so that we don't get pip complaining
that it cannot meet our conflicting requirements.

Specifically we are setting:
* Werkzeug==0.14.1
* six==1.11.0
* inflection==0.3.1

## Other PRs
https://github.com/alphagov/digitalmarketplace-apiclient/pull/133
https://github.com/alphagov/digitalmarketplace-content-loader/pull/57

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '36.0.1'
+__version__ = '36.0.2'
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
          'boto3==1.4.5',
          'contextlib2==0.4.0',
          'cryptography==1.9',
-         'inflection==0.2.1',
+         'inflection==0.3.1',
          'mailchimp3==2.0.11',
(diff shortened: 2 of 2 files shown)


dmutils/__init__.py
```diff
@@ -4,4 +4,4 @@
 import flask_featureflags  # noqa
 
 
-__version__ = '36.0.1'
+__version__ = '37.0.1'
```
katstevens: Major bump? Is this a breaking change?

samuelhwilliams: Ooh, I think I thought I was going to make it one and then decided not to. It can just be a minor, I think.

setup.py
```diff
@@ -40,8 +40,9 @@
          'python-json-logger==0.1.4',
          'pytz==2015.4',
          'pyyaml==3.11',
-         'six==1.9.0',
+         'six==1.11.0',
          'unicodecsv==0.14.1',
+         'Werkzeug==0.14.1',
```
Wynndow: Why is this new requirement being added?

PR #381, opened 2018-04-16
Remove csrf token utility

samuelhwilliams:
## Summary
FlaskWTF==0.14.2 seems to now include `csrf_token` in a FlaskForm's data
attribute, whereas previously this was not the case. Our API doesn't
like receiving `csrf_token` for some of its validated fields, so we need
to strip it out before passing it across. This will do that.

### DIFF
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -5,7 +5,7 @@ from wtforms.validators import DataRequired, Length, Optional
 from werkzeug.datastructures import ImmutableMultiDict
 import pytest
 
-from dmutils.forms import EmailField
+from dmutils.forms import EmailField, remove_csrf_token
 
 
 class EmailFieldFormatTestForm(FlaskForm):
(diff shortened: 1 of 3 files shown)


tests/test_forms.py
```diff
@@ -141,3 +141,20 @@ def test(self, app, required_field_email, optional_field_email, unspecified_fiel
...
                 assert form.errors["unspecified_email"] == ["Please enter a valid email address."]
+
+
+class TestRemoveCsrfToken:
+    def test_csrf_token_removed_if_present(self):
+        data = {'key': 'value', 'key2': 'value2', 'csrf_token': '123tgredsca2345ywt34rwe'}
+        cleaned_data = remove_csrf_token(data)
+        assert 'csrf_token' not in cleaned_data
+
+    def test_original_data_not_mutated_by_remove_csrf_token(self):
```
Wynndow: To check the data isn't mutated, you could test the value of the csrf key?

Wynndow: Shouldn't the csrf key _not_ be in the data? Should you assert that the other keys are still there?

samuelhwilliams: By mutated I mean that the original dict shouldn't be changed - i.e. the method should make a copy of the data and then remove `csrf_token` only from the copy, leaving the original as-is.

This test checks that the original dict (`data`) still contains `csrf_token` after processing.

PR #382, opened 2018-04-16
Revert Flask-WTF upgrade

samuelhwilliams:
## Summary
When we upgraded to Flask-WTF==0.14 we didn't realise that the way CSRF tokens were generated changed fundamentally, meaning that tokens generated in an app with Flask-WTF==0.14 (e.g. supplier-fe) would not be recognised by a view in another app running Flask-WTF<0.14 (e.g. user-fe). This led to us being unable to logout from the supplier app.

We will need to spend a bit of time working out how we are going to manage this upgrade, but at the moment we will need to remain on Flask-WTF==0.12 until we have worked it out. Our tests will generate deprecation warnings until this is sorted, so we aren't likely to forget.

### DIFF
--- a/dmutils/forms.py
+++ b/dmutils/forms.py
@@ -39,7 +39,7 @@ class EmailValidator(Regexp):
 
     def __init__(self, **kwargs):
         kwargs.setdefault("message", "Please enter a valid email address.")
-        super(EmailValidator, self).__init__(self._email_re, **kwargs)
+        return super(EmailValidator, self).__init__(self._email_re, **kwargs)
 
 
 def remove_csrf_token(data):
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,10 @@ setup(
     install_requires=[
          'Flask-FeatureFlags==0.6',
          'Flask-Script==2.0.5',
-         'Flask-WTF==0.14.2',
-         'Flask>=0.10.1',
+         'Flask-WTF==0.12',
+         'Flask>=0.10',
          'Flask-Login>=0.2.11',
-         'boto3==1.4.5',
+         'boto3==1.4.4',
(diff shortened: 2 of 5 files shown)


dmutils/forms.py
```diff
@@ -39,7 +39,7 @@ class EmailValidator(Regexp):
 
     def __init__(self, **kwargs):
         kwargs.setdefault("message", "Please enter a valid email address.")
-        super(EmailValidator, self).__init__(self._email_re, **kwargs)
+        return super(EmailValidator, self).__init__(self._email_re, **kwargs)
```
Wynndow: Why do we need this return?

setup.py
```diff
@@ -25,10 +25,10 @@
     install_requires=[
          'Flask-FeatureFlags==0.6',
          'Flask-Script==2.0.5',
-         'Flask-WTF==0.14.2',
-         'Flask>=0.10.1',
+         'Flask-WTF==0.12',
+         'Flask>=0.10',
```
Wynndow: Why do we do `>=`. We have Flask pinned at exactly `0.10.1` in the apps.

dmutils/forms.py
samuelhwilliams: We don't, but all I've done here is exactly revert the previous commits, taking us back to exactly where we were before. This should make the changes less contentious, but... /shrug

setup.py
samuelhwilliams: Again, it's an exact revert, so we can deal with these questions at a later date (in theory).

katstevens:
Sorry all - I should have tested this out more thoroughly (all the apps I pulled it into seemed fine, but maybe I wasn't up to date everywhere?).

PR #385, opened 2018-04-23
timing: add logged_duration context manager

risicle:
Here is a generic context manager that can be used to log the time taken to execute a given block of code. There should be enough flexibility built into it to use it both for logging normal events that have a duration we're interested in (calls to various APIs etc) or to use for micro-profiling sections of our production code. When execution leaves the wrapped block, a `condition` is called to determine whether a message should be logged in this particular case.

In the default condition, any current Flask `request` is inspected and a message is only logged if the the zipkin "trace" header has been specified - this would allow us to enable detailed profiling only for chosen requests sent by ourselves, however I fully expect people to be creative with the flexibility given here. Perhaps we could choose to log only blocks whose execution took longer than a certain threshold, or only those that completed successfully (didn't result in an exception).

Example, extremely minimal use:

```
with logged_duration(
    logger=logging.getLogger("boring.thing"),
    message="Spent {duration_real}s doing a boring thing",
):
    do_my()
    boring = thing()
```

**Update:**  those paying attention will notice that I've added another ability to the context manager. The object that the manager *yields* is now the dictionary that will be used to base `log_context` on. This allows the block to annotate the dictionary with e.g. results of the activity, e.g.:

```
with logged_duration(message="Received result {foo} in {duration_real}s") as log_context:
    ...
    log_context["foo"] = do_something()
    ...
```

This should make it more capable of taking over as the general log event emitted by e.g. api calls.

### DIFF
--- a/dmutils/timing.py
+++ b/dmutils/timing.py
@@ -1,7 +1,7 @@
 from contextlib import contextmanager
 import logging
 import sys
-from time import perf_counter, process_time
+import time
 
 from flask import request
 from flask.ctx import has_request_context
@@ -20,7 +20,7 @@ def default_message(log_context):
 
 
 def default_condition(log_context):
-    return has_request_context() and getattr(request, "sampling_decision", False)
+    return has_request_context() and getattr(request, "is_sampled", False)
 
 
 def default_log_func(logger, message, log_level, log_context):
@@ -64,25 +64,25 @@ def logged_duration(
         :param log_level: Numeric log level to use.
         :param condition: Callable accepting a single dictionary argument of the ``log_context``. Should return a
                           boolean signalling whether a log message should be emitted or not. Default condition emits a
-                          log if there is a current Flask request with a True `sampling_decision` attribute. A
+                          log if there is a current Flask request with a True `is_sampled` attribute. A
                           condition of True or None will cause logs to *always* be emitted.
         :param log_func:  The actual logging function which will be called if `condition` passes. Arguments passed are:
                           ``logger``, ``message``, ``log_level`` (all verbatim as passed to ``logged_duration``) and
                           ``log_context``.
     """
-    original_real_time = perf_counter()
+    original_real_time = time.perf_counter()
     # NOTE this is *process* time, not *thread* time. if multiple threads are running in this process it will include
     # their cpu time too. getting *thread* time will be possible in python 3.7+ (but even then it doesn't work on older
     # macos)
-    original_process_time = process_time()
+    original_process_time = time.process_time()
 
     log_context = {}
 
     try:
         yield log_context
     finally:
-        duration_real = perf_counter() - original_real_time
-        duration_process = process_time() - original_process_time
+        duration_real = time.perf_counter() - original_real_time
+        duration_process = time.process_time() - original_process_time
 
         log_context["duration_real"] = duration_real
         log_context["duration_process"] = duration_process
--- a/dmutils/timing.py
+++ b/dmutils/timing.py
@@ -1,0 +1,82 @@
+from contextlib import contextmanager
+import logging
+import sys
+from time import perf_counter, process_time
+
+from flask import request
+from flask.ctx import has_request_context
+
+
+#
+# logged_duration's defaults are exposed here so that a caller is able to defer to the default or simply use a wrapper
+# for the default when customizing the call.
+#
+
+
+def default_message(log_context):
+    return "Block {} in {{duration_real}}s of real-time".format(
+        "executed" if sys.exc_info()[0] is None else "raised {}".format(sys.exc_info()[0].__name__)
+    )
+
+
+def default_condition(log_context):
+    return has_request_context() and getattr(request, "sampling_decision", False)
+
+
+def default_log_func(logger, message, log_level, log_context):
+    final_message = message(log_context) if callable(message) else message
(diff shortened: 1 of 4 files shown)


tests/test_timing.py
```diff
@@ -0,0 +1,305 @@
...
+    yield
+
+
+class SentinelError(Exception):
+    """An exception that can't be mistaken for a genuine problem"""
+    pass
+
+
+def _duration_real_gt_048(log_context):
+    return log_context["duration_real"] > 0.095
```
samuelhwilliams: Method name here doesn't match up with the actual value.

risicle: Well that would be *too* obvious wouldn't it?

dmutils/timing.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def default_message(log_context):
+    return "Block {} in {{duration_real}}s of real-time".format(
+        "executed" if sys.exc_info()[0] is None else "raised {}".format(sys.exc_info()[0].__name__)
+    )
+
+
+def default_condition(log_context):
+    return has_request_context() and getattr(request, "sampling_decision", False)
```
samuelhwilliams: `sampling_decision` feels like a weird phrase to me. Does it have an origin I'm not aware of or is it something you've come up with yourself?

dmutils/timing.py
```diff
@@ -0,0 +1,91 @@
...
+from time import perf_counter, process_time
+
+from flask import request
+from flask.ctx import has_request_context
+
+
+#
+# logged_duration's defaults are exposed here so that a caller is able to defer to the default or simply use a wrapper
+# for the default when customizing the call.
+#
```
risicle: Oh and this is something I'd like opinions on - I'm tempted here to, rather than exposing these defaults as I'm currently doing, setting them as _attributes_ on the function itself. The advantage being that to make use of them you don't have to do any additional imports - just importing the function is enough. It also has the advantage that they are more strongly bound to `logged_duration` - I can imagine here adding more tools to `timing.py` and then being in a situation where it's like "`default_message` for *what*?" But I don't really want to have to lengthen the names any more than I have to...

Thoughts?

dmutils/timing.py
galund: FWIW I've made a couple of suggestions on that PR.

dmutils/timing.py
```diff
@@ -0,0 +1,91 @@
...
+                ...
+
+        Logging action is triggered whether execution fell out of the block naturally or the code raised an exception.
+        If log messages are not desired on exeptions these can always be excluded by a custom ``condition``.
+
+        Can also be used as a function decorator.
+
+        :param logger:    The logger to log the message to. It's best to set this at the very least to give the log
+                          reader a clue as to what block of code is being timed.
+        :param message:   Message format string to emit. Can be a callable which should return the desired message
```
galund: Is this a standard thing, cos I find it a bit weird - wouldn't it be more clear to have two alternative parameters, i.e. `message` and `message_factory` or something? I'm not sure I really understand the use case for having a function that generates the message based a context anyway..

risicle: Yes I've seen it quite a lot, notably in the `mock` module. When you have multiple parameters you've got to add logic around which overrides which etc. For a use case see the default log message (function). That's the case I added it for because otherwise I'd always be implying that the block succeeded.

tests/test_timing.py
```diff
@@ -0,0 +1,336 @@
...
+import re
+import sys
+from time import sleep
+
+from flask import request
+import pytest
+
+
+class malleable_ANY:
+    def __init__(self, condition):
```
galund: You should probably explain here why this is useful.

risicle: *Ah* - actually I wasn't going to because I only wanted this to be here temporarily before I promoted these to `-test-utils` where they *would* be documented properly.

galund: 👍

samuelhwilliams:
Let me know when this is good for a final review.

risicle:
Dare I say now?

samuelhwilliams:
🤔  So I've added the following to the suppliers dashboard (localhost/suppliers), but I'm not getting the message printed on my logs.

```
        with logged_duration(message="Received result {supplier} in {duration_real}s",
                             logger=current_app.logger,
                             log_level=logging.INFO) as log_context:
            supplier = data_api_client.get_supplier(
                current_user.supplier_id
            )['suppliers']
            log_context['supplier'] = supplier
```

What am I doing wrong?

tests/test_timing.py
```diff
@@ -0,0 +1,545 @@
...
+def mock_time_functions():
+    with mock.patch.multiple("time", perf_counter=mock.DEFAULT, process_time=mock.DEFAULT) as time_mocks:
+        # to simulate the behaviour of these time functions, we create them all as closures capturing these common
+        # "state" variables tracking the apparent pseudo-time according to each sampling function
+        perf_counter_state = random.uniform(1000, 10000)
+        process_time_state = random.uniform(1000, 10000)
+
+        # the sampling functions just return the current state
+        time_mocks["perf_counter"].side_effect = lambda: perf_counter_state
+        time_mocks["process_time"].side_effect = lambda: process_time_state
```
samuelhwilliams: What's the difference between using `side_effect` with a lambda here, and just setting `return_value`?

PR #386, opened 2018-04-24
Add formatters

samuelhwilliams:
## Summary
We are going to store important framework lifecycle datetimes on the
API, so we need to be able to format these correctly for display on the
frontend. We currently use a different display format for these, so this
adds a formatter that will give the desired display e.g. '5pm BST,
Monday 1 July 2010'.

Also adds a second formatter `monthyearformatter`, currently used to
display the month and year when a supplier completed a previous
declaration when applying to a framework.

Also moves some style guide formatting to a more central function so

Bump version to 36.8.0

 ## Ticket
https://trello.com/c/BVPgZRZc/65-store-datetimes-relating-to-the-framework-lifecycle-on-the-api

### DIFF
--- a/dmutils/dates.py
+++ b/dmutils/dates.py
@@ -63,4 +63,5 @@ def update_framework_with_formatted_dates(framework):
         'applicationsCloseAt': utctoshorttimelongdateformat(framework['applicationsCloseAtUTC']),
         'intentionToAwardAt': dateformat(framework['intentionToAwardAtUTC']),
         'frameworkLiveAt': dateformat(framework['frameworkLiveAtUTC']),
+        'frameworkExpiresAt': dateformat(framework['frameworkExpiresAtUTC']),
     })
--- a/dmutils/dates.py
+++ b/dmutils/dates.py
@@ -1,7 +1,7 @@
 from datetime import datetime, timedelta
 from workdays import workday
 
-from dmutils.formats import dateformat, utctoshorttimelongdateformat
+from .formats import dateformat, utctoshorttimelongdateformat
 
 
 def get_publishing_dates(brief):
--- a/dmutils/dates.py
+++ b/dmutils/dates.py
@@ -1,6 +1,8 @@
 from datetime import datetime, timedelta
 from workdays import workday
 
+from dmutils.formats import dateformat, utctoshorttimelongdateformat
+
 
 def get_publishing_dates(brief):
     APPLICATION_OPEN_DAYS = {'1 week': 7, '2 weeks': 14}
@@ -48,3 +50,17 @@ def _get_todays_date():
 
 def _get_length_of_brief(brief):
     return brief['requirementsLength'] if brief.get('requirementsLength') else '2 weeks'
+
+
+def update_framework_with_formatted_dates(framework):
+    """Takes a framework record (as returned by eg data_api_client.get_framework) and inserts new keys
+    for any date/time values that should be displayed consistently on the frontends.
+
+    This is currently used for important framework lifecycle/application dates."""
+    framework.update({
+        'clarificationsCloseAt': utctoshorttimelongdateformat(framework['clarificationsCloseAtUTC']),
+        'clarificationsPublishAt': utctoshorttimelongdateformat(framework['clarificationsPublishAtUTC']),
+        'applicationsCloseAt': utctoshorttimelongdateformat(framework['applicationsCloseAtUTC']),
+        'intentionToAwardAt': dateformat(framework['intentionToAwardAtUTC']),
+        'frameworkLiveAt': dateformat(framework['frameworkLiveAtUTC']),
+    })
--- a/dmutils/formats.py
+++ b/dmutils/formats.py
@@ -9,6 +9,7 @@ DISPLAY_SHORT_DATE_FORMAT = '%-d %B'
 DISPLAY_DATE_FORMAT = '%A %-d %B %Y'
 DISPLAY_TIME_FORMAT = '%H:%M:%S'
 DISPLAY_DATETIME_FORMAT = '%A %-d %B %Y at %I:%M%p %Z'
+DISPLAY_SHORTTIME_LONGDATE_FORMAT = '%-I%p %Z, %A %-d %B %Y'
 
 
 def timeformat(value, default_value=None):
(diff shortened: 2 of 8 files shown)


Wynndow:
Needs a version bump.

dmutils/formats.py
```diff
@@ -65,11 +66,26 @@ def datetimeformat(value, default_value=None, localize=True):
...
     """
-    formatted_date = _format_date(value, default_value, DISPLAY_DATETIME_FORMAT, localize=localize)
-    if formatted_date:
-        # en_GB locale uses uppercase AM/PM which contravenes our style guide
-        return formatted_date.replace('AM', 'am').replace('PM', 'pm').replace(" 0", " ")
-    return formatted_date
+    return _format_date(value, default_value, DISPLAY_DATETIME_FORMAT, localize=localize)
+
+
+def utctoshorttimelongdateformat(value, default_value=None, localize=True):
```
Wynndow: It feels a bit weird that all the functions in here don't use snake case for their name like we do everywhere else.

dmutils/formats.py
```diff
@@ -121,12 +137,21 @@ def _use_gmt_timezone(date_string):
...
-        return default_value
-    if not isinstance(value, datetime):
-        value = datetime.strptime(value, DATETIME_FORMAT)
-    if value.tzinfo is None:
-        value = pytz.utc.localize(value)
-    if localize:
-        return value.astimezone(EUROPE_LONDON).strftime(fmt)
+        value = default_value
+
     else:
```
Wynndow: Can this `else:` be dropped?

tests/test_formats.py
```diff
@@ -66,6 +66,24 @@ def test_datetimeformat(dt, formatted_datetime):
...
+    (datetime(2012, 8, 10, 9, 8, 7, 6, tzinfo=pytz.utc), "10am BST, Friday 10 August 2012"),
+    (datetime(2012, 8, 1, 9, 8, 7, 6, tzinfo=pytz.utc), "10am BST, Wednesday 1 August 2012"),
+    (datetime(2012, 8, 1, 22, 59, 7, 6, tzinfo=pytz.utc), "11pm BST, Wednesday 1 August 2012"),
+    # Daylight savings edge case
+    (datetime(2012, 3, 25, 0, 59, 7, 6, tzinfo=pytz.utc), "12am GMT, Sunday 25 March 2012"),
+    (datetime(2012, 3, 25, 1, 59, 7, 6, tzinfo=pytz.utc), "2am BST, Sunday 25 March 2012"),
+    # Fall back to default if no valid date supplied
+    (None, None),
+))
+def test_utctoshorttimelongdateformat(dt, formatted_datetime):
```
Wynndow: Worth testing that the `default_value` and `localize` parameters work as expected when not left as default values?

dmutils/formats.py
katstevens: I think it's an old convention to distinguish it as a template filter - none of the Jinja built-in filters use snake case, for example. 

The old Django docs have disappeared off the internet now, but I think once upon a time Django custom template filters had naming restrictions because of how they were registered/scoped within the app. But I'm pretty sure there's no practical restriction on doing it there now. Or here!

Wynndow: ah, interesting. Thanks Kat!

dmutils/dates.py
```diff
@@ -1,6 +1,8 @@
 from datetime import datetime, timedelta
 from workdays import workday
 
+from dmutils.formats import dateformat, utctoshorttimelongdateformat
```
Wynndow: Could just be `from .formats...`

samuelhwilliams: I mean yeah, it can... /shrug

Wynndow: An efficiency saving of 7 bytes.

samuelhwilliams: Totally worth doing. Done!

dmutils/formats.py
```diff
@@ -5,10 +5,12 @@
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 DATE_FORMAT = "%Y-%m-%d"
+DISPLAY_MONTH_YEAR_FORMAT = "%B %Y"
 DISPLAY_SHORT_DATE_FORMAT = '%-d %B'
 DISPLAY_DATE_FORMAT = '%A %-d %B %Y'
 DISPLAY_TIME_FORMAT = '%H:%M:%S'
 DISPLAY_DATETIME_FORMAT = '%A %-d %B %Y at %I:%M%p %Z'
+DISPLAY_SHORTTIME_LONGDATE_FORMAT = '%I%p %Z, %A %-d %B %Y'
```
Wynndow: If you changed this to '%-I%p %Z, %A %-d %B %Y' you could drop the `startswith('0'):` check in `utctoshorttimelongdateformat `, or do you need it available left padded elsewhere?

samuelhwilliams: Ah nice, yep.

dmutils/dates.py
```diff
@@ -48,3 +50,18 @@ def _get_todays_date():
 
 def _get_length_of_brief(brief):
     return brief['requirementsLength'] if brief.get('requirementsLength') else '2 weeks'
+
+
+def update_framework_with_formatted_dates(framework):
```
Wynndow: Is this perhaps a bit prescriptive? What if for some reason we want to display the same date differently in two places? As we have the Jinja filters, can we not just handle the conversion on the frontend?

Wynndow: Although I guess you still have the unformatted date in the record. Why do we need this method then?

samuelhwilliams: It is a little prescriptive, you're right. But it will hopefully keep these things consistent. When we were using the `dates.yml` file, we knew that everywhere we displayed that date it would be in a single format. This helps us achieve the same consistency, but we also have the flexibility to use the raw UTC datetime to get to any other format if we _need_ to.

Having this here helps us be consistent unless we have a good reason not to be, I think.

samuelhwilliams: If you don't agree with that logic though, we could talk about taking it out. I'm not totally attached to this, but it made sense to me insofar as trying to preserve the existing format as much as possible.

Wynndow: I guess if we know we're going to be using this particular set of dates in this particular format a bunch, then having this makes sense. If it's only going to be used in one place I think it could be a bit unnecessary. However we've got it now, so lets keep it!

PR #388, opened 2018-04-30
Move api stubs here from dm-apiclient

samuelhwilliams:
## Summary
API stubs used to be defined in the apiclient, however they now require
some formatting that's held in this repository. Rather than introduce an
extra dependency to the apiclient, which would likely introduce
dependency conflicts when both libs are required by a frontend app,
we'll move the stubs to utils.

Bump to 36.6.0.

 ## Ticket
https://trello.com/c/BVPgZRZc/65-store-datetimes-relating-to-the-framework-lifecycle-on-the-api

### DIFF
--- a/dmutils/api_stubs.py
+++ b/dmutils/api_stubs.py
@@ -3,12 +3,64 @@ from datetime import datetime as dt
 from .formats import DATETIME_FORMAT
 
 
-def framework(status="open",
+def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_limit=False, unit_singular='service',
+        unit_plural='services'):
+    if not name:
+        name = slug.replace("-", " ").title()
+
+    return {
+        "id": lot_id,
+        "slug": slug,
+        "name": name,
+        "allowsBrief": allows_brief,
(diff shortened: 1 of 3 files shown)


dmutils/api_stubs.py
```diff
@@ -0,0 +1,156 @@
+from datetime import datetime as dt
+
+from dmutils.formats import DATETIME_FORMAT
```
Wynndow: Should this just be `from .formats import ...`

tests/test_api_stubs.py
```diff
@@ -0,0 +1,437 @@
+from datetime import datetime as dt
+from dmapiclient import api_stubs
```
Wynndow: Not importing from dmapiclient anymore.

samuelhwilliams:
Addressed.

samuelhwilliams:
@Wynndow Some updates here you may want to glance over, although hopefully nothing too contentious. Added framework agreement details to the framework stub and made the lot a bit more customisable.

Wynndow:
@samuelhwilliams 👍

PR #390, opened 2018-05-02
Handle MailChimp responses with naughty json

Wynndow:
If we get a RequestException from the MailChimp client when calling
`subscribe_new_email_to_list`, we inspect the response which is
integrated into the error. We do this is a special case we handle
depending on the reason for the failure. This includes looking at the
json data. If the response doesn't have any data (which we've seen), or
the data is malformed, we get a JSONDecodeError. We should handle this
properly as it's already thrown one of our scripts off.

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Digital Marketplace MailChimp integration."""
 
+from json.decoder import JSONDecodeError
 from hashlib import md5
 from requests.exceptions import RequestException, HTTPError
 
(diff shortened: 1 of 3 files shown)


dmutils/email/dm_mailchimp.py
```diff
@@ -93,17 +94,20 @@ def subscribe_new_email_to_list(self, list_id, email_address):
...
+                    "looks fake or invalid, please enter a real email address." in e.response.json().get("detail", "")
+                ):
+                    self.logger.error(
+                        "Expected error: Mailchimp failed to add user ({}) to list ({}). API error: The email address looks fake or invalid, please enter a real email address.".format(  # noqa
+                            hashed_email,
+                            list_id
+                        ),
+                        extra={"error": str(e)}
+                    )
+                    return True
```
galund: I'm confused by what this function returns. Why True in this case?

Wynndow: I'll have a look, but is the logic not 'Return False in all failure cases except this one special reason, then return True'? Which is what it still does?

galund: Yeah ok I think I can see how that can make sense now :-)

Wynndow: Cool. I'll still have a dig though. Would be good to know.

PR #392, opened 2018-05-10
Password reset token timing

katstevens:
Trello: https://trello.com/c/TieWrBG5/44-password-reset-link-from-logged-in-session

When a user changes their password, a confirmation email is sent containing a reset link (alerting them in case the change was made by a malicious user). 

The token used in that link is created within a few milliseconds of the password change, however  `decode_password_reset_token` currently requires the most recent password change to be at least 1 second prior to that token's generation time. 

Therefore we need to relax the restriction by 1 second, to enable a valid email link to be generated immediately following a password change.

(I also tidied `test_tokens` into classes, as the module was getting confusing.)

### DIFF
--- a/tests/email/test_tokens.py
+++ b/tests/email/test_tokens.py
@@ -126,6 +126,7 @@ class TestDecodePasswordReset:
 
     @pytest.mark.parametrize(
         'decode_time, expected_result', [
+            ('2016-01-02 03:04:04', 'error'),  # Cannot decode before the token has been generated
             ('2016-01-02 03:04:05', 'ok'),
             ('2016-01-03 03:04:05', 'ok'),
             ('2016-01-03 03:04:06', 'error'),
@@ -138,7 +139,6 @@ class TestDecodePasswordReset:
             with freeze_time('2016-01-02 03:04:05'):
                 token = generate_token(password_reset_token, "Key", 'PassSalt')
 
-        with email_app.app_context():
             with freeze_time(decode_time):
                 if expected_result == 'ok':
                     assert decode_password_reset_token(token, data_api_client) == password_reset_token
--- a/tests/email/test_tokens.py
+++ b/tests/email/test_tokens.py
@@ -41,200 +41,195 @@ def password_reset_token():
     return {'user': 123, 'email': 'test@example.com'}
 
 
-def test_can_generate_and_decode_token():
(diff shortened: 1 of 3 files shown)


tests/email/test_tokens.py
```diff
@@ -41,177 +41,205 @@ def password_reset_token():
...
+            self, email_app, data_api_client, password_reset_token):
+        token = generate_token(password_reset_token, "Key", 'PassSalt')[1:]
 
-    with freeze_time('2016-01-02 03:04:05'):
         with email_app.app_context():
             assert decode_password_reset_token(token, data_api_client) == {'error': 'token_invalid'}
 
+    @pytest.mark.parametrize(
+        'decode_time, expected_result', [
+            ('2016-01-02 03:04:05', 'ok'),
```
samuelhwilliams: Would be good to test `('2016-01-02 03:04:04', 'error'),` as well.

tests/email/test_tokens.py
```diff
@@ -41,177 +41,205 @@ def password_reset_token():
...
-):
-    with freeze_time('2016-01-01T11:00:00.30Z'):
-        # Token was generated an hour earlier than password was changed
-        token = generate_token(password_reset_token, "Key", 'PassSalt')
-
-    with freeze_time('2016-01-01T13:00:00.30Z'):
-        # Token is two hours old; password was changed an hour ago
         with email_app.app_context():
-            assert decode_password_reset_token(token, data_api_client) == {'error': 'token_invalid'}
+            with freeze_time(decode_time):
```
samuelhwilliams: Does this need to be in a different `email_app.app_context`?

katstevens: Good question. I'll try it out.

tests/email/test_tokens.py
katstevens: Interesting! We don't fail when the token is generated in the future (i.e. after we decode the token... that we are about to generate). I... don't know how this would happen in practice?

katstevens: I've added in an extra line to make it explicit that this shouldn't happen.

katstevens: I think we're OK - the Fernet timestamp is UTC (https://github.com/fernet/spec/blob/master/Spec.md), and we're using UTC everywhere too.

PR #393, opened 2018-05-11
Wrap any and all errors ocurring in mandrill code

galund:
We no longer restrict to just mandrill's own error class, as in fact that library can throw other exceptions, and we want to wrap all of them.

https://trello.com/c/s7rmFSjk/295-mandrill-email-crash-2018-05-10-1824

### DIFF
--- a/tests/email/test_dm_mandrill.py
+++ b/tests/email/test_dm_mandrill.py
@@ -3,7 +3,6 @@
 
 import mock
 import pytest
-from mandrill import Error
 
 from dmutils.config import init_app
 from dmutils.email.dm_mandrill import send_email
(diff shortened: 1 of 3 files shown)


tests/email/test_dm_mandrill.py
```diff
@@ -139,8 +138,8 @@ def test_calls_send_email_with_alternative_reply_to(email_app, mandrill):
 
 def test_should_throw_exception_if_mandrill_fails(email_app, mandrill):
     with email_app.app_context():
-
-        mandrill.messages.send.side_effect = Error('this is an error')
+        something_bad = Exception('this is an error')
```
samuelhwilliams: Might be nice to parametrize this test to check for `JSONDecodeError` (as well as the generic `Exception`), as this is what we know we want to catch in this instance.

galund: Yep, might look a bit weird because everyone looking at it will know that if the behaviour is as expected for Exception, it'll also be so for JSONDecodeError, but if you think it makes things more clear then happy to.

galund:
Okay, how does that look @samuelhwilliams ?

Thanks @CrystalPea for offering to get this pushed out to all the apps.

CrystalPea:
@galund no worries, please just slack me when it's merged ☺️

PR #394, opened 2018-05-17
Log duration of calls to S3

samuelhwilliams:
## Summary
We think we have some performance issues with calls to S3 on certain
views. This replaces the existing logs for these calls and inserts the
duration of the call so that we can start to investigate the issue more
thoroughly.

It also adds logs/durations to any calls out to S3 that were missing
them, with a more restrictive condition that the call must be 'slow' or
sampled.

This PR also defines a 'slow' request to an external provider as one
that takes 250ms or longer.

---

Open to conversation on where to draw the line for a 'slow' request, but this feels like a reasonable (if not generous) limit to set in the first instance.

### DIFF
--- a/tests/test_timing.py
+++ b/tests/test_timing.py
@@ -110,16 +110,44 @@ _messages_expected = OrderedDict((
     (
         timing.logged_duration.default_message,
         (
-            AnyStringMatching(r"Block (executed|raised \w+) in \{duration_real\}s of real-time"),
-            AnyStringMatching(r"Block (executed|raised \w+) in [0-9eE.-]+s of real-time"),
+            {
+                'success': "Block executed in {duration_real}s of real-time",
+                'error': AnyStringMatching(r"Block raised \w+ in \{duration_real\}s of real-time"),
+            },
+            {
+                'success': AnyStringMatching(r"Block executed in [0-9eE.-]+s of real-time"),
+                'error': AnyStringMatching(r"Block raised \w+ in [0-9eE.-]+s of real-time"),
+            },
+        ),
+    ),
+    (
+        timing.different_message_for_success_or_error(
+            success_message='Block succeeded in {duration_real}s',
+            error_message='Block raised {exc_info[0]} in {duration_real}s',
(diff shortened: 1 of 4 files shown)


risicle:
Updating the `log_context` dict *after* the s3 call means that the information won't get passed through in the exception case. On that matter, if you're going to have an unconditional log message it should probably try and cope better with the exception case and not output a line that implies it succeeded. Or just skip the exception case by checking if `sys.exc_info()[0] is None`.

risicle:
(wonder if there's a good solution to making a log line "explain" *why* it logged - in order to avoid situations where someone's going "wtf well this time the view ran it didn't make that call at all" not realizing the 250ms thing...)

samuelhwilliams:
Have re-ordered to insert the context before the call to S3, good spot.

risicle:
(...**whatif** the return value of the condition callable were exposed as a `logReason`? The truthiness of the value would still serve to signal the decision, signalling "no" would just be like returning a falsey "log reason")

risicle:
> good spot

:dog2:

samuelhwilliams:
@risicle added a function that allows logging different messages based on success/failure and moved the `logged_duration` call to the correct part of the `s3.list` call, as the earlier parts are just generators.

samuelhwilliams:
@risicle updated tests

tests/test_timing.py
```diff
@@ -110,16 +110,44 @@ def _default_and_no_exception(log_context):
...
+        ),
+    ),
+    (
+        timing.different_message_for_success_or_error(
+            success_message='Block succeeded in {duration_real}s',
+            error_message='Block raised {exc_info[0]} in {duration_real}s',
+        ),
+        (
+            {
+                'success': AnyStringMatching(r"Block succeeded in \{duration_real\}s"),
```
risicle: you can just use a literal string here, rite? 'snot actually using any re features. Same on line 114.

samuelhwilliams: Urrr yep.

PR #395, opened 2018-05-17
Log duration of email calls

samuelhwilliams:
## Summary
We now log the duration of calls to AWS S3, but we should expand this to
our other third party providers. This PR introduces `logged_duration`
wrappers around the main calls to Mailchimp, Mandrill, and Notify. It
also introduces a generic partial that can be used to quickly wrap calls
to third-party providers and emit a consistent log entry, which involved
refactoring the current S3 logged_duration calls slightly.

 ## Ticket
https://trello.com/c/M0hAGLYT/303

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -7,6 +7,8 @@ from requests.exceptions import RequestException, HTTPError
 
 from mailchimp3 import MailChimp
 
+from dmutils.timing import logged_duration_for_external_request as log_external_request
+
 PAGINATION_SIZE = 1000
 
 
@@ -33,7 +35,8 @@ class DMMailChimpClient(object):
         def wrapper(*args, **kwargs):
             for i in range(1 + self.retries):
                 try:
-                    return method(*args, **kwargs)
+                    with log_external_request(service='Mailchimp'):
+                        return method(*args, **kwargs)
                 except HTTPError as e:
                     exception = e
                     if exception.response.status_code == 504:
@@ -45,12 +48,13 @@ class DMMailChimpClient(object):
 
     def create_campaign(self, campaign_data):
         try:
-            campaign = self._client.campaigns.create(campaign_data)
+            with log_external_request(service='Mailchimp'):
+                campaign = self._client.campaigns.create(campaign_data)
             return campaign['id']
         except RequestException as e:
             self.logger.error(
-                "Mailchimp failed to create campaign for '{0}'".format(
-                    campaign_data.get("settings").get("title")
+                "Mailchimp failed to create campaign for '{campaign_title}'".format(
+                    campaign_title=campaign_data.get("settings", {}).get("title")
                 ),
                 extra={"error": str(e)}
             )
@@ -58,7 +62,8 @@ class DMMailChimpClient(object):
 
     def set_campaign_content(self, campaign_id, content_data):
         try:
-            return self._client.campaigns.content.update(campaign_id, content_data)
+            with log_external_request(service='Mailchimp'):
+                return self._client.campaigns.content.update(campaign_id, content_data)
         except RequestException as e:
             self.logger.error(
                 "Mailchimp failed to set content for campaign id '{0}'".format(campaign_id),
@@ -68,7 +73,8 @@ class DMMailChimpClient(object):
 
     def send_campaign(self, campaign_id):
         try:
-            self._client.campaigns.actions.send(campaign_id)
+            with log_external_request(service='Mailchimp'):
+                self._client.campaigns.actions.send(campaign_id)
             return True
         except RequestException as e:
             self.logger.error(
@@ -81,14 +87,15 @@ class DMMailChimpClient(object):
         """ Will subscribe email address to list if they do not already exist in that list else do nothing"""
         hashed_email = self.get_email_hash(email_address)
         try:
-            return self._client.lists.members.create_or_update(
-                list_id,
-                hashed_email,
-                {
-                    "email_address": email_address,
-                    "status_if_new": "subscribed"
-                }
-            )
+            with log_external_request(service='Mailchimp'):
+                return self._client.lists.members.create_or_update(
+                    list_id,
+                    hashed_email,
+                    {
+                        "email_address": email_address,
+                        "status_if_new": "subscribed"
+                    }
+                )
         except RequestException as e:
             # As defined in mailchimp API documentation, this particular error message may arise if a user has requested
             # mailchimp to never add them to mailchimp lists. In this case, we resort to allowing a failed API call (but
@@ -99,10 +106,8 @@ class DMMailChimpClient(object):
                     "looks fake or invalid, please enter a real email address." in e.response.json().get("detail", "")
                 ):
                     self.logger.error(
-                        "Expected error: Mailchimp failed to add user ({}) to list ({}). API error: The email address looks fake or invalid, please enter a real email address.".format(  # noqa
-                            hashed_email,
-                            list_id
-                        ),
+                        f"Expected error: Mailchimp failed to add user ({hashed_email}) to list ({list_id}). "
+                        "API error: The email address looks fake or invalid, please enter a real email address.",
                         extra={"error": str(e)}
                     )
                     return True
@@ -120,8 +125,9 @@ class DMMailChimpClient(object):
     def subscribe_new_emails_to_list(self, list_id, email_addresses):
         success = True
         for email_address in email_addresses:
-            if not self.subscribe_new_email_to_list(list_id, email_address):
-                success = False
+            with log_external_request(service='Mailchimp'):
+                if not self.subscribe_new_email_to_list(list_id, email_address):
+                    success = False
         return success
 
     def get_email_addresses_from_list(self, list_id, pagination_size=100):
--- a/dmutils/email/dm_mandrill.py
+++ b/dmutils/email/dm_mandrill.py
@@ -6,6 +6,7 @@ from flask._compat import string_types
 from mandrill import Mandrill
 from dmutils.email.exceptions import EmailError
 from dmutils.email.helpers import hash_string
+from dmutils.timing import logged_duration_for_external_request as log_external_request
 
 
 def send_email(to_email_addresses, email_body, api_key, subject, from_email, from_name, tags, reply_to=None,
@@ -40,12 +41,15 @@ def send_email(to_email_addresses, email_body, api_key, subject, from_email, fro
             } for email_address in to_email_addresses]
         }
 
-        result = mandrill_client.messages.send(message=message, async=True)
+        with log_external_request(service='Mandrill', logger=logger):
+            result = mandrill_client.messages.send(message=message, async=True)
+
     except Exception as e:
         # Anything that Mandrill throws will be rethrown in a manner consistent with out other email backends.
         # Note that this isn't just `mandrill.Error` exceptions, because the mandrill client also sometimes throws
         # things like JSONDecodeError (sad face).
         logger.error("Failed to send an email: {error}", extra={'error': e})
         raise EmailError(e)
+
     logger.info("Sent {tags} response: id={id}, email={email_hash}",
                 extra={'tags': tags, 'id': result[0]['_id'], 'email_hash': hash_string(result[0]['email'])})
--- a/dmutils/timing.py
+++ b/dmutils/timing.py
@@ -1,4 +1,5 @@
 from contextlib import contextmanager
+import inspect
 import logging
 import sys
 import time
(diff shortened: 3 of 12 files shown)


dmutils/timing.py
```diff
@@ -118,3 +99,58 @@ def logged_duration(
...
+    Mailchimp, S3, ...) to generate log messages on these events in a standardised manner.
+
+    This implementation will not log durations outside of a Flask request context (e.g. when running scripts).
+
+    Use to wrap a call to a third-party service like so [note the final () call which accepts additional args]:
+    >>> with partial_logged_duration_for_external_request('Notify', 'send email')():
+    >>>     notify_client.send_email('user@email.com')
+    """
+    if not description:
+        description = inspect.stack()[1].function
```
risicle: Ahh very close to the feature I was meaning to add to `logged_duration` that would include the location information of the call but to do so would have to peek "up" a stack frame (or two)...

dmutils/timing.py
```diff
@@ -118,3 +99,58 @@ def logged_duration(
...
+    return has_request_context() and getattr(request, "is_sampled", False)
+
+
+def different_message_for_success_or_error(success_message, error_message):
+    """Can be passed into `logged_duration` as `message=different_message_for_success_or_error(x, y)` in order to
+    generate different log messages depending on whether the block completed successfully or raised an exception."""
+    return lambda _: success_message if sys.exc_info()[0] is None else error_message
+
+
+def partial_logged_duration_for_external_request(service, description=None, success_message=None, error_message=None):
```
risicle: Might not need the `partial_` in the name here - will probably confuse people and to some extent it's an "implementation detail"

samuelhwilliams: I included it because I don't think it's immediately apparent that you need to call `()` the result of this function otherwise. I suspect it would be more confusing with calling out that it's a partial, but I'm open to being convinced otherwise.

risicle: Oh no I see what you mean by it now. Do you think it would be simpler to just merge the two argument lists and avoid the (potentially confusing?) double-function-call by actually calling `logged_duration` yourself? Don't know.

dmutils/timing.py
```diff
@@ -118,3 +99,58 @@ def logged_duration(
...
+    """A public condition that will return True if the duration is above the threshold we have defined as acceptable for
+    calls to external services (e.g. Notify, Mailchimp, S3, etc)."""
+    return log_context['duration_real'] > SLOW_EXTERNAL_CALL_THRESHOLD
+
+
+def request_is_sampled(log_context):
+    """A public condition that returns True if the request has the X-B3-Sampled flag set in its headers. While this is
+    the default condition for logged_duration, exposing it publically allows it to be easily combined with other
+    conditions."""
+    return has_request_context() and getattr(request, "is_sampled", False)
```
risicle: Looks an awful lot like the default condition. Should probably at least make them aliases of each other.

risicle:
Should I dare mention the t-word?

dmutils/timing.py
samuelhwilliams: I considered that, except the default condition is 'private' and could be changed, and that ideally shouldn't change this 'public' interface.

Maybe I'm being overly pedantic.

samuelhwilliams:
Yeah I will work on tests soon...

dmutils/timing.py
risicle: Fair enough

risicle:
Testing logging is a pain, that's all...

dmutils/timing.py
samuelhwilliams: Good idea, I've removed the partial-ness.

CrystalPea:
@katstevens spent some time on trying to test logging recently, she might have some useful insight :)

samuelhwilliams:
@risicle @Wynndow @CrystalPea tests done, I think this is good for a review now 🤞

tests/test_timing.py
```diff
@@ -1,8 +1,8 @@
 from dmutils import timing
+from helpers import MalleableAny, AnySupersetOf, AnyStringMatching
```
risicle: now that these are in `-test-utils` you could just pull that in

tests/helpers.py
```diff
@@ -21,3 +28,95 @@ def name(self):
...
+        log_catcher.uninstall_all()
+
+        matching_records = [True for record in log_catcher.records if message == record.msg]
+        assert len(matching_records) == count, f'{len(matching_records)} log records were seen that matched ' \
+                                               f'`{message}`: expected {count}'
+
+
+def assert_external_service_log_entry(service='\w+', description='.+', successful_call=True, extra_modules=None,
+                                      count=1):
+    """A subclass of AssertLogEntry specialised to inspect for the standardised message that is logged when
```
risicle: Your docstrings appear to still hark back to a time when this was a class.

tests/helpers.py
```diff
@@ -21,3 +28,95 @@ def name(self):
...
+
+    finally:
+        log_catcher.uninstall_all()
+
+        matching_records = [True for record in log_catcher.records if message == record.msg]
+        assert len(matching_records) == count, f'{len(matching_records)} log records were seen that matched ' \
+                                               f'`{message}`: expected {count}'
+
+
+def assert_external_service_log_entry(service='\w+', description='.+', successful_call=True, extra_modules=None,
```
risicle: Mayyyyybe call the `service` arg `service_pattern` and `description` `description_pattern`? a bit clearer over what these are intended to be used for...

tests/test_timing.py
samuelhwilliams: Ah nice, I didn't know those had made their way in now. 👍

tests/helpers.py
```diff
@@ -21,3 +28,95 @@ def name(self):
...
+    assign the context manager to a variable you can access the `records` attribute to see all log records and make
+    further inspections.
+
+    Example:
+        with AssertLogEntry(message=AnyStringMatching('^My unknown string with a random word \w+$', count=2)) as logs:
+            do_something_that_produces_a_log()
+
+        assert any('some other log message' in record.msg for record in logs.records)
+    """
+    log_catcher = logcapture.LogCapture(names=modules, install=False)
```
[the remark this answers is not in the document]
samuelhwilliams: Our logging setup is stdlib Python logging. I'm not sure what your concern is.

It works fine in all my playing around with it. 🤷‍♂️

tests/helpers.py
samuelhwilliams: Yeahhhhh I guess I coullldddd. But then the `with` context blocks will probably end up taking two lines, and I don't know if I like that more than the short names and updated docstring.

dmutils/email/dm_mailchimp.py
```diff
@@ -45,20 +48,22 @@ def wrapper(*args, **kwargs):
...
-            campaign = self._client.campaigns.create(campaign_data)
+            with log_external_request(service='Mailchimp'):
+                campaign = self._client.campaigns.create(campaign_data)
             return campaign['id']
         except RequestException as e:
             self.logger.error(
-                "Mailchimp failed to create campaign for '{0}'".format(
-                    campaign_data.get("settings").get("title")
+                "Mailchimp failed to create campaign for '{campaign_title}'".format(
+                    campaign_title=campaign_data.get("settings").get("title")
```
galund: If you want to chain `get` calls then (unless I've misremembered) you have to make sure the default value is a dictionary, so in this case you need `...get("settings", {}).get(...`

tests/helpers.py
galund: "Unicode-replacement-character unicode-replacement-character" to you too

dmutils/email/dm_mailchimp.py
samuelhwilliams: I mean I didn't change that bit, so it clearly works as-is... But yeah that's better practice. I'll update it if you promise to finish reviewing so this can go in. ;)

dmutils/email/dm_mandrill.py
```diff
@@ -6,6 +6,7 @@
 from mandrill import Mandrill
 from dmutils.email.exceptions import EmailError
 from dmutils.email.helpers import hash_string
+from dmutils.timing import logged_duration_for_external_request as log_external_request
```
galund: If we don't like the name of this thing, then maybe now is the time to change it, cc @risicle

dmutils/timing.py
```diff
@@ -10,27 +11,6 @@
...
-    calls to external services (e.g. Notify, Mailchimp, S3, etc)."""
-    return log_context['duration_real'] > SLOW_EXTERNAL_CALL_THRESHOLD
-
-
-def request_is_sampled(log_context):
-    """A public condition that returns True if the request has the X-B3-Sampled flag set in its headers. While this is
-    the default condition for logged_duration, exposing it publically allows it to be easily combined with other
-    conditions."""
-    return has_request_context() and getattr(request, "is_sampled", False)
-# End public conditions ----------------------
```
galund: I'm not clear why we lost these comments, but it seems to have made the diff much bigger than it needs to be (or something has).

tests/helpers.py
```diff
@@ -21,3 +26,69 @@ def name(self):
...
+def assert_log_entry(modules, message, count=1):
+    """A context manager that can be used to wrap a block of code under test to assert that a specific message is
+    logged. Works well with the AnyStringMatching malleable above to assert a message meeting a given format.
+
+    This is designed to most easily assert a single kind of message being logged one or more times, but if you
+    assign the context manager to a variable you can access the `records` attribute to see all log records and make
+    further inspections.
+
+    Example:
+        with AssertLogEntry(message=AnyStringMatching('^My unknown string with a random word \w+$', count=2)) as logs:
```
galund: Have we changed our name about the name/casing of this thing since the docblock was written, or am I just confused?

tests/helpers.py
galund: How do these regex's get invoked, this is a thing I haven't seen before.

dmutils/email/dm_mailchimp.py
risicle: On this subject I saw something interesting the other day https://sedimental.org/ https://github.com/mahmoud/glom

dmutils/email/dm_mandrill.py
risicle: I called it `logged_duration` so that the line reads `with logged_duration...` i.e. "do the following with its duration being logged...

What other people want to call other things I don't know. I can see that here it's possibly ok to alias it to a shorter-but-less-clear name for brevity, but newcomers to `timing.py` would still see the function with its proper name for clear signposting.

tests/helpers.py
risicle: Oh that's a point - the default value here `service='\w+'` is missing the `r` prefix. The `\w` will be being turned by python into a ... something ... character.

galund: In which case *how does it work at all* :-D

galund:
Rob's comment linking to `glom` https://sedimental.org/ got disappeared somehow by github (grrr) but I just wanted to say oooooh and aaaaah and niiiiiiice.

tests/email/test_dm_mailchimp.py
```diff
@@ -10,298 +11,301 @@
...
+                dm_mailchimp_client, 'subscribe_new_email_to_list', autospec=True) as subscribe_new_email_to_list:
+            subscribe_new_email_to_list.side_effect = [False, True]
 
-        res = dm_mailchimp_client.subscribe_new_email_to_list('list_id', 'example@example.com')
-
-        assert res is False
-        assert logger.error.call_args == mock.call(
-            'Mailchimp failed to add user (foo) to list (list_id)', extra={'error': 'error sending'}
-        )
+            with assert_external_service_log_entry(count=2):
```
katstevens: Do we need to assert anything from log_catcher here? There are a few other examples below (for both expected success and expected failure) where we capture logs with the context manager but don't check the log message content. Or are we not too bothered about that?

tests/helpers.py
risicle: It seems `\w` isn't an escaped sequence in python, but I had to try it to find out.

tests/email/test_dm_mailchimp.py
samuelhwilliams: The `assert_external_service_log_entry` actually performs assertions automatically for the external service log messages. If you need to make _extra_ assertions (e.g. for other/additional log messages), you can store the log catcher in a variable to get access to all of the log records that are caught. I've done this in a few of the tests to allow them to make their old/existing assertions for other log messages.

tests/helpers.py
samuelhwilliams: Ah yeah good spot. Originally the `assert_log_entry` thing was a class. I need to update the docstring.

dmutils/timing.py
samuelhwilliams: I moved this block down in the file. It does make for an ugly diff (sorry), but it feels better grouped at the bottom of the file.

tests/helpers.py
samuelhwilliams: @galund the regexes are applied by the `AnyStringMatching` class that @risicle added to test-utils.

PR #398, opened 2018-05-27
Add a wtform error formatter

samuelhwilliams:
## Summary
We want to format errors with a standard interface so that we can pass
them into templates and always interact with them in the same way,
enhancing the capabilities of our templates to react to errors in a
consistent way. This adds a method `dmutils.forms.format_wtform_errors`
for parsing and formatting errors from Flask-WTForms into the same
format generated by content-loader form errors.

 ## Ticket
https://trello.com/c/RSYEM3FL/448

### DIFF
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from itertools import product
 
 from flask_wtf import FlaskForm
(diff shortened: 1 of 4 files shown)


tests/test_forms.py
```diff
@@ -158,3 +169,62 @@ def test_silently_passes_if_csrf_token_not_present(self):
         data = {'key': 'value', 'key2': 'value2'}
         cleaned_data = remove_csrf_token(data)
         assert cleaned_data == data
+
+
+class TestFormatWTFormErrors:
+    def test_empty_list_if_no_errors(self, app):
+        with app.app_context():
```
katstevens: So annoying that we need this `app_context` for a whole bunch of utils tests 😿 Not something to fix here, anyway.

samuelhwilliams: I can probably make it less bad. But yeah, it's still annoying. :(

samuelhwilliams: Mmm, nope, I can't really make it less bad.

I'll make a tech debt ticket :( Hopefully I'll get some time to pick it up later.

PR #400, opened 2018-06-04
Add class specialisations for using WTForms with the frontend toolkit

lfdebrux:
These are the changes I mentioned in this [PR](https://github.com/alphagov/digitalmarketplace-utils/pull/) and in retro today.

The idea is to make it so that when using WTForms there is an unambiguous and obvious way of linking it up with the frontend toolkit macro. 

To be clear that these are specialisations, the class names all have the 'DM' prefix.

## Example

Say we want a set of radio buttons. In WTForms you would use `RadioField`, but now there is a `DMRadioField` that you can use. You create your form in the same way:

```
class DMRadioForm(FlaskForm):
    radio = DMRadioField(choices=[...])
```

The improvement comes when it is time to use it in the Jinja template. To use a DMRadioField object with the selection-buttons macro, the template code would look like:

```
{%
  with
  name = form.radio.name,
  type = "radio",
  value = form.radio.value,
  error = form.radio.error,
  options = form.radio.options
%}
  {% include "toolkit/forms/selection-buttons.html" %}
{% endwith %}
```

which (hopefully 🤞) is self-explanatory and easy to write.


samuelhwilliams:
Needs a version bump in `dmutils/__init__`. Either a minor, or a major if we want to say that forms must be updated to this standard when this is pulled in. In the latter case, we'll also need to update the changelog.

I'd probably prefer doing it as a major. 🤔  Even though it's not technically breaking backwards compatibility, we should try to get this standard applied to the frontend apps (as long as it can go in fairly easily). Open to being convinced it should just be a minor though.

lfdebrux:
@samuelhwilliams I'm happy for it to be a major version if you are. I'm happy to start changing frontend apps myself, as far as I'm aware WTForms currently isn't used in many apps.

samuelhwilliams:
I think it's used in 3 apps, so it shouldn't be awful to pull in. Hopefully easier once I get the existing validation masthead consolidation stuff in properly (for supplier and buyer...)

lfdebrux:
Tests have been added, which resulted in some fixes, as tests do

dmutils/forms/dm_fields.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .mixins import DMFieldMixin
+
+
+__all__ = ['DMBooleanField', 'DMDecimalField', 'DMEmailField', 'DMHiddenField',
+           'DMIntegerField', 'DMRadioField', 'DMStringField', 'DMStripWhitespaceStringField']
+
+
+class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
+    pass
```
DilwoarH: should we include all of these classes that don't do anything?

dmutils/forms/dm_fields.py
```diff
@@ -0,0 +1,65 @@
...
+            self.options = []
+            for choice in choices:
+                option = {}
+                option['value'] = choice[0]
+                option['label'] = choice[1]
+                if len(choice) > 2 and choice[2]:
+                    option['description'] = choice[2]
+
+            # construct a choices argument suitable for wtforms.fields.RadioField
+            choices = [(option['value'], option['label']) for option in self.options]
```
DilwoarH: might be more efficient to do the following:

```
self.options = []
wtforms_choices = []
for choice in choices:
    option = {}
    option['value'] = choice[0]
    option['label'] = choice[1]
    if len(choice) > 2 and choice[2]:
        option['description'] = choice[2]
    
    wtforms_choices.append( (option['value'], option['label']) )

super().__init__(label, validators, choices=wtforms_choices, **kwargs)
```

dmutils/forms/dm_fields.py
lfdebrux: So I included specialisations for all classes that I saw being used in the in the app codebase. Although the class body is empty, the fact that it inherits from the mixin means that the new class gets all the features of  https://github.com/alphagov/digitalmarketplace-utils/blob/ldeb-wtforms-dmfields/dmutils/forms/dm_fields.py.

I'll add a docstring to the module that explains this.

dmutils/forms/dm_fields.py
lfdebrux: I'd prefer to keep the two separate as I think it's clearer to read. Is that okay with you?

galund:
Would like to have a general discussion about this, but am vaguely thinking I'd be in favour of just having these replacing/augmenting `forms.fields` rather than in their own `dm_fields` thing. Appreciate others have expressed different views so we should definitely discuss before deciding.

dmutils/forms/mixins.py
```diff
@@ -0,0 +1,70 @@
...
+        - a `question` property that contains the label text
+        - a `value` property for the data that is displayed
+        - an `error` property for the field validation error
+
+    Derived classes which include this mixin should have a
+    subclass of `wtforms.Field` in their base classes.
+    '''
+    def __init__(self, label=None, validators=None, hint=None, **kwargs):
+        self.hint = hint
+        self.question = label
```
samuelhwilliams: It feels like this could get out of sync with the WTForm native `label` attribute. I think there are at least a few forms that update the label based on some dynamic context when the form is initialized.

Can we do a proxy on `self.label.text` using the `@property` decorator rather than copy the value across?

lfdebrux: Done

dmutils/forms/mixins.py
```diff
@@ -54,10 +54,13 @@ class DMFieldMixin:
     '''
     def __init__(self, label=None, validators=None, hint=None, **kwargs):
         self.hint = hint
-        self.question = label
 
         super().__init__(label=label, validators=validators, **kwargs)
 
+    @property
+    def question(self):
+        return self.label
```
samuelhwilliams: This should probably be a proxy on `self.label.text` specifically. Using `self.label` will emit the text inside HTML `<label>` tags, which we probably wouldn't want in most cases.

lfdebrux: Ah yes that makes sense

dmutils/forms/fields.py
```diff
@@ -17,7 +17,7 @@
 class StripWhitespaceStringField(StringField):
```
galund: So the suggestion now is that all of the implementation here should be moved into `dm_fields` classes; these classes would inherit from the `dm_fields` ones and would be deprecated.

lfdebrux: Do we want to stick a `DeprecationWarning` on them?

galund: Yeah, reckon a warning emitted by the constructor is the right thing here? And move any tests over from the old thing to the new thing.

dmutils/forms/dm_fields.py
```diff
@@ -0,0 +1,84 @@
...
+        if choices:
+            # We do this the long way rather than using a comprehension because
+            # we want to be able to accept (value, label) tuples as well.
+            self.options = []
+            for choice in choices:
+                option = {}
+                option['value'] = choice[0]
+                option['label'] = choice[1]
+                if len(choice) > 2 and choice[2]:
+                    option['description'] = choice[2]
```
samuelhwilliams: It feels like we probably need to append the constructed `option` dict to `self.options`, unless I'm missing something magical. 🤔  I can't see where it would otherwise get persisted here.

tests/forms/test_fields.py
```diff
@@ -0,0 +1,33 @@
...
+class StripWhitespaceStringFieldForm(wtforms.Form):
+    field = StripWhitespaceStringField()
+
+
+@pytest.mark.parametrize('form, field_class', (
+    (EmailForm(), EmailField),
+    (StripWhitespaceStringFieldForm(), StripWhitespaceStringField),
+))
+def test_form_with_field(form, field_class):
+    assert isinstance(form.field, field_class)
```
samuelhwilliams: I'm not sure I understand what this is testing.

lfdebrux: To be honest I don't remember.

I will take the test out.

dmutils/forms/dm_fields.py
lfdebrux: Fixed

dmutils/forms/fields.py
lfdebrux: Done

dmutils/forms/dm_fields.py
```diff
@@ -0,0 +1,202 @@
...
+from .validators import EmailValidator
+
+
+__all__ = ['DMBooleanField', 'DMDecimalField', 'DMHiddenField', 'DMIntegerField',
+           'DMRadioField', 'DMStringField', 'DMEmailField', 'DMStripWhitespaceStringField',
+           'DMDateField']
+
+
+class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
+    pass
```
galund: From my experience in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/798 we are missing some detail here required to make this work, specifically we need a `options ` I think.

lfdebrux: Got it, will add 👍

lfdebrux:
I've just rebased the changes on master, and also added `options` to DMBooleanField as @galund suggested, and made this PR into a true breaking change by completely removing the old field code.

dmutils/forms/mixins.py
```diff
@@ -0,0 +1,110 @@
...
+    'text'
+    >>> form.text.question
+    'Text field'
+    >>> form.text.hint
+    'Type text here.'
+
+For more examples of these mixins in action, see `dmutils/forms/dm_fields.py`.
+
+For more about mixins in general, see this `online article`_,
+or the very excellent `Python Cookbook`_ by David Beazley.
```
risicle: *insert amazon affiliate link*

risicle:
Would there be a neat way of adding a method to these field instances which will do all of the template-calling for us? So we could just do `{{ someform.somefield.get_dmfrontend_widget() }}`?

Or even just override wtforms standard way of doing this - `__html__()` or something?

dmutils/forms/mixins.py
lfdebrux: Also available at all good bookstores

lfdebrux:
@katstevens 6 apps in total:

    1. admin-frontend
    2. brief-responses-frontend
    3. supplier-frontend
    4. buyer-frontend
    5. briefs-frontend
    6. user-frontend

I've already done the work for buyer-frontend [here](alphagov/digitalmarketplace-buyer-frontend#826)

PR #402, opened 2018-06-04
Add 'supplier_framework' to api_stubs

Wynndow:
It'd be useful to have for our tests. I've created it to drop into this PR: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/886

### DIFF
--- a/dmutils/api_stubs.py
+++ b/dmutils/api_stubs.py
@@ -3,6 +3,15 @@ from datetime import datetime as dt
 from .formats import DATETIME_FORMAT
 
 
+def _derive_framework_family(slug):
+    if slug.startswith('g-cloud'):
+        return 'g-cloud'
+    elif slug.startswith("digital-outcomes-and-specialists"):
+        return 'digital-outcomes-and-specialists'
+    else:
+        return slug
+
+
 def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_limit=False, unit_singular='service',
         unit_plural='services'):
     if not name:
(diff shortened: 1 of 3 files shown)


dmutils/api_stubs.py
```diff
@@ -213,3 +215,87 @@ def supplier(id=1234, contact_id=4321, other_company_registration_number=0, comp
         data['suppliers']['registrationCountry'] = 'country:NZ'
 
     return data
+
+
+def supplier_framework(
+    agreed_variations=True,
+    supplier_id=1234,
+    framework_slug='g-cloud-7',
```
samuelhwilliams: Is there a big issue with using a more recent framework as the defaults (e.g. 10, and prefill from 9)?

tests/test_api_stubs.py
```diff
@@ -247,6 +247,155 @@ def test_further_competition_vs_direct_award(self):
...
+                'frameworkFramework': 'digital-outcomes-and-specialists',
+                'frameworkSlug': 'digital-outcomes-and-specialists-2',
+                'onFramework': False,
+                'prefillDeclarationFromFrameworkSlug': 'digital-outcomes-and-specialists',
+                'supplierId': 4444244,
+                'supplierName': 'My Little Company',
+            }
+        }
+
+    def test_agreed_variatins(self):
```
samuelhwilliams: typo in `variations`

dmutils/api_stubs.py
```diff
@@ -213,3 +215,87 @@ def supplier(id=1234, contact_id=4321, other_company_registration_number=0, comp
         data['suppliers']['registrationCountry'] = 'country:NZ'
 
     return data
+
+
+def supplier_framework(
+    agreed_variations=True,
```
samuelhwilliams: It's probably not the end of the world with this as-is, but it would maybe be nice to have this able to inject actual/different/multiple variations. _Maybe_.

dmutils/api_stubs.py
Wynndow: I was just trying to keep the stubs in sync. default for the framework stub is g7.

dmutils/api_stubs.py
Wynndow: I think I'll leave that as a task for if/when it comes up.

PR #403, opened 2018-06-05
Update API stubs for frameworks and lots

samuelhwilliams:
## Summary
Fix lot capitalisation to only capitalise the first letter, and set
further_competition/direct_award flags based on framework family.

There are lots of places in frontend apps where we want to create known
lots for G-Cloud/Digital Outcomes and Specialists frameworks, so let's
simplify this by creating simple pre-configured lots. We'll also update
the framework stub to populate these lots correctly for our two known
frameworks if no argument is supplied.

### DIFF
--- a/dmutils/api_stubs.py
+++ b/dmutils/api_stubs.py
@@ -15,8 +15,7 @@ def _derive_framework_family(slug):
 def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_limit=False, unit_singular='service',
         unit_plural='services'):
     if not name:
-        name = slug.replace("-", " ")
-        name = name[:1].upper() + name[1:]
+        name = slug.replace("-", " ").capitalize()
 
     return {
         "id": lot_id,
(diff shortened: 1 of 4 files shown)


dmutils/api_stubs.py
```diff
@@ -15,7 +15,8 @@ def _derive_framework_family(slug):
 def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_limit=False, unit_singular='service',
         unit_plural='services'):
     if not name:
-        name = slug.replace("-", " ").title()
+        name = slug.replace("-", " ")
+        name = name[:1].upper() + name[1:]
```
Wynndow: How about `name = slug.capitalize().replace("-", " ")`

dmutils/api_stubs.py
```diff
@@ -28,6 +29,41 @@ def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_li
...
+g_cloud_6_lots = __as_a_service_lots
+g_cloud_7_lots = __as_a_service_lots
+g_cloud_8_lots = __as_a_service_lots
+g_cloud_9_lots = __cloud_lots
+g_cloud_10_lots = __cloud_lots
+
+
+def dos_lots():
+    return [
+        lot(lot_id=5, slug='digital-outcomes', name='Digital outcomes', allows_brief=True),
```
Wynndow: Could add `one_service_limit=True` for outcomes, specialists and user-research-participants?

dmutils/api_stubs.py
samuelhwilliams: That's exactly what I wanted. 🤦‍♂️

dmutils/api_stubs.py
samuelhwilliams: 👍

samuelhwilliams:
Updated & addressed.

dmutils/api_stubs.py
```diff
@@ -28,6 +28,42 @@ def lot(lot_id=1, slug="some-lot", name=None, allows_brief=False, one_service_li
...
+
+def __cloud_lots():
+    return [
+        lot(lot_id=9, slug='cloud-hosting', name='Cloud hosting'),
+        lot(lot_id=10, slug='cloud-software', name='Cloud software'),
+        lot(lot_id=11, slug='cloud-support', name='Cloud support')
+    ]
+
+
+g_cloud_4_lots = __as_a_service_lots
```
Wynndow: Are g4 - g6 likely to get used ever? I know I added something yesterday to stubs which referenced G6, but it feels like there should be a line somewhere (I don't know where it is though).

samuelhwilliams: Hopefully not, although these are here to hopefully mean there are fewer barriers to dropping these in as replacements for existing test fixtures. We can look at removing those when we're consistently using these stubs rather than text/json files.

PR #405, opened 2018-06-06
Centralise Notify domain overrides

samuelhwilliams:
## Summary
All of our apps define some email domains that Notify should not
actually send emails to (e.g. the email domains we use for test accounts
on dev environments). These are currently duplicated across all of the
apps. Our scripts also need access to these override domains, so it
seems sensible to centralise them - since I doubt they need to change
much per app/script.

This does a major version bump to 39.0.0 so that we can enforce the
update as utils is pulled into things.

## Ticket
https://trello.com/c/YIJ25eiP/479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,37 @@
 
 Records breaking changes from major version bumps
 
+## 39.0.0
+
+PR [?]
+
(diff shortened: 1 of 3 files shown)


risicle:
Hmmm... this means that:

1. This isn't really configuration any more
2. If we want to make an alteration to it it's yet another repository to bubble through, possibly requiring a major `-utils` version bump at an inconvenient time...

katstevens:
I think I agree with Rob? Having settings in utils doesn't seem like a good trend to me. Is there somewhere else we can centralise these instead? Environment vars?

risicle:
I can't think of anywhere else we can centralize these that is used in local deployments. Don't talk too lous or some people will take this as a "let's create a new repo...".

samuelhwilliams:
Definitely open to other ideas.

Alternatively, can just splat the domain translations raw into the scripts that need them, but that seems pretty awful.

samuelhwilliams:
Will probably just go with https://github.com/alphagov/digitalmarketplace-scripts/pull/232 for now then

PR #406, opened 2018-06-06
Add a DateField for WTForms

lfdebrux:
When using WTForms there is currently no default way to validate the
date fields provided by the frontend toolkit.

This commit provides a new `DateField` class that can be used with
WTForms to intrepret and validate data from one of these fields
correctly.


dmutils/forms/validators.py
```diff
@@ -10,3 +17,60 @@ class EmailValidator(Regexp):
...
+        except KeyError:
+            raise ValidationError(field.gettext("Invalid field name '%s'." % self.fieldname))
+        if other.data and not field.data > other.data:
+            d = {
+                'other_label': hasattr(other, 'label') and other.label.text or self.fieldname,
+                'other_name': self.fieldname
+            }
+            message = self.message
+            if message is None:
+                message = field.gettext('Field must be greater than %(other_name)s.')
```
risicle: Very fancy - you're probably the first of us to bother using gettext.

dmutils/forms/fields.py
```diff
@@ -28,3 +35,83 @@ def __init__(self, label=None, **kwargs):
             ),
         ))
         super(EmailField, self).__init__(label, **kwargs)
+
+
+class DateField(Field):
+    '''
+    A date field(set) that uses a day, month and year field.
+
+    It behaves like a WTForms.FieldForm, but it can be used
```
risicle: Do you mean `FormField` here instead of `FieldForm`?

lfdebrux: Woops, good spot

dmutils/forms/validators.py
```diff
@@ -10,3 +17,60 @@ class EmailValidator(Regexp):
...
+
+            raise ValidationError(message % d)
+
+
+class Y2K:
+    """
+    Prepare for the 21st Century.
```
risicle: *I* like this joke but it might not make much sense for people who are too young to know what relevance "Y2K" has to for digit years...

On this note, is it worth maybe including this validator in the `DateField` by default?

lfdebrux:
Kids these days...

If we decide to keep the class I will rename it to `FourDigitYear`.

I thought about including the four digit year check as part of the
processing, but the reasons I kept it as a validator were that a) you can't
then set the validation error message in a way that is consistent with
WTForms, and b) by including the validator explicitly it acts as
documentation saying that this date must have a four digit year.

On the other hand, I don't think we'd ever want to accept a two digit year.
Thoughts @galund?

dmutils/forms/validators.py
lfdebrux: Kids these days... If we decide to keep the class I will rename it to `FourDigitYear`.

I thought about including the four digit year check as part of the processing, but the reasons I kept it as a validator were that a) you can't then set the validation error message in a way that is consistent with WTForms, and b) by including the validator explicitly it acts as documentation saying that this date must have a four digit year.

On the other hand, I don't think we'd ever want to accept a two digit year. Thoughts @galund?

tests/test_date_field.py
```diff
@@ -0,0 +1,255 @@
...
+from dmutils.forms.validators import GreaterThan, Y2K
+
+import datetime
+
+
+class DateForm(Form):
+    date = DateField()
+
+
+def data_fromtuple(t, prefix='date'):
```
risicle: Oh also we tend to underscore-prefix names that aren't meant for external use (appreciate this might not work with the fixtures below)

PR #408, opened 2018-06-14
tracing: default parent span header name should be `X-B3-ParentSpanId`

risicle:
Don't know how this ended up like this. See https://github.com/openzipkin/b3-propagation

Gonna have some industrial scale bumping to do...

### DIFF
--- a/tests/test_request_id.py
+++ b/tests/test_request_id.py
@@ -291,7 +291,7 @@ _parent_span_id_related_params = (
         {},
         # extra_req_headers
         (
-            ("X-B3-PARENTSPAN", "colossal-edifice",),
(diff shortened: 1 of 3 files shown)


tests/test_request_id.py
```diff
@@ -291,7 +291,7 @@
         {},
         # extra_req_headers
         (
-            ("X-B3-PARENTSPAN", "colossal-edifice",),
+            ("X-B3-PARENTSPANID", "colossal-edifice",),
```
katstevens: Should this test care about case sensitivity? (Guessing probably not...)

risicle: deliberately not.

PR #410, opened 2018-06-14
tracing: include zipkin sampled & debug headers in log output

risicle:
If I'm going to be going to the effort of bumping utils out to everything, may as well include these.

This includes the values of the "sampled" and "debug" zipkin flags in log output. Please only take issue with the json key names `tracingIsSampled` & `tracingDebugFlag` if you have a significantly better suggestion as this is something we could discuss forever...

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -162,19 +162,19 @@ class JSONFormatter(BaseJSONFormatter):
 
         log_record['logType'] = "application"
 
-        missing_keys = []
+        missing_keys = {}
         for attempt in range(self._max_missing_key_attempts):
             try:
-                log_record['message'] = log_record['message'].format(**log_record)
-
+                log_record['message'] = log_record['message'].format(**log_record, **missing_keys)
+            except KeyError as e:
+                missing_keys[e.args[0]] = f"{{{e.args[0]}: missing key}}"
+            else:
+                # execution should only ever reach this point once - when the .format() succeeds
                 if missing_keys:
-                    logger.warning("Missing keys when formatting log message: {}".format(missing_keys))
+                    logger.warning("Missing keys when formatting log message: {}".format(tuple(missing_keys.keys())))
 
                 break
 
-            except KeyError as e:
-                missing_keys.append(e.args[0])
-                log_record[e.args[0]] = f"{{{e.args[0]}: missing key}}"
         else:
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Some of our logs in Kibana already include these details under the keys `debug_flag` and `is_sampled`. Not sure how they get pulled in, and/or why we need this to log them separately, but we should make sure we aren't duplicating this under different keys.

risicle:
Oh. I don't think that's deliberate. Hmm...

risicle:
I was under the impression that if a field weren't explicitly mentioned in `LOG_FORMAT` or `LOG_FORMAT_EXTRA_JSON_KEYS` it wouldn't get output...

dmutils/logging.py
```diff
@@ -148,6 +150,8 @@ def process_log_record(self, log_record):
             ("span_id", "spanId",),
             ("parent_span_id", "parentSpanId",),
             ("app_name", "application",),
+            ("tracing_is_sampled", "tracingIsSampled",),
+            ("tracing_debug_flag", "tracingDebugFlag",),
```
katstevens: I think there's a test for these field renamings that we can add these too?

risicle: There are tests but are we going to add a test for every single pair we add here?

risicle: Oh I think I see the bit you mean...

katstevens: I'm hoping we're not going to add too many more!

risicle:
Yup. Our logs do indeed spit out the raw unexpected bits of log context and this slipped through due to this module's preference for unit testing rather than "does it actually work?" testing. More tomorrow.

risicle:
I've pushed a new version which does a couple of extra things: adds a high level test of some logging behaviour because it was lacking clarity over what was *supposed* to happen in various cases involving "required" keys and it also failed to pick up the fact that the "missing key" functionality was injecting spurious keys into the final json output as a side effect of its dictionary-stuffing-with-`{foo: missing key}` functionality.

I also fixed that and switched the "missing keys" code's innards around a bit based on some feedback a while ago from @katstevens. See that commit's comment for more detail.

risicle:
Oh and I also switched the output key names to remove the `tracing` prefix as the previous decision seems to have gone that way...

PR #411, opened 2018-06-15
tracing: change design to perform "promotion" of span ids in-app 

risicle:
Trello https://trello.com/c/UOXf4vZZ/308-make-use-of-zipkin-headers-in-preference-to-request-id

Relying on the hosting platform to promote `X-B3-SpanId` header values to `X-B3-ParentSpanId` headers and assign fresh `X-B3-SpanId` values appears to be deprecated behaviour.

### DIFF
--- a/dmutils/request_id.py
+++ b/dmutils/request_id.py
@@ -1,5 +1,5 @@
-import uuid
 from itertools import chain
+from random import SystemRandom
 
 from flask import request, current_app
 
@@ -9,6 +9,8 @@ class RequestIdRequestMixin(object):
         A mixin intended for use against a flask Request class, implementing extraction (and partly generation) of
         headers approximately according to the "zipkin" scheme https://github.com/openzipkin/b3-propagation
     """
+    # a single class-wide random instance should be good enough for now
+    _spanid_random = _traceid_random = SystemRandom()
 
     @property
     def request_id(self):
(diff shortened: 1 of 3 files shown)


risicle:
(this would be something it would be nice to bring out as part of the bumpathon I seem to need to do anyway...)

dmutils/request_id.py
```diff
@@ -9,6 +9,8 @@ class RequestIdRequestMixin(object):
         A mixin intended for use against a flask Request class, implementing extraction (and partly generation) of
         headers approximately according to the "zipkin" scheme https://github.com/openzipkin/b3-propagation
     """
+    # a single class-wide random instance should be good enough for now
```
galund: I don't really follow the "for now" aspect here. If there's no conceivable reason to change this, then don't say "for now". If there is a reason this maybe should change one day, then call out the specific?

dmutils/request_id.py
```diff
@@ -70,19 +74,30 @@ def _get_first_header(self, header_names):
...
         return dict(chain(
             (
                 (header_name, self.trace_id)
                 for header_name in current_app.config['DM_TRACE_ID_HEADERS']
             ) if self.trace_id else (),
             (
-                (header_name, self.span_id)
+                (header_name, new_span_id)
                 for header_name in current_app.config['DM_SPAN_ID_HEADERS']
+            ) if self.trace_id else (),
```
galund: Why do we only add a span id if there's a trace id?

dmutils/request_id.py
risicle: God knows, we might find using the `SystemRandom` to be overkill for this purpose and it sucks up our instance's entropy. Don't know. But it's a comment.

dmutils/request_id.py
risicle: Because it doesn't really make sense without a trace id - trace id's presence is kinda a signifier that zipkin is in use at all here.

PR #419, opened 2018-07-04
Move env helpers from scripts to utils

katstevens:
As discussed in this issue: https://github.com/alphagov/digitalmarketplace-scripts/issues/257

The env helpers are useful for constructing dynamic fully-qualified urls within the apps, for example in email templates. I added in some tests, which revealed the static assets helper doesn't work for local envs (which makes sense, as there's no router app to handle them). For now I tweaked the function to raise a `NotImplementedError` if anyone tries to use it for local assets.

The `get_sent_emails` helper function feels like it belongs with the other Mandrill utils.

Original files can be found here:
- https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/helpers/env_helpers.py
- https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/helpers/email_helpers.py

I have left off the csv functions for now, as there was some funny stuff happening in those functions regarding unicode which I want to investigate first.

I'll do a separate PR to pull in this version of the utils to the scripts, and deprecate the old functions there.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '40.3.0'
+__version__ = '40.4.0'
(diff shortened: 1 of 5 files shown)


katstevens:
That's a good idea, I'll give that a whirl now.

katstevens:
All seems fine once I added in the new FlaskFeatureFlags repo 👍

PR #420, opened 2018-07-06
request logging: include pid, "thread ident" and flask "endpoint"

risicle:
This will aid in performance analysis. Specifically adding it only to the main "request" message as it should remain the same for the whole request - the pid for other messages logged on this request should be discoverable through following a span id. Doing this reduces log volume as opposed to adding the information to each & every log message produced.

I considered hashing the pid and thread ident for extra paranoia but thought that was a bit much because that would also require me to source a seed from some configuration variable...


risicle:
Added a commit to this which also includes the flask request's `endpoint` attribute which should prove handy.

risicle:
Note that `process` and `thread` information is available on `LogRecord` objects by default, but I don't see a straightforward way of selectively including them only in specific log messages - they are designed to be included when the formatter is being configured. This is why I'm manually grabbing them and putting them in as `extra` here. Even if this doesn't seem completely bonkers to people I guess it's probably worth documenting this thought in the comments...

benvand:
👍 LGTM

risicle:
Had a last minute mind-change on the parameter names, using `_process` and `_thread` - as close as possible to the "standard" `process` & `thread`.

PR #421, opened 2018-07-09
add `timed_render_template`

risicle:
This is a simple `logged_duration`-wrapped version of flask's `render_template` and `render_template_string` which will output a `DEBUG` log message either when the request has been sent with zipkin "sample" mode or the request takes more than `SLOW_RENDER_THRESHOLD` seconds. The idea being in frontends we can freely use this in place of flask's regular `render_template` and get some insight as to the behaviour of the view "for free". Will be quite useful for performance profiling and with any luck might give us an extra clue about weird stuff.

What do people think about 250ms? Was taken from sam's similar "external call threshold"...

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '40.2.0'
+__version__ = '40.3.0'
--- a/dmutils/flask.py
+++ b/dmutils/flask.py
@@ -1,0 +1,34 @@
+from functools import partial
+
+from flask import render_template, render_template_string
+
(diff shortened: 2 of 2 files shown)


katstevens:
Functions look good - a short docstring would be useful (the PR description would be fine). 

I'm stuck in the usual naming-things quandary though - I'm in two minds whether `flask.py` is confusing alongside `flask_init.py`, or whether this should go in `timing.py` (are any non-timing flask helpers likely to be added here?), or whether this is fine as-is.

risicle:
Yeah I don't know about naming - I didn't put it in `timing.py` as I thought it should be seen as a flask-enhancement we're doing, leaving room to add any further enhancements that *aren't* timing related.

risicle:
Added docstring.

PR #422, opened 2018-07-11
request logging: add "early" log messages and timing information

risicle:
This does two things to flask app request logging:

 - If the zipkin `is_sampled` property is set on the request, will cause a log message to be emitted when flask *receives* a request (`before_request`). This could be useful in cases where the app for some reason doesn't manage to log its request completing.
 - Adds timing information to the existing `after_request` log events. This should allow us to deduce how much time is spent queueing for a python thread/process.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '40.4.0'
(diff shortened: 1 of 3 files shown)


risicle:
:steam_locomotive: :train: :train:

risicle:
Just wait till we get python 3.7 and I add `thread_time` too https://docs.python.org/3/library/time.html#time.thread_time

PR #424, opened 2018-07-13
timed_render_template: message fix

risicle:
Change log message to remain correct if triggered by zipkin is_sampled.


risicle:
I considered that, but didn't want to make the log message function basically need to duplicate the logic of the condition callable...

PR #426, opened 2018-07-18
request logging: change "process" and "thread" log parameter names 

risicle:
Our log pipeline seems to strip leading-underscore parameters

Also quietly switch `timed_render_template`'s `SLOW_RENDER_THRESHOLD` back to 0.25s - turns out this isn't quite so common after instances have warmed up their template bytecode caches...


risicle:
That's a question.

I'm guessing it's in the lambda function.

PR #431, opened 2018-07-23
EmailValidator for wtforms: replace our own regex based routine with one stolen from Notify

risicle:
https://trello.com/c/7m7tPP83/379-email-validation-should-we-use-notifys
https://trello.com/c/LD2eVPTY/368-supplier-user-can-be-created-but-no-email-sent

This is a fairly straightforward port. You'll see the tests needed minimal changes, mostly moving addresses over from what was previously considered "valid" to what's now considered "invalid" and making up a few replacement valid ones that are a bit tamer.

~~Thinking again, I'm actually probably going to make this~~ a major bump as technically the validator's constructor call signature has changed, not accepting a regex override as the previous implementation, inheriting from the regex validator, did.

### DIFF
--- a/dmutils/forms/validators.py
+++ b/dmutils/forms/validators.py
@@ -8,15 +8,66 @@ FourDigitYear -- validate that a four digit year value is provided
 
 import re
 
-from wtforms.validators import Regexp, ValidationError
+from wtforms.validators import ValidationError
 
 
-class EmailValidator(Regexp):
-    _email_re = re.compile(r"^[^@\s]+@[^@\.\s]+(\.[^@\.\s]+)+$")
+class EmailValidator:
+    # Largely copied from https://github.com/alphagov/notifications-utils/blob/\
+    #   67889886ec1476136d12e7f32787a7dbd0574cc2/notifications_utils/recipients.py
+    #
+    # regexes for use in validate_email_address.
+    # invalid local chars - whitespace, quotes and apostrophes, semicolons and colons, GBP sign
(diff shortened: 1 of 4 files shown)


dmutils/forms/validators.py
```diff
@@ -8,15 +8,66 @@
...
+    # Largely copied from https://github.com/alphagov/notifications-utils/blob/\
+    #   67889886ec1476136d12e7f32787a7dbd0574cc2/notifications_utils/recipients.py
+    #
+    # regexes for use in validate_email_address.
+    # invalid local chars - whitespace, quotes and apostrophes, semicolons and colons, GBP sign
+    # Note: Normal apostrophe eg `Firstname-o'surname@domain.com` is allowed.
+    _INVALID_LOCAL_CHARS = r"\s\",;:@£“”‘’"
+    _email_regex = re.compile(r'^[^{}]+@([^.@][^@]+)$'.format(_INVALID_LOCAL_CHARS))
+    _hostname_part = re.compile(r'^(xn-|[a-z0-9]+)(-[a-z0-9]+)*$', re.IGNORECASE)
+    _tld_part = re.compile(r'^([a-z]{2,63}|xn--([a-z0-9]+-)*[a-z0-9]+)$', re.IGNORECASE)
```
Wynndow: what's with the `xn`s here and for the hostname?

risicle: Ah https://en.wikipedia.org/wiki/Internationalized_domain_name

PR #432, opened 2018-07-26
logging: add AppStackLocationFilter, include in default flask configuration

risicle:
https://trello.com/c/U24anuJR/58-improving-profiling-and-live-debuggability

This annotates a log record to include extra information on the first stack frame that's *within* the app, as it's often more illuminating than the location deep within a utility or library function where the log message is actually output.

As configured here initially, only include this information for `WARNING` or `ERROR` messages or messages originating from a "sampled" request.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '41.1.0'
+__version__ = '41.2.0'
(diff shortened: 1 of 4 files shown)


lfdebrux:
Awesome feature? Is there going to be documentation for this? Would be good to make a note of what it is and what the log key is somewhere in the manual.

risicle:
There will be once it's in and I'm sure it does exactly what I want.

lfdebrux:
Looking over the code it makes sense to me

risicle:
Got `is_sampled` test coverage now too.

PR #434, opened 2018-07-31
AppStackLocationFilter: cope with relative co_filename paths, make _findCaller() more robust

risicle:
This is me being dumb and yet more proof that relative imports should be obliterated.

I should also expand the tests to cover the relative case, but want to get this out quick-ish.


risicle:
(though urgency is now over: not having this simply blocks our pipeline temporarily)

PR #435, opened 2018-08-06
Error handling in utils.flask_init

katstevens:

Copies `render_error_page` and `csrf_handler` from Brief Responses app error handlers. Now with added tests!

This should mean we only need to register handlers for 'unusual' errors in the apps themselves (eg `QuestionNotFoundError`, which can still call the new `dmutils.render_error_page` function). So Brief Responses FE's `app.main.errors` would look like this:

```
from app.main import main
from dmapiclient import APIError
from dmcontent.content_loader import QuestionNotFoundError
from dmutils.errors import render_error_page

@main.app_errorhandler(APIError)
def api_error_handler(e):
    return render_error_page(e)


@main.app_errorhandler(QuestionNotFoundError)
def content_loader_error_handler(e):
    return render_error_page(400)
```

Technically this is not a breaking change, as any 'duplicate' error handlers registered by the app in `main.errors` will overwrite those registered earlier by `flask_init.init_app`. Each app will continue using its own error templates for the moment (these will eventually be moved to the FE toolkit).

Requires a small tweak to conftest to allow access to the session during tests.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
-__version__ = '42.0.1'
+__version__ = '42.1.0'
--- a/dmutils/errors.py
+++ b/dmutils/errors.py
@@ -1,0 +1,56 @@
+from flask import redirect, render_template, url_for, flash, session, request, current_app
+from flask_wtf.csrf import CSRFError
+
+
+def csrf_handler(e):
+    """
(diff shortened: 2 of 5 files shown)


lfdebrux:
Looks good. Were you able to run functional tests with this and a modified app?

katstevens:
Good thinking! I'll give it a whirl now.

katstevens:
@lfdebrux yep, the functional tests pass for Brief Responses.

PR #437, opened 2018-08-08
Add supplier_questions to the external routes

lfdebrux:
alphagov/digitalmarketplace-brief-responses-frontend requires this route when sending emails to buyers and suppliers about clarification questions on a DOS opportunity.


lfdebrux:
Woops!

PR #438, opened 2018-08-09
Error handling bug fixes

katstevens:

Fixes a bug introduced in https://github.com/alphagov/digitalmarketplace-utils/pull/435.

Errors returned by the API Client have the attribute `status_code` instead of `code`. 

Also:
- maps 400 errors to a 400.html template
- redirects 401 and 403 errors to the login page

### DIFF
--- a/dmutils/errors.py
+++ b/dmutils/errors.py
@@ -1,5 +1,6 @@
 from flask import redirect, render_template, url_for, flash, session, request, current_app
 from flask_wtf.csrf import CSRFError
+from jinja2.exceptions import TemplateNotFound
 
 
 def csrf_handler(e):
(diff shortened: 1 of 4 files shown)


dmutils/errors.py
```diff
@@ -50,6 +50,10 @@ def render_error_page(e=None, status_code=None):
         503: "errors/500.html",
     }
 
+    # Handle exceptions with .status_code, not .code (e.g. dmapiclient.HTTPError)
+    if hasattr(e, 'status_code'):
```
lfdebrux: I think this line might raise an exception if you supply render_error_template with a status_code instead of an exception; I couldn't see any unit tests where that path is tested though

katstevens: Erk, good spot. Can you tell I did this in a rush 😹

lfdebrux: After all that this line would have worked fine with `None` 😂

katstevens: I like the new way better! 😎

dmutils/flask_init.py
```diff
@@ -69,6 +69,8 @@ def inject_global_template_variables():
 
     # Register error handlers for CSRF errors and common error status codes
     application.register_error_handler(400, errors.csrf_handler)
+    application.register_error_handler(401, errors.redirect_to_login)
+    application.register_error_handler(403, errors.redirect_to_login)
```
risicle: *scratch head* wonder what ever throws these...

PR #441, opened 2018-08-13
Optional error message in templates

katstevens:

In order to pull in the error handling utils/templates to the Buyer FE, we need to:
- handle custom error messages in the templates
- register a handler for 410 'Gone' errors (there is already a template in the FE toolkit for this)


katstevens:
The existing 503 error handler on the Buyer FE passes through the `e.response` as the error message:
https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/main/errors.py#L36

There's probably a decision to be made about what level of visibility we want for error messages that could potentially be displayed to the user, but it's good to have the option there if we need it.

PR #442, opened 2018-08-13
Exclude external routes from list_routes output

lfdebrux:
This output makes more sense to me, feel free to reject if you disagree.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -4,4 +4,4 @@ from .flask_init import init_app, init_manager
 import flask_featureflags  # noqa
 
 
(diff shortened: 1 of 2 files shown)


lfdebrux:
@risicle I'm not sure what you mean? I assumed this was just a command line tool useful for documentation purposes (which is what I was using it for)

katstevens:
🤔 I think it's useful to know which external routes are available for the app? Can we list them separately?

katstevens:
Cool 👍 I'd forgotten that we even had this method available!

risicle:
Sure yeah, I've just never seen this used yet.

PR #444, opened 2018-08-21
Refactor dmwidgets and add TextArea widget

lfdebrux:
This PR has two main things:

## New widget
I've added a TextArea widget (needed for some apps including briefs-frontend)

## Refactored widgets
I've tried to make it easier to add new widgets based on DMJinjaWidgetBase by using a more declarative style; for instance, look at [DMSelectionButtonBase](https://github.com/alphagov/digitalmarketplace-utils/pull/444/files?utf8=%E2%9C%93&diff=unified#diff-e33286de9492b8dbcf36134ad0a6117cL34)

---

This PR also has some more attributes for forms and more tests for widgets thrown in for free

### DIFF
--- a/dmutils/forms/mixins.py
+++ b/dmutils/forms/mixins.py
@@ -56,9 +56,10 @@ class DMFieldMixin:
     Derived classes which include this mixin should have a
     subclass of `wtforms.Field` in their base classes.
     '''
-    def __init__(self, label=None, validators=None, hint=None, **kwargs):
+    def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
         super().__init__(label=label, validators=validators, **kwargs)
         self.hint = hint or getattr(self.__class__, 'hint', None)
+        self.question_advice = question_advice or getattr(self.__class__, 'question_advice', None)
         self.type = getattr(self.__class__, 'type', self.type)
 
     @property
--- a/dmutils/forms/mixins.py
+++ b/dmutils/forms/mixins.py
@@ -58,8 +58,14 @@ class DMFieldMixin:
     '''
     def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
         super().__init__(label=label, validators=validators, **kwargs)
-        self.hint = hint or getattr(self.__class__, 'hint', None)
-        self.question_advice = question_advice or getattr(self.__class__, 'question_advice', None)
+        if hint:
+            self.hint = hint
+        if question_advice:
+            self.question_advice = question_advice
+
+        # wtforms.Field overwrites self.type on init
+        # if we want to specify it on a subclass
+        # this line will bring it back
         self.type = getattr(self.__class__, 'type', self.type)
 
     @property
--- a/dmutils/forms/mixins.py
+++ b/dmutils/forms/mixins.py
@@ -66,6 +66,10 @@ class DMFieldMixin:
     def question(self):
         return self.label.text
 
+    @question.setter
+    def question(self, value):
+        self.label.text = value
+
     @property
     def value(self):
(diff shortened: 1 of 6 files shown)


dmutils/forms/widgets.py
```diff
@@ -1,68 +1,93 @@
...
-        self.template_args = ["error", "name", "hint", "question", "value"] + self.template_args
-        if hide_question:
-            self.template_args.remove("question")
+    def __init__(self, hide_question=False, **kwargs):
+        self.__template__ = None
+
+        self.__context__ = {}
+        self.__context__.update({k: getattr(self, k) for k in dir(self) if not k.startswith("_")})
+        # kwargs can be used to add constants to the template context
+        self.__context__.update(kwargs)
```
risicle: Hmmm... "data classes"... :thinking:

dmutils/forms/mixins.py
```diff
@@ -56,15 +56,20 @@ class DMFieldMixin:
     Derived classes which include this mixin should have a
     subclass of `wtforms.Field` in their base classes.
     '''
-    def __init__(self, label=None, validators=None, hint=None, **kwargs):
+    def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
         super().__init__(label=label, validators=validators, **kwargs)
         self.hint = hint or getattr(self.__class__, 'hint', None)
+        self.question_advice = question_advice or getattr(self.__class__, 'question_advice', None)
         self.type = getattr(self.__class__, 'type', self.type)
```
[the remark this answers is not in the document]
lfdebrux: _sigh_ see this is why we don't code without tests

dmutils/forms/widgets.py
lfdebrux: 3.7 only sadly

dmutils/forms/mixins.py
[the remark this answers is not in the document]
risicle: Ah ok so I've got to go and find the tests that cover this to get an idea why it's designed this way...

dmutils/forms/widgets.py
```diff
@@ -1,68 +1,93 @@
...
+
+        return self._render(context)
 
+    def _render(self, *args, **kwargs):
         # cache the template
         # this cannot be done in __init__ as the flask app may not exist
-        if not self.template:
-            self.template = current_app.jinja_env.get_template(self.template_file)
+        if not self.__template__:
+            self.__template__ = current_app.jinja_env.get_template(self.__template_file__)
```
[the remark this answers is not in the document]
lfdebrux: I don't think so

dmutils/forms/mixins.py
```diff
@@ -56,15 +56,26 @@ class DMFieldMixin:
...
         super().__init__(label=label, validators=validators, **kwargs)
-        self.hint = hint or getattr(self.__class__, 'hint', None)
+        if hint:
+            self.hint = hint
+        if question_advice:
+            self.question_advice = question_advice
+
+        # wtforms.Field overwrites self.type on init
+        # if we want to specify it on a subclass
+        # this line will bring it back
```
risicle: Ah iiiiinteresting. :+1:

(is the attribute name `type` already determined for us or do we get a choice? calling things `type` is a pet peeve of mine, and not just because it leads to people inadvertently redefining `type()`...)

lfdebrux: Yeah, we want there to be an attribute called `type` because some of the form templates expect a variable with that name.

Normally WTForms saves the class name to `type`, which is a pain if you want it to be something else.

risicle: :grumble: ah well.

PR #445, opened 2018-08-21
Add clarificationQuestionsPublishedBy to test brief

lfdebrux:
Some tests expect this key to be present

### DIFF
--- a/dmutils/api_stubs.py
+++ b/dmutils/api_stubs.py
@@ -210,13 +210,14 @@ def brief(status="draft",
         }
     }
 
-    if status in ("live", "closed"):
+    if status is not "draft":
         brief["briefs"]["publishedAt"] = "2016-03-29T10:11:14.000000Z"
         brief["briefs"]["applicationsClosedAt"] = "2016-04-07T00:00:00.000000Z"
         brief["briefs"]["clarificationQuestionsClosedAt"] = "2016-04-02T00:00:00.000000Z"
         brief["briefs"]["clarificationQuestionsAreClosed"] = clarification_questions_closed
         brief["briefs"]["clarificationQuestionsPublishedBy"] = "2016-04-02T00:00:00.000000Z"
-    elif status is "withdrawn":
+
+    if status is "withdrawn":
         brief["briefs"]["withdrawnAt"] = "2016-05-07T00:00:00.000000Z"
     elif status is "unsuccessful":
         brief["briefs"]["unsuccessfulAt"] = "2016-05-07T00:00:00.000000Z"
--- a/dmutils/api_stubs.py
+++ b/dmutils/api_stubs.py
@@ -209,12 +209,19 @@ def brief(status="draft",
             "clarificationQuestions": clarification_questions or [],
         }
     }
+
     if status in ("live", "closed"):
-        brief['briefs']['publishedAt'] = "2016-03-29T10:11:14.000000Z"
-        brief['briefs']['applicationsClosedAt'] = "2016-04-07T00:00:00.000000Z"
-        brief['briefs']['clarificationQuestionsClosedAt'] = "2016-04-02T00:00:00.000000Z"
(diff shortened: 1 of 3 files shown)


dmutils/api_stubs.py
```diff
@@ -214,6 +214,7 @@ def brief(status="draft",
         brief['briefs']['applicationsClosedAt'] = "2016-04-07T00:00:00.000000Z"
         brief['briefs']['clarificationQuestionsClosedAt'] = "2016-04-02T00:00:00.000000Z"
         brief['briefs']['clarificationQuestionsAreClosed'] = clarification_questions_closed
+        brief['briefs']['clarificationQuestionsPublishedBy'] = "2016-04-02T00:00:00.000000Z"
```
katstevens: We should probably have all these fields for statuses `awarded`, `cancelled` and `unsuccessful` as well. This code probably dates back to before those statuses existed 🙀

lfdebrux: Nightmare. Is there somewhere I can find an up to date schema by any chance? 😜

lfdebrux:
So the merge is blocked because the tests are failing; to be honest though I think these tests need a complete rewrite; adding a single field to a dictionary shouldn't cause an entire test suite to fail.

Does anyone have any objections to me doing this in this PR?

katstevens:
Go ahead! The api stubs definitely need some love.

risicle:
There's no reason that "adding a single dictionary" shouldn't cause many tests to fail. It just means that the tests are actually asserting that all the desired features are working as expected in every tested case rather than just the one or two cases the author decided to check that for. I often find bugs in my code in cases where features are used in combinations I hadn't considered.

lfdebrux:
I may have overdone it with the tests; the diff is so long GitHub won't show it...

dmutils/api_stubs.py
katstevens: Ha! I haven't done schemas for the serializations yet, only the actual DB fields. These dates are all properties inferred from a combo of the `published_at` date and some framework-specific fields in the `data`  JSON field: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models/main.py#L1655

lfdebrux: Added

dmutils/api_stubs.py
```diff
@@ -209,11 +209,19 @@ def brief(status="draft",
...
     if status in ("live", "closed"):
-        brief['briefs']['publishedAt'] = "2016-03-29T10:11:14.000000Z"
-        brief['briefs']['applicationsClosedAt'] = "2016-04-07T00:00:00.000000Z"
-        brief['briefs']['clarificationQuestionsClosedAt'] = "2016-04-02T00:00:00.000000Z"
-        brief['briefs']['clarificationQuestionsAreClosed'] = clarification_questions_closed
+        brief["briefs"]["publishedAt"] = "2016-03-29T10:11:14.000000Z"
+        brief["briefs"]["applicationsClosedAt"] = "2016-04-07T00:00:00.000000Z"
+        brief["briefs"]["clarificationQuestionsClosedAt"] = "2016-04-02T00:00:00.000000Z"
+        brief["briefs"]["clarificationQuestionsAreClosed"] = clarification_questions_closed
+        brief["briefs"]["clarificationQuestionsPublishedBy"] = "2016-04-02T00:00:00.000000Z"
```
katstevens: Sorry, I wasn't clear about what each status needs (it's certainly not clear from the API!) Each of `live`, `closed`, `awarded`, `unsuccessful`, `withdrawn` and `cancelled` will have these 5 fields in the serialization, because they will all have the `published_at` datestamp set. It's only `draft` that doesn't have that published date.

The `withdrawnAt`/`unsuccessfulAt`/`cancelledAt` dates below are fine. `awarded` briefs have a `awardedBriefResponseId` which is probably not a good idea to add in to an api stub, the test using the stub can deal with that itself.

lfdebrux: How's it look now?

PR #447, opened 2018-08-24
Remove flask-featureflags integration, bump flask to v0.12.4

risicle:
https://trello.com/c/ZH9trgjH/419-snyk-alert-for-flask-010x-vulnerability

This broadly does three things:

 - Removes any integration with flask-featureflags and drops it as a dependency. The changelog entry should reflect the various things that are no longer supported as a result of this.
 - Bumps the Flask dependency to v0.12.4 to, among other things, fix a minor security vulnerability with 0.10.x
- Drops the `CSRFError` handling workaround required for flask 0.10.x's deficiency, instead registering the handler in a more obvious way.

Making an upgrade to this version will require bumping the app's Flask to 0.12.4, which itself can require some changes. Luckily I've already done this for all (?) the apps and PRs should be following for them. I'll probably only merge this once people are happy with them too.

### DIFF
--- a/dmutils/errors.py
+++ b/dmutils/errors.py
@@ -1,23 +1,15 @@
 from flask import redirect, render_template, url_for, flash, session, request, current_app
-from flask_wtf.csrf import CSRFError
 from jinja2.exceptions import TemplateNotFound
 
 
-def csrf_handler(e):
+def csrf_handler(csrf_error):
     """
-    Workaround for a bug in Flask 0.10.1.
-    CSRFErrors are caught under 400 BadRequest exceptions, so this heavy-handed solution
-     catches all 400s, and immediately discards non-CSRFError instances.
-
-    :param e: CSRF exception instance
+    :param csrf_error: CSRF exception instance
     :param session: Flask session instance
     :param request: Flask request instance
     :param logger: app logger instance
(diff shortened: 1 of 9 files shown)


dmutils/errors.py
```diff
@@ -5,16 +5,17 @@
...
     :param session: Flask session instance
     :param request: Flask request instance
     :param logger: app logger instance
     :return: redirect to login with flashed error message
     """
+
+    # This is the remnant of a workaround for a bug in Flask 0.10.1. In that case, CSRFErrors
+    # are caught under 400 BadRequest exceptions, so this handler can be registered to catch
+    # all 400s, but will immediately discards non-CSRFError instances. Can be removed once
+    # we're absolutely certain nothing is still using Flask 0.10.1.
```
katstevens: I think we can remove it now. If the app has pulled in this version of utils it will have Flask 0.12, and `flask_init` will only supply this function with recognisable `CSRFError` instances (as opposed to, say, `BadRequest`).

katstevens: For clarity we could rename the `e` arg to `csrf_error`?

risicle: :+1:

PR #450, opened 2018-09-04
Expose helper functions for validating email addresses and getting multiple email addresses from a string

lfdebrux:
A lot of our scripts send out emails to users based on addresses that they supply. These addresses have historically been collected from a form and stored in the database without any validation of the input. This means that we now have some failures where the user has tried to supply multiple addresses in one string (see [this ticket](https://trello.com/c/DHDgBuns)).

This PR tries to make it so that in future we can change our scripts so that a) they can handle strings that contain multiple email addresses; and b) they can try to validate the email addresses before they call Notify.

It does this by exposing the email validation algorithm that was recently added by @risicle as a helper function, and also by adding a new helper function that splits a string based on a set of characters to generate a list of possible email addresses.

### DIFF
--- a/dmutils/email/helpers.py
+++ b/dmutils/email/helpers.py
@@ -5,6 +5,10 @@ import hashlib
 import re
 
 
+# List of characters that could be used to separate email addresses,
+# in decreasing precedence order.
+EMAIL_ADDRESS_SEPARATORS = [";", ",", "/"]
+
 # Largely copied from https://github.com/alphagov/notifications-utils/blob/\
 #   67889886ec1476136d12e7f32787a7dbd0574cc2/notifications_utils/recipients.py
 #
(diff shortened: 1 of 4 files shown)


dmutils/email/helpers.py
```diff
@@ -2,9 +2,105 @@
...
+    ['bob@invalid', 'bob.blob@job.com']
+    >>> email_addresses("bob@blob;bob.blob@job.com;bob@blob..invalid")
+    ['bob@blob', 'bob.blob@job.com', 'bob@blob..invalid']
+    """
+
+    addresses = [string]
+    for sep in separators:
+        if sep in string:
+            addresses = string.split(sep)
+            break  # earlier separators take precedence
```
Wynndow: This is probably pretty unlikely, but what if the string uses different separators? `'bob@blob', 'bob.blob@job.com'; 'bob.bob@blob.com'` for example.
This is probably too much of an edge case to worry about but thought it worth thinking about.

tests/forms/test_validators.py
```diff
@@ -0,0 +1,54 @@
+
+import pytest
+import mock
+
+from wtforms.validators import ValidationError
+
+import dmutils.forms.validators
+
+
+@pytest.fixture
```
Wynndow: Should these live in `conftest.py`? I'm not sure. We seem to create some fixtures in test files and some in `conftest.py`

dmutils/email/helpers.py
lfdebrux: I thought about it but decided it was probably too much of an edge case to be worth the additional complexity.

To be fair it should be a fairly straightforward change, basically it would be

```
    for sep in separators:
        if sep in string:
            string = string.split(sep)
```

but I worried it might have weird and hard to debug side-effects.

I guess if something like this appears in the wild we can deal with it then.

tests/forms/test_validators.py
risicle: I think we put them in `conftest.py` when they would appear to be more generally useful, but keep them in-test-module when they're rather specific.

lfdebrux: My understanding from the pytest documentation is that fixtures that are shared across modules go in `conftest.py`.

I guess there would be a benefit to having a global email address fixture, currently this specific fixture is used only for my tests, so I don't think it needs to be in `conftest.py` until that changes.

PR #454, opened 2018-09-06
Make the DMNotifyClient class a bit more ergonomic

lfdebrux:
Since we are customising the NotifyClient for our own purposes, we might 
as well make it fit our systems a bit better.

This commit makes two small changes to make using DMNotifyClient a bit 
less annoying; it automatically gets the API key from the current app 
config, and it looks up the template uuid for you if you want it to.

You don't have to use these features, but they might make your life 
better ;)

### DIFF
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -17,21 +17,35 @@ class DMNotifyClient:
 
     def __init__(
             self,
-            govuk_notify_api_key,
+            govuk_notify_api_key=None,
             govuk_notify_base_url='https://api.notifications.service.gov.uk',
-            logger=None,
             redirect_domains_to_address=None,
+            *,
+            logger=None,
+            templates=None,
     ):
         """
-            :param govuk_notify_api_key:
+            :param govuk_notify_api_key: defaults to current_app.config["DM_NOTIFY_API_KEY"]
             :param govuk_notify_base_url:
-            :param logger: logger to log progress to, taken from current_app if Falsey
             :param redirect_domains_to_address: dictionary mapping email domain to redirected email address - emails
                 sent to a email with a domain in this mapping will instead be sent to the corresponding value set here.
                 if `redirect_domains_to_address` is `None` will fall back to looking for a
                 `DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS` setting in the current flask app's config (if available).
+
+        The following arguments are keyword-only and should only be used if you want to operate outside
+        of a Flask app context.
+            :param logger: logger to log progress to, taken from current_app if Falsey
+            :param templates: a dictionary of template names to template uuids, so that you can use
+                descriptive names when specifying a template. This defaults to current_app.config["NOTIFY_TEMPLATES"].
         """
+        if govuk_notify_api_key is None:
+            govuk_notify_api_key = current_app.config["DM_NOTIFY_API_KEY"]
+
+        if templates is None:
+            self.templates = current_app.config.get("NOTIFY_TEMPLATES", {}) if current_app else {}
+
         self.logger = logger or current_app.logger
+
         self.client = self._client_class(govuk_notify_api_key, govuk_notify_base_url)
         self._redirect_domains_to_address = (
             current_app.config.get("DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS")
@@ -66,11 +80,11 @@ class DMNotifyClient:
         return reference in self.get_delivered_references()
 
     @staticmethod
-    def get_reference(email_address, template_id, personalisation=None):
+    def get_reference(to_email_address, template_id, personalisation=None):
         """
         Method to return the standard reference given the variables the email is sent with.
 
-        :param email_address: Emails recipient
+        :param to_email_address: Emails recipient
         :param template_id: Emails template ID on Notify
         :param personalisation: Template parameters
         :return: Hashed string 'reference' to be passed to client.send_email_notification or self.send_email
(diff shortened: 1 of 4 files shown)


dmutils/email/dm_notify.py
```diff
@@ -82,32 +89,38 @@ def get_reference(email_address, template_id, personalisation=None):
         return hash_string(details_string)
 
     @staticmethod
-    def get_error_message(email_address, error):
+    def get_error_message(email_address, template_name, error):
```
lfdebrux: @samuelhwilliams has helpfully pointed out that this is a breaking change; but, as it isn't used anywhere outside of `DMNotifyClient`, I reckon it's probably fine to leave this change as a minor bump.

dmutils/email/dm_notify.py
```diff
@@ -82,32 +94,38 @@ def get_reference(email_address, template_id, personalisation=None):
...
```
lfdebrux: @samuelhwilliams has helpfully pointed out that this is a breaking change; but, as it isn't used anywhere outside of DMNotifyClient, I reckon it's probably fine to leave this change as a minor bump.

lfdebrux:
@benvand Good point. This should be a breaking change then.

Given that, I decided to break more stuff ;)

dmutils/email/dm_notify.py
```diff
@@ -17,21 +17,33 @@ class DMNotifyClient:
...
+                Keyword-only argument.
         """
+        if govuk_notify_api_key is None:
+            govuk_notify_api_key = current_app.config["DM_NOTIFY_API_KEY"]
+
+        if templates is None:
+            self.templates = current_app.config.get("NOTIFY_TEMPLATES", {}) if current_app else {}
+
         self.logger = logger or current_app.logger
+
```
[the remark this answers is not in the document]
lfdebrux: I wanted to make it more clear that it wasn't associated with the if statement above. It probably isn't necessary, but I felt it was on my eyes at least.

lfdebrux:
Breaking changes:

- `DMNotifyClient,__init__` parameter `logger` is now keyword-only
- `DMNotifyClient.get_error_message` method has been deleted
- `DMNotifyClient.send_email` parameter `email_address` has been renamed to `to_email_address`
- `DMNotifyClient.get_reference` parameter `email_adress` has been renamed to `to_email_address`

- `dm_mandrill` now contains a single class `DMMandrillClient`
- `dm_mandrill.send_email` has been deleted. Its functionality has been moved to `DMMandrillClient.send_email`, however the function signature has changed.
- `dm_mandrill.get_sent_emails` has been deleted. Its functionality has been moved to `DMMandrillClient.get_sent_emails`, however the function signature has changed.

dmutils/email/dm_notify.py
```diff
@@ -66,54 +80,66 @@ def has_been_sent(self, reference):
...
-        :param template_id: Template accessible on the Notify account whose  api_key you instantiated the class with
+        :param to_email_address: String email address for recipient
+        :param template_name: Template accessible on the Notify account whose api_key you instantiated the class with.
+                              Can either be a UUID or a key to the current_app.config["NOTIFY_TEMPLATES"] dictionary.
         :param personalisation: The template variables, dict
         :param allow_resend: if False instantiate the delivered reference cache and ensure we are not sending duplicates
         :return: response from the api. For more information see https://github.com/alphagov/notifications-python-client
         """
-        reference = reference or self.get_reference(email_address, template_id, personalisation)
+        template_id = self.templates.get(template_name, template_name)
```
katstevens: Hmmm I'm not sure about this conflation of `template_id` and `template_name`. If we want to be able to pass either to this method, we should probably have separate keyword args for them. I can just see us getting confused when we come to use it in one of the apps.

dmutils/email/dm_notify.py
```diff
@@ -17,21 +17,35 @@ class DMNotifyClient:
...
                 `DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS` setting in the current flask app's config (if available).
+
+        The following arguments are keyword-only and should only be used if you want to operate outside
+        of a Flask app context.
+            :param logger: logger to log progress to, taken from current_app if Falsey
+            :param templates: a dictionary of template names to template uuids, so that you can use
+                descriptive names when specifying a template. This defaults to current_app.config["NOTIFY_TEMPLATES"].
         """
+        if govuk_notify_api_key is None:
+            govuk_notify_api_key = current_app.config["DM_NOTIFY_API_KEY"]
```
katstevens: If we're doing this, we can also refactor `dmutils.email.user_account_email.send_user_account_email` to remove the key from the client initialisation. That function also is specifying `template_id` as a kwarg on `send_email`...

lfdebrux:
These commits were cherry-picked into #445 so I'm closing this PR

dmutils/email/dm_notify.py
lfdebrux: Yeah I totally missed that 😳 There is a PR I'm working on now that will fix that bug (#456)

dmutils/email/dm_notify.py
```diff
@@ -66,54 +80,66 @@ def has_been_sent(self, reference):
...
         if not allow_resend and self.has_been_sent(reference):
             self.logger.info(
-                "Email {reference} (template {template_id}) has already been sent to {email_address} through Notify",
+                "Email with reference '{reference}' has already been sent",
                 extra=dict(
-                    email_address=hash_string(email_address),
-                    template_id=template_id,
+                    client=self.client.__class__,
+                    to_email_address=hash_string(to_email_address),
+                    template_name=template_name,
```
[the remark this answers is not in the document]
lfdebrux: My understanding (from @risicle) is that they get emitted in the JSON logs. A similar pattern already existed in this module. I should probably check that this is the case experimentally though...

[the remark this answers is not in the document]
lfdebrux: My understanding (from @risicle) is that they get emitted in the JSON logs. A similar pattern already existed in this module. I should probably check that this is the case experimentally though...

PR #455, opened 2018-09-07
18 flask 1x upgrade

benvand:
http://flask.pocoo.org/docs/1.0/changelog/#version-1-0
https://trello.com/c/bBDir2Gp/18-flask-1x-upgrade

* Prevent overwriting of imported `logger` module in dmutils/logging
* New fixture to stream logs to StreamIO object (replaces `log_to_file` method)
* New fixture to mock `logger` attribute of `app`
* Changes to look for 'flask.app' new logger name
  * http://flask.pocoo.org/docs/1.0/changelog/#version-1-0
* Break out some parametrized tests to reduce complexity
* Use new fixtures where possible
* Add new `autouse` fixture to prevent stdout logging in tests

### DIFF
--- a/tests/test_timing.py
+++ b/tests/test_timing.py
@@ -109,8 +109,8 @@ _messages_expected = OrderedDict((
                 'error': "{name}: {street} - {duration_process}s",
             },
             {
-                'success': AnyStringMatching(r"conftest\.foobar: (\{.*\}|eccles) - [0-9eE.-]+s"),
-                'error': AnyStringMatching(r"conftest\.foobar: (\{.*\}|eccles) - [0-9eE.-]+s"),
+                'success': AnyStringMatching(r"flask\.app\.foobar: (\{.*\}|eccles) - [0-9eE.-]+s"),
+                'error': AnyStringMatching(r"flask\.app\.foobar: (\{.*\}|eccles) - [0-9eE.-]+s"),
             },
         ),
     ),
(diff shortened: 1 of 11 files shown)


tests/test_timing.py
```diff
@@ -511,33 +510,30 @@ def test_logged_duration_real_logger(
...
+                    log_level=log_level,
+                    condition=condition,
+                ) as log_context:
+                    _sleep(sleep_time)
+                    if inject_context is not None:
+                        log_context.update(inject_context)
+                    if raise_exception is not None:
+                        raise raise_exception("Boo")
+
+    stream.seek(0)
```
risicle: Though it doesn't directly matter here, I think it's probably a better habit to use `StringIO`'s `.getvalue()` method for this rather than reaching in and seeking around the file. If someone were to make an assertion like this and leave the (imaginary) file descriptor not at the end of the file & continue on with a couple of extra steps of the test, you'd get some confusing mixed up output and it wouldn't be clear why.

benvand: Yeah I tried that but it meant that I had to mess with the comprehension below quite a lot and I wanted to leave it as similar to before as possible. `get_value` returns the whole string joined by linebreaks 😒

risicle: :confused: Isn't that exactly what `stream.read()` does?

benvand: Good call, can use getvalue and remove the seek line 💯

PR #456, opened 2018-09-10
Update user_account_email kwarg for new client changes

benvand:


lfdebrux:
Sorry I missed this one.

I'll rewrite the test to patch the DMNotifyClient with autospec, that should make sure similar issues are caught in future.

risicle:
Technically I think this should've been another major bump to 45.0.0 - it's not like the numbers cost anything. "Bugfixes" should be transparently upgradable.

PR #457, opened 2018-09-11
changelog: mention rename of `template_id` parameter to `template_name`

risicle:
This is a potentially breaking change, previously not mentioned.


lfdebrux:
This was fixed in #456

PR #460, opened 2018-09-14
[Bugfix] Fixes issue where the selection button widget type couldn't be customised for a DMBooleanField

lfdebrux:
Whilst refactoring forms for the supplier-frontend (PR to follow, I promise), I hit a bug where when I wanted to use a `DMBooleanField` with a slightly different widget I couldn't.

This PR fixes the order in which values are added to the template context to fix this bug; it also adds a few more tests so that the test suite captures more of the behaviour I want/need.

### DIFF
--- a/dmutils/forms/fields.py
+++ b/dmutils/forms/fields.py
@@ -44,9 +44,7 @@ __all__ = ['DMBooleanField', 'DMDecimalField', 'DMHiddenField', 'DMIntegerField'
 
 
 class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
-    widget = DMCheckboxInput(hide_question=True)
-
-    type = "checkbox"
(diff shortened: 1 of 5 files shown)


dmutils/forms/fields.py
```diff
@@ -44,9 +44,7 @@
 
 
 class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
-    widget = DMCheckboxInput(hide_question=True)
-
-    type = "checkbox"
+    widget = DMCheckboxInput(type="checkbox", hide_question=True)
```
katstevens: I might be missing something obvious here, but why did the custom type previously work for the `DMRadioInput` and not `DMCheckboxInput`?

lfdebrux: Fixed now

PR #463, opened 2018-09-28
Remove unittest tests

katstevens:
Trello: https://trello.com/c/eSNBb9bi/571-remove-unittest-tests-in-dmutils-in-favour-of-pytest

An actual quick win - 14 minutes from branch creation to PR 😹 

However I'm not sure how picky we need to be about asserting False in these tests. I've gone for `self.assertFalse(x)` --> `assert not x` and `self.assertEqual(x, False)` --> `assert x is False`.


lfdebrux:
Nice one!

PR #464, opened 2018-10-01
Set flask env config var

katstevens:
Trello: https://trello.com/c/yauaJQ4G/175-flaskenv-defaults-to-production-and-shows-a-harmless-but-worrying-debug-message-in-runserver

Addresses a harmless-but-worrying warning message introduced in Flask 1.0 (see See http://flask.pocoo.org/docs/1.0/config/#environment-and-debug-features). A new config var `ENV` defaults to production, and we weren't setting this properly.

Previously, when running an app locally, you'd get this warning:
```
12:55:43          api |  * Serving Flask app "app" (lazy loading)
12:55:43          api |  * Environment: production
12:55:43          api |    WARNING: Do not use the development server in a production environment.
12:55:43          api |    Use a production WSGI server instead.
12:55:43          api |  * Debug mode: on
```

Now it looks like this:
```
 * Serving Flask app "app" (lazy loading)
 * Environment: development
 * Debug mode: on
```

Having `ENV` set to production (value can be either `production` or `development`) should be fine for preview/staging (that's what they are using right now!). As far as I can work out, Flask is only using it for displaying the helpful message above, and toggling debug mode.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '44.5.0'
+__version__ = '44.6.0'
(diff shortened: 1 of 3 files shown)


katstevens:
Thanks both!

PR #466, opened 2018-10-08
get_api_endpoint_from_stage: teach this about the antivirus api being on port 5008

risicle:


dmutils/env_helpers.py
```diff
@@ -16,6 +16,7 @@ def get_api_endpoint_from_stage(stage, app='api'):
     dev_ports = {
         'api': 5000,
         'search-api': 5001,
+        'antivirus-api': 5008,
```
katstevens: I think we have tests for these as well - `TestGetAPIEndpointFromStage`

risicle: Ohhhhright

risicle: Fixed

risicle:
:thunderbird_2:

PR #470, opened 2018-11-02
"external" routes: add route for direct_award_service_page

risicle:
https://trello.com/c/kjIZjvPz/224-external-route-for-public-g-cloud-service-page-frameworkfamily-services-serviceid

### DIFF
--- a/dmutils/external.py
+++ b/dmutils/external.py
@@ -26,6 +26,14 @@ def list_opportunities(framework_family):
     raise NotImplementedError()
 
 
+@external.route('/<framework_family>/services/<service_id>')
+def direct_award_service_page(framework_family, service_id):
+    if framework_family == 'suppliers':
+        # nginx would otherwise route this pattern to the Supplier FE
+        abort(404)
+    raise NotImplementedError()
+
+
 @external.route('/g-cloud/suppliers')
 def suppliers_list_by_prefix():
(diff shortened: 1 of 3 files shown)


dmutils/external.py
```diff
@@ -26,6 +26,14 @@ def list_opportunities(framework_family):
     raise NotImplementedError()
 
 
+@external.route('/<framework_family>/services/<service_id>')
+def direct_award_service_page(framework_family, service_id):
+    if framework_family == 'suppliers':
+        # nginx would otherwise route this pattern to the Brief Responses FE
```
katstevens: This was my mistake on the ticket description - `/suppliers` would route to Suppliers FE (Brief Responses would be `/suppliers/opportunities` - dunno if we should bother checking for that string here as well?)

risicle: I'm thoroughly confused again. What is it I'm supposed to be doing now?

katstevens: It's fine, it's just the comment that mentions the wrong app.

Although `/suppliers/opportunities/services/1234`, `/users/services/1234` etc would get routed wrongly as well, I think it's much less likely that a user would try and hack the URL to those, than for `/suppliers/services/1234` (which could conceivably be a 'supplier's service page').

risicle: This right now?

PR #471, opened 2018-11-05
Pin MailChimp3 library to 2.0.11

benvand:
Pin MailChimp3 library to 2.0.11 (most commonly used across apps) after non semver'd breaking changes

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '44.8.2'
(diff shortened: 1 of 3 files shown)


katstevens:
We might have to revert this change to the `mc_secret`/`mc_api` keyword: https://github.com/alphagov/digitalmarketplace-utils/pull/465/files

benvand:
@katstevens Yeah, tests caught that too 🏆 
There are other changes to the tests in that PR but I'm going to leave them in because they're good changes anyway.

benvand:
@katstevens I've updated the PR and commit message too, but this passed locally so should be good for review proper now

PR #472, opened 2018-11-06
169-file-upload-sniffing-type-restrictions

benvand:
Add fleep checks to existing extension checks https://trello.com/c/LI12IWmy/169-file-upload-sniffing-type-restrictions

* New function get_first_128_bytes
  * Get first 128 bytes but replace file pointer when done
  * Use in `file_is_empty`, no need to load whole file in to memory
  * Use in fleep checking, only needs first 128 bytes
* Move `get_extension` definition above all the methods that call it
* New function `extension_matches_possible_file_formats`
  * Checks provided extension against fleeps file type guesses
* New function `get_possible_extensions_from_format`
  * Return file type guesses for provided file object
* Update opendoc, pdf, csv, zip and image checking to include fleep checks
* Do not depend on S3 module for FILE_SIZE_LIMIT in file_is_less_than_5mb
  * If that changes in the s3 module our `file_is_less_than_5mb` validator will no longer be checking that a file is less than 5mb
* Add fleep dependency to setup.py
* Import ordering in test file
* Add test files for use in tests
* Test for correct file as well as correct extension

169-file-upload-sniffing-type-restrictions: Update tests for suffix checking to onclude fleep requirements

New image test files

udates and tests

test f ts

foo


dmutils/documents.py
```diff
@@ -132,56 +134,93 @@ def upload_service_documents(uploader, upload_type, documents_url, service, requ
...
 def file_is_less_than_5mb(file_):
-    return get_file_size(file_) < FILE_SIZE_LIMIT
+    return get_file_size(file_) < 5400000
+
+
+def get_first_128_bytes(file_object):
+    """Get the contents of a file like object but replace the pointer when we're finished"""
+    position = file_object.tell()
+    file_object.seek(0)
+    contents = file_object.read()
```
katstevens: Should this be `file_object.read(128)`?

dmutils/documents.py
```diff
@@ -132,56 +134,93 @@ def upload_service_documents(uploader, upload_type, documents_url, service, requ
...
+    :param file_object: File object open as bytes.
+    :param extension: File object open as bytes.
+    :return: boolean: Does extension match guess at format?
+    """
+    return fleep.get(get_first_128_bytes(file_object)).extension_matches(extension.strip('.'))
 
 
 def file_is_open_document_format(file_object):
-    return get_extension(file_object.filename) in [
-        ".pdf", ".pda", ".odt", ".ods", ".odp"
```
katstevens: Does anyone use `.pda` docs? I know it's not in the fleek list, but should we allow `.pda` extensions without the format check?

benvand: There's no such thing as far as I've been able to work out. Not in terms of a type of open doc. I brought it up on slack a while ago and no one had heard of it and said :partytrashfire:

dmutils/documents.py
```diff
@@ -132,56 +134,93 @@ def upload_service_documents(uploader, upload_type, documents_url, service, requ
...
+
+    return extension == ".zip" and extension_matches_possible_file_formats(file_object, extension)
 
 
 def file_is_image(file_object):
-    """Checks file extension as being JPG. or PNG."""
-    return get_extension(file_object.filename) in [
-        ".jpg", ".jpeg", ".png"
-    ]
+    """Checks file extension and format as being JPG. or PNG. JPEG is notmalised to JPG"""
```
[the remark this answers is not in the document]
benvand: Done 👍

dmutils/documents.py
benvand: Doh! Good catch, yeah needs to be 128. Could also be `read()[:127]` but that loads the whole file I think

PR #473, opened 2018-11-07
New python script for creating frozen requirements.txt file

benvand:
https://trello.com/c/nAbDKL43/228-pip-is-case-insensitive-and-our-requirements-freeze-job-is-not

New python based `freeze-requirements` that fails when pip detects anything in the new requirements file that wasn't installed in the virtualenv, including packages with the wrong case specified in their name (eg rEqUeStS)

This is an MVP and direct rip off of our makefile `freeze-requirements` command but written in python.

https://github.com/alphagov/notifications-api/blob/master/Makefile#L96-L105
https://github.com/alphagov/digitalmarketplace-api/blob/master/Makefile#L31-L40

Creates an identical `requirements.txt` to the one we already have in supplier-frontend and briefs.

Is ran like:
`python -m dmutils.repoutils.freeze-requirements requirements-app.txt `


dmutils/repoutils/freeze-requirements.py
```diff
@@ -0,0 +1,108 @@
...
+    freeze-requirments <target> [options]
+
+Options:
+    <target>             A requirements file for us to freeze
+    --outfile=<outfile>  Output file for frozen requirement values [default: requirements.txt]
+    --verbose
+    -h, --help           Show this screen
+
+Examples:
+    python -m dmutils.repoutils.freeze-requirments requirements-app.txt
```
katstevens: I just tried running this locally (against the supplier FE, in the virtualenv) and it seems fine. However the current `make freeze-requirements` is done outside of the supplier FE virtualenv. The `make freeze-requirements` Makefile command have to be something like `${VIRTUALENV_ROOT}/bin/python -m [etc]`, so we should probably update this example. (What happens if we're bootstrapping and haven't installed utils yet? Does this ever happen in practice?)

Also there are a few typos `freeze-requirments` 😸

benvand: Doh, typos, will correct.

Looks like the current `make freeze-requirements` will operate inside or outside of the venv using whatever python interpreter it finds in `echo $PATH`.
But you're right, in the case of our repos this _shouldn't_ be run outside of the venv because we want to ensure we're using the right utils version.

If we specify that here we'd have to say something like 'You can't use this module without installing it' which is true for any module.

What we do in other cases, like for `flake8` and `py.test` is force this in the makefile like so:

```
.PHONY: test-flake8
test-flake8: virtualenv
	${VIRTUALENV_ROOT}/bin/flake8 .
```

So we could do:
```
.PHONY: freeze-requirements
freeze-requirements: virtualenv
	${VIRTUALENV_ROOT}/bin/python -m dmutils.repoutils.freeze-requirements requirements-app.txt
```

benvand: typos done 👍

benvand: (ie there should always be a venv available)

dmutils/repoutils/freeze-requirements.py
```diff
@@ -0,0 +1,111 @@
...
+    logger.info(f'Creating virtualenv: {virtualenv_name}')
+    venv.main(args=(virtualenv_name,))
+
+    logger.info(f'Installing requirements from {target} in virtualenv: {virtualenv_name}')
+    install_cmd = subprocess.Popen(
+        [f'{virtualenv_name}/bin/pip', 'install', '-r', target],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    stdout, stderr = install_cmd.communicate()
```
lfdebrux: I think using `subprocess.run` instead of `Popen` could simplify a lot of the code here?

lfdebrux:
```suggestion
    install_cmd = subprocess.run(
        [f'{virtualenv_name}/bin/pip', 'install', '-r', target],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE
    )
    if install_cmd.returncode >= 1:
```

benvand: Quality catch! I was looking at the python 2 docs for subprocess https://docs.python.org/2/library/subprocess.html and it had all these warnings about pipe.

Python 3 docs on the other hand:
https://docs.python.org/3/library/subprocess.html#subprocess.run

Looks like I can use `capture_output=True` and do away with the pipes even.

This is the subprocess method we've all been waiting for! 💯 😍

benvand: Damn, `capture_output` added in `3.7` :(

benvand:
@lfdebrux Updated to use `run`

dmutils/repoutils/freeze-requirements.py
```diff
@@ -0,0 +1,111 @@
...
+    logger.info(f'Creating virtualenv: {virtualenv_name}')
+    venv.main(args=(virtualenv_name,))
+
+    logger.info(f'Installing requirements from {target} in virtualenv: {virtualenv_name}')
+    install_cmd = subprocess.run(
+        [f'{virtualenv_name}/bin/pip', 'install', '-r', target],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    stdout, stderr = str(install_cmd.stdout.decode('utf-8')), str(install_cmd.stderr.decode('utf-8'))
```
lfdebrux: Sneaky trick, you can use `universal_newlines=True` as an argument to run and then you can get rid of this line 😁 

In Python 3.7 their is a more sensible argument called `text`.

lfdebrux:
```diff
         stderr=subprocess.PIPE
+      universal_newlines=True
     )
-    stdout, stderr = str(install_cmd.stdout.decode('utf-8')), str(install_cmd.stderr.decode('utf-8'))
```

benvand: Damn, nice subprocess-fu, done 👍

lfdebrux: You can tell I've been reading through the manual recently 😉

benvand: Thanks for that , going to merge this now 👍

PR #475, opened 2018-11-19
Update freeze_requirements to strip install args on freeze -r

benvand:
I misread the original `freeze-requirements` Makefile target. It pipes in the `requirements-app.txt` minus the `--` lines to pip freeze but the `requirements.txt` gets the `--` lines.
This updates the script with that behaviour.

```
freeze-requirements:
       rm -rf venv-freeze
       python3 -m venv venv-freeze
       $$(pwd)/venv-freeze/bin/pip install -r requirements-app.txt
       echo '# This file is autogenerated. Do not edit it manually.' > requirements.txt
       cat requirements-app.txt >> requirements.txt
       echo '' >> requirements.txt
       $$(pwd)/venv-freeze/bin/pip freeze -r <(sed '/^--/d' requirements-app.txt) | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
       rm -rf venv-freeze
```


dmutils/repoutils/freeze_requirements.py
```diff
@@ -72,12 +73,14 @@ def exit_failure(messages, code=1):
...
-    freeze_cmd = subprocess.run(
-        [f'{virtualenv_name}/bin/pip', 'freeze', '-r', target],
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        universal_newlines=True
-    )
+    with open(target, 'rb') as requirements_file, tempfile.NamedTemporaryFile() as tmp:
+        tmp.write(re.sub(b"--.+?\n", b"", requirements_file.read()))
+        freeze_cmd = subprocess.run(
+            [f'{virtualenv_name}/bin/pip', 'freeze', '-r', f'{tmp.name}'],
```
lfdebrux: As an alternative to creating a tempfile you could use `shell=True` in the arguments to subprocess.run; then the context manager would be removed and the command to subprocess would be the same as in the Makefile:

```suggestion
            f"{virtualenv_name}/bin/pip freeze -r <(sed '/^--/d' {target})"],
            shell=True,
```

The functionality would be exactly the same, real question is which is more readable.

benvand: I wanted to do as much as possible in python, including the replace. I think part of the reason I made this mistake in the first place is my unfamiliarity with bash so more python > more bash for me.

I've also heard it's best to avoid `shell=True` although I'm not 100% sure I understand why.

lfdebrux: Awesome, leave it as it is then 👍

PR #481, opened 2018-12-04
538 set sensible timeout on mailchimp client

benvand:
Add sensible timeout to Mailchimp client  …
We should time out before 30 seconds.
Our hosting provder (paas) has a timeout of 30 secs for a request.
If we hang connecting the MailChimp (ie if mailchimp is down) PaaS severs the connection displaying an ugly 504 page.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '45.1.0'
+__version__ = '45.1.1'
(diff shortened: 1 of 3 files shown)


benvand:
@lfdebrux it propagates that new argument right down to the creation of the unix socket. So I raised an exception from as close to the OS as I could get in python.
Also had a look at what was raised from `requests` when you get a timeout:
https://stackoverflow.com/questions/100841/artificially-create-a-connection-timeout-error/21905365
by requesting a non-routable IP.

But in terms of integration none. Not sure how we could do it. Rerouting requests in the preview env made to the mailchimp domain to a non-routable ip _could_ work. But I'm not even sure if that's possible.

PR #482, opened 2018-12-05
add common error handlers for "api" flask apps

risicle:
https://trello.com/c/czdGHEeg/262-api-error-handlers

This takes the work done by @katstevens bringing the frontend's error handlers into `-utils` and does something similar for "api" flask app's error handlers.

It stops short of also pulling in & commonizing auth-handling routines, and so what is actually pulled in is just a few key functions and the ability for `init_app` to hook them up in a default-yet-customizable way.

Another thing that I _didn't_ do was go down the whole path that modern werkzeug seems to be taking of making almost everything a subclass of `HTTPError` which has the ability to "render itself" if asked. I felt this was too much of a "forced" abstraction for two reasons:

1. It felt like a layering violation making our models spew `ValidationError`s that were actually `HTTPError`s.
2. The mechanism doesn't really allow for easy "universal" response customization, e.g. wrapping the whole thing in json, because the inheritance tree of exceptions splits out at the top level and we'd have to either customize all the subclasses individually or cook up something clever using mixins. So I thought, for now, it's better to leave this sort of processing done by a simple handler function.

I have (untidy) patches for each of the apis that seem to prove that this treatment is universal enough to work for all of them.


dmutils/flask_init.py
```diff
@@ -1,7 +1,33 @@
...
+    (401, fe_errors.redirect_to_login,),
+    (403, fe_errors.redirect_to_login,),
+    (404, fe_errors.render_error_page,),
+    (410, fe_errors.render_error_page,),
+    (503, fe_errors.render_error_page,),
+    (500, fe_errors.render_error_page,),
+)))
+
+
+api_error_handlers = MappingProxyType(OrderedDict(chain(
```
benvand: What are we using the `MappingProxyType` for here? Can we just use a dict (not even an ordered one I think)? Also you can just `+` tuples together rather than importing `chain`.


```
api_error_handlers = MappingProxyType(OrderedDict(chain(
    (
        (api_errors.ValidationError, api_errors.validation_error_handler,),
    ),
    ((code, api_errors.json_error_handler) for code in default_exceptions),
)))
```

```
dict(((api_errors.ValidationError, api_errors.validation_error_handler),) + ((code, api_errors.json_error_handler) for code in default_exceptions)
```

dmutils/errors/api.py
```diff
@@ -0,0 +1,29 @@
...
+        # initially we'll try and assume this is an HTTPException of some sort.  for the most part, the default
+        # HTTPExceptions render themselves in the desired way if returned as a response. the only change we want to
+        # make is to enclose the error description in json.
+        response = e.get_response()
+        response.set_data(json.dumps({"error": e.description}))
+        response.mimetype = current_app.config["JSONIFY_MIMETYPE"]
+
+        return response
+    except Exception:
+        # either `e` wasn't an HTTPException or something went wrong in trying to jsonify it
```
benvand: Will the error be logged anywhere?

dmutils/authentication.py
```diff
@@ -0,0 +1,25 @@
+from werkzeug.exceptions import Unauthorized
+
+
```
benvand: Does this just send back the Auth header they sent to us?

dmutils/errors/api.py
risicle: Yyyyyyes - I think that should have already happened by this point.

dmutils/authentication.py
risicle: Not quite - this is used in cases when no auth header was provided in the request, the response adding a `WWW-Authenticate` header describing what scheme we want (`Bearer` in most cases) and what "auth domain" this is for (arbitrary but we'll mostly be using `main` I think).

dmutils/flask_init.py
risicle: Uh yeah we could probably do `((api_errors.ValidationError, api_errors.validation_error_handler,),) + tuple((code, api_errors.json_error_handler) for code in default_exceptions)`

As for `MappingProxyType`, it just makes the dict immutable, which makes it easier to reason about and harder to do weird and unexpected things to.

PR #483, opened 2018-12-28
DMMailChimpClient: add get_lists_for_email(...) and permanently_remove_email_from_list(...) methods

risicle:
To support https://trello.com/c/QROOUpoX/272-users-whose-personal-data-is-removed-are-not-removed-from-mailchimp-mailing-lists

Note this includes a bump to `mailchimp3` v3.x.

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -3,7 +3,8 @@
 
 from json.decoder import JSONDecodeError
 from hashlib import md5
-from typing import Mapping, Sequence
+from logging import Logger
+from typing import Callable, Iterator, Mapping, Sequence, Union
 
 from requests.exceptions import RequestException, HTTPError
 
@@ -25,22 +26,22 @@ class DMMailChimpClient(object):
 
     def __init__(
         self,
-        mailchimp_username,
-        mailchimp_api_key,
-        logger,
-        retries=0
+        mailchimp_username: str,
+        mailchimp_api_key: str,
+        logger: Logger,
+        retries: int=0,
     ):
         self._client = MailChimp(mc_user=mailchimp_username, mc_api=mailchimp_api_key, timeout=25)
         self.logger = logger
         self.retries = retries
 
     @staticmethod
-    def get_email_hash(email_address):
+    def get_email_hash(email_address: Union[str, bytes]) -> bytes:
         """md5 hashing of lower cased emails has been chosen by mailchimp to identify email addresses"""
         formatted_email_address = str(email_address.lower()).encode('utf-8')
         return md5(formatted_email_address).hexdigest()
 
-    def timeout_retry(self, method):
+    def timeout_retry(self, method: Callable) -> Callable:
         def wrapper(*args, **kwargs):
             for i in range(1 + self.retries):
                 try:
@@ -55,7 +56,7 @@ class DMMailChimpClient(object):
 
         return wrapper
 
-    def create_campaign(self, campaign_data):
+    def create_campaign(self, campaign_data: Mapping) -> Union[str, bool]:
         try:
             with log_external_request(service='Mailchimp'):
                 campaign = self._client.campaigns.create(campaign_data)
@@ -72,7 +73,7 @@ class DMMailChimpClient(object):
             )
         return False
 
-    def set_campaign_content(self, campaign_id, content_data):
+    def set_campaign_content(self, campaign_id: str, content_data: Mapping):
         try:
             with log_external_request(service='Mailchimp'):
                 return self._client.campaigns.content.update(campaign_id, content_data)
@@ -86,7 +87,7 @@ class DMMailChimpClient(object):
             )
         return False
 
-    def send_campaign(self, campaign_id):
+    def send_campaign(self, campaign_id: str):
         try:
             with log_external_request(service='Mailchimp'):
                 self._client.campaigns.actions.send(campaign_id)
@@ -101,7 +102,7 @@ class DMMailChimpClient(object):
             )
         return False
 
-    def subscribe_new_email_to_list(self, list_id, email_address):
+    def subscribe_new_email_to_list(self, list_id: str, email_address: str):
         """Will subscribe email address to list if they do not already exist in that list else do nothing"""
         hashed_email = self.get_email_hash(email_address)
         try:
@@ -144,7 +145,7 @@ class DMMailChimpClient(object):
             )
             return False
 
-    def subscribe_new_emails_to_list(self, list_id, email_addresses):
+    def subscribe_new_emails_to_list(self, list_id: str, email_addresses: str) -> bool:
         success = True
         for email_address in email_addresses:
             with log_external_request(service='Mailchimp'):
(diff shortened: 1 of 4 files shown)


dmutils/email/dm_mailchimp.py
```diff
@@ -161,3 +163,35 @@ def get_email_addresses_from_list(self, list_id, pagination_size=100):
             offset += pagination_size
 
             yield from [member['email_address'] for member in member_data['members']]
+
+    def get_lists_for_email(self, email_address: str) -> Sequence[Mapping]:
```
jonodrew: I like typing, but I thought when I tried this we agreed not to do it piecemeal?

risicle: I don't remember how much I was on board with that - I think it's *specifically* useful here because it's an API method, and helps explain the interface.

risicle: (I wonder if newer versions of `mock` are considering type-checking autospecced functions that have typing...?)

tests/email/test_dm_mailchimp.py
```diff
@@ -364,3 +364,74 @@ def test_offset_increments_until_no_members(self):
...
+        with mock.patch.object(
+            dm_mailchimp_client._client.lists.members,
+            'delete_permanent',
+            autospec=True,
+        ) as del_perm:
+            del_perm.side_effect = RequestException("No thoroughfare")
+
+            with assert_external_service_log_entry(successful_call=False, extra_modules=['mailchimp'], count=1):
+                result = dm_mailchimp_client.permanently_remove_email_from_list(
+                    "Trousers.Potato@purse.net",
```
jonodrew: I feel potato/Irish comments on a public, government-owned repo are not appropriate.

tests/email/test_dm_mailchimp.py
```diff
@@ -364,3 +364,74 @@ def test_offset_increments_until_no_members(self):
                     mock.call('a_list_id', count=100, offset=600),
                     mock.call('a_list_id', count=100, offset=700),
                 ]
+
+    def test_get_lists_for_email(self):
+        dm_mailchimp_client = DMMailChimpClient('username', DUMMY_MAILCHIMP_API_KEY, logging.getLogger('mailchimp'))
+        with mock.patch.object(dm_mailchimp_client._client.lists, 'all', autospec=True) as all_lists:
+            all_lists.return_value = {
+                "lists": [
+                    {"id": "gadz00ks", "name": "Pistachios", "irrelevant": "custard"},
```
jonodrew: What's the point of these `irrelevant` and `meaningless` keys here?

dmutils/email/dm_mailchimp.py
```diff
@@ -161,3 +163,35 @@ def get_email_addresses_from_list(self, list_id, pagination_size=100):
...
+        """
+            Returns a sequence of all lists the email_address has an association with (note: even if that association is
+            "unsubscribed" or "cleaned").
+        """
+        with log_external_request(service='Mailchimp'):
+            return tuple(
+                {
+                    "list_id": ml["id"],
+                    "name": ml["name"],
+                } for ml in self._client.lists.all(get_all=True, email=email_address)["lists"]
```
jonodrew: I'm not sure what an `ml` is. Could you use a more descriptive variable name?

tests/email/test_dm_mailchimp.py
risicle: What's Irish here?

tests/email/test_dm_mailchimp.py
risicle: It's to prove that keys other than `id` and `name` don't affect behaviour.

dmutils/email/dm_mailchimp.py
risicle: ml, "mailing list", thought that was pretty universal.

tests/email/test_dm_mailchimp.py
jonodrew: line 427: `paddy_kellys_budget` ?

Paddy is a stereotypical Irish name, and [Paddy Kelly is Irish](https://en.wikipedia.org/wiki/Michael_Patrick_Kelly).

tests/email/test_dm_mailchimp.py
jonodrew: Could you use variable names that reflect that, then?

dmutils/email/dm_mailchimp.py
jonodrew: I don't think it is. Could you change it?

risicle:
Each test checks the behaviour of a single method - its return value/exception and any side effects it has. It so happens that the "permanently remove" method has two majorly different behaviours it exhibits, so there are two separate tests.

tests/email/test_dm_mailchimp.py
risicle: `irrelevant`.

dmutils/email/dm_mailchimp.py
```diff
@@ -161,3 +163,35 @@ def get_email_addresses_from_list(self, list_id, pagination_size=100):
...
+                    list_id=list_id,
+                    subscriber_hash=hashed_email,
+                )
+            return True
+        except RequestException as e:
+            self.logger.error(
+                f"Mailchimp failed to permanently remove user ({hashed_email}) from list ({list_id})",
+                extra={"error": str(e), "mailchimp_response": get_response_from_request_exception(e)},
+            )
+        return False
```
katstevens: If it's going to be a script (Jenkins job?) that runs this method, do we want failures to be noisier? Or should the script handle that?

katstevens:
I think the tests are among some of @risicle's more readable ones 😹 There's a lot going on with the object patches and logging assertions, but I'm not sure how we'd re-write it to be much clearer.

setup.py
```diff
@@ -31,7 +31,7 @@
          'botocore<1.11.0',
          'contextlib2>=0.4.0',
          'cryptography<2.4,>=2.3',
-         'mailchimp3<2.1.0,>=2.0.11',
+         'mailchimp3==3.0.6',
```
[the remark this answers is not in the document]
risicle: I ... think ... so?

dmutils/email/dm_mailchimp.py
risicle: I probably agree, but was doing it this way to match the rest of the methods in this class, which appear to do the return-boolean thing.

dmutils/email/dm_mailchimp.py
risicle: Are you trying to tempt me to do the whole class?

setup.py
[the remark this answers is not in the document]
risicle: Yeah I think it was init arg order changes which I've ... compensated for here....? (goes to check)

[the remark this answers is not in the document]
risicle: Ah yes it was an init kwarg name change, which I've dealt with.

risicle:
> Also next time let's use a different European country's literary references for our test data

*You're* welcome to do as you like...

PR #484, opened 2019-01-09
Update mailchimp error handling for new version

benvand:
`mailchimp3` has changes the errors returned in it's client. This PR updates the `DMMailChimpClient` to gracefully handle these errors.

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -9,17 +9,22 @@ from typing import Callable, Iterator, Mapping, Sequence, Union
 from requests.exceptions import RequestException, HTTPError
 
 from mailchimp3 import MailChimp
+from mailchimp3.mailchimpclient import MailChimpError
 
 from dmutils.timing import logged_duration_for_external_request as log_external_request
 
 PAGINATION_SIZE = 1000
 
 
-def get_response_from_request_exception(exc):
-    try:
-        return exc.response.json()
-    except (AttributeError, ValueError, JSONDecodeError):
-        return {}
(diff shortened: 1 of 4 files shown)


dmutils/email/dm_mailchimp.py
```diff
@@ -9,17 +9,22 @@
...
-    except (AttributeError, ValueError, JSONDecodeError):
-        return {}
+def get_response_from_exception(exc):
+    if isinstance(exc, RequestException):
+        try:
+            return exc.response.json()
+        except (AttributeError, ValueError, JSONDecodeError):
+            pass
+    if isinstance(exc, MailChimpError):
+        return exc.args[0]
```
katstevens: We don't mind if this doesn't return JSON, right - we just log the error? Is there a risk of logging PII?

katstevens:
We have a few tests around `RequestException` in `test_dm_mailchimp` - worth adding in a quick check there for the new exception type?

dmutils/email/dm_mailchimp.py
benvand: The `MailChimpError` is always passed an empty dictionary or the response json as its first and only argument in 
https://github.com/charlesthk/python-mailchimp/blob/master/mailchimp3/mailchimpclient.py

This will log PII if the Mailchimp api sends PII back in error messages excepting in the cases where we explicitly avoid doing so (which we do for the error we received last night)
https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/email/dm_mailchimp.py#L136

benvand:
@risicle LOL

@katstevens Done 👍

PR #485, opened 2019-01-10
This improves the send_email capability for the DMNotifyClient

jonodrew:
I've added a way for users of this API to send an email with a reply-to address. This is capability that has been [added to the Notify library](https://docs.notifications.service.gov.uk/python.html#email-reply-to-id-optional), which this wraps. 

I've set the argument to default to `None`, and changed the tests so they reflect the new argument.

To use the reply-to email, it has to first be set up with Notify. It's really only suitable for reusing a common reply-to email address, but luckily that's our only use case for the moment.

### DIFF
--- a/tests/email/test_dm_notify.py
+++ b/tests/email/test_dm_notify.py
@@ -115,7 +115,25 @@ class TestDMNotifyClient(PatchExternalServiceLogConditionMixin):
                 self.email_address,
                 self.template_id,
                 personalisation=None,
-                reference='niC4qhMflcnl8MkY82N7Gqze2ZA7ed1pSBTGnxeDPj0='
+                reference='niC4qhMflcnl8MkY82N7Gqze2ZA7ed1pSBTGnxeDPj0=',
+                email_reply_to_id=None
+            )
+
+    def test_send_email_passes_correct_reply_to(self, dm_notify_client, notify_send_email):
+        """
+        Test the `send_email` function with a `reply_to_address_id` and confirm it's passed through
+        """
+        with mock.patch(self.client_class_str + '.' + 'send_email_notification') as email_mock:
+            email_mock.return_value = notify_send_email
+
+            dm_notify_client.send_email(self.email_address, self.template_id, reply_to_address_id='123456789')
+
+            email_mock.assert_called_with(
+                self.email_address,
+                self.template_id,
+                personalisation=None,
+                reference='niC4qhMflcnl8MkY82N7Gqze2ZA7ed1pSBTGnxeDPj0=',
+                email_reply_to_id='123456789'
             )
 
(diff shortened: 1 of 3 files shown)


tests/email/test_dm_notify.py
```diff
@@ -115,7 +115,8 @@ def test_send_email_creates_reference(self, dm_notify_client, notify_send_email)
                 self.email_address,
                 self.template_id,
                 personalisation=None,
-                reference='niC4qhMflcnl8MkY82N7Gqze2ZA7ed1pSBTGnxeDPj0='
+                reference='niC4qhMflcnl8MkY82N7Gqze2ZA7ed1pSBTGnxeDPj0=',
+                email_reply_to_id=None
```
lfdebrux: Why is `email_reply_to_id` set to `None` here?

lfdebrux:
Code looks good, not sure about the changes to the tests though.

Commit message could be more self-explanatory also?

tests/email/test_dm_notify.py
lfdebrux: 🤦‍♂️

jonodrew:
> Code looks good, not sure about the changes to the tests though.
> 
> Commit message could be more self-explanatory also?

Is the commit message better like that?

lfdebrux:
> > Code looks good, not sure about the changes to the tests though.
> > Commit message could be more self-explanatory also?
> 
> Is the commit message better like that?

I was more referring to the first line; since this is a pretty small change the first line of the commit message should quickly show what it does in detail: add reply_to email address parameter to send_email

jonodrew:
> 
> I was more referring to the first line; since this is a pretty small change the first line of the commit message should quickly show what it does in detail: add reply_to email address parameter to send_email

I did not know that! Have you got a guide for stuff like this? Other places have been quite hot on very explicit messages so a guide to how we do it here would be helpful

lfdebrux:
> > I was more referring to the first line; since this is a pretty small change the first line of the commit message should quickly show what it does in detail: add reply_to email address parameter to send_email
> 
> I did not know that! Have you got a guide for stuff like this? Other places have been quite hot on very explicit messages so a guide to how we do it here would be helpful

There are other guides, but this is the most popular one https://chris.beams.io/posts/git-commit/

If you find yourself going tl;dr just look at the [seven rules](https://chris.beams.io/posts/git-commit/#seven-rules)

katstevens:
I'd add in a basic test to check that the param is actually getting passed through to the client, otherwise looks good.

jonodrew:
> I'd add in a basic test to check that the param is actually getting passed through to the client, otherwise looks good.

To the underlying Notify client, or to `DMNotifyClient` ? The latter seems to make sense because the `mock` reports it's called with the new parameter. I'm not sure how to check it's being passed to the underlying client.

PR #486, opened 2019-01-21
repoutils.freeze_requirements: upgrade pip to latest version in newly created venv

risicle:
This makes me feel extremely dirty, but at least it's consistently inconsistent between different peoples different machines. Increasing numbers of packages with their sub-dependencies are sensitive to pip
versions, so freezes can occasionally fail if the system pip is used as it is often quite old.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -23,7 +23,7 @@ in (with args; {
 
     hardeningDisable = pkgs.stdenv.lib.optionals pkgs.stdenv.isDarwin [ "format" ];
 
-    VIRTUALENV_ROOT = "venv${pythonPackages.python.pythonVersion}";
+    VIRTUALENV_ROOT = (toString (./.)) + "/venv${pythonPackages.python.pythonVersion}";
     VIRTUAL_ENV_DISABLE_PROMPT = "1";
     SOURCE_DATE_EPOCH = "315532800";
 
@@ -37,7 +37,8 @@ in (with args; {
         ${pythonPackages.python}/bin/python -m venv $VIRTUALENV_ROOT
       fi
       source $VIRTUALENV_ROOT/bin/activate
-      make requirements${pkgs.stdenv.lib.optionalString forDev "-dev"}
+      pip install --upgrade pip==18.0  # some packages are sensitive to "old" pips
+      make -C ${toString (./.)} requirements${pkgs.stdenv.lib.optionalString forDev "-dev"}
     '';
   }).overrideAttrs (if builtins.pathExists localOverridesPath then (import localOverridesPath args) else (x: x));
 })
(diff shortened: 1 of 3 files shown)


benvand:
> packages with their sub-dependencies are sensitive to pip versions

Is there an example package we can test this with?

risicle:
I stubled upon this updating `-api`'s `-apiclient` dependency to the newest one. It insisted on using "vanilla python" pip (10.0 for me) which was unable to find `cffi >= 1.1` for `bcrypt`.

benvand:
I could not reproduce this issue. What is the actual problem. It seems like there could be a cleaner solution.

benvand:
```
benvandersteen@GDS5077:~$ pip3 --version
pip 9.0.1 from /usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages (python 3.6)
```

katstevens:
Mine is `pip 10.0.1`...

benvand:
I tried it with different combinations. The output of `pip install` changed slightly depending on pip version but nothing else.

risicle:
For me, freezing with pip 10.0 resulted in the above error failing to be able to find a dependency `cffi >= 1.1`. I suggest you two don't have vanilla python installs, because I know I have. Also there is the OS difference which could play a role.

benvand:
What above error? What do you mean by 'vanilla' python?

risicle:
> I stumbled upon this updating `-api`'s `-apiclient` dependency to the newest one. It insisted on using "vanilla python" pip (10.0 for me) which was unable to find `cffi >= 1.1` for `bcrypt`.

I don't have the exact error line right now. Vanilla python as in, isn't taking heed of any homedir-based python environment additions or system-installed additions. Even so, I suggest the fact that all three of us are clearly using different `pip` versions to execute the freeze procedure is a possible bad thing and should at least be "standardized".

risicle:
(my guess however is that it's a macos/linux difference)

benvand:
I think we should roll this back.
1) We can't reproduce the problem at all
2) If the _version_ of pip can result in different dependencies being installed then it also doesn't solve that problem and we should find a different solution.

risicle:
I don't think it stops different versions of packages being installed (well, a light breeze can do that: pip is *not* a reproducible build system) - in the case I found it results in it simply failing.

risicle:
So in short you'd prefer us to all just be running random difference versions of pip when this performing this process?

risicle:
(I'd be all for telling it to simply install version `N` of pip, but I just thought people would find that annoying)

benvand:
But pip 9.0.1, 10.0.0, 10.0.1 and 18.1 can all install `cffi`. So we can't reproduce the problem. And this if therefore just a hack for an error we saw once, on a single dev environment, and were never able to reproduce 🤷‍♂️  

We can't merge every fix like that; imagine what our codebase would be like.

PR #487, opened 2019-01-22
Revert PR 486

risicle:
https://github.com/alphagov/digitalmarketplace-utils/pull/486

Because it's ok for my dev environment to be broken.


risicle:
Something tells me we'll be revisiting this when we want to run the freeze procedure on jenkins or travis or pyup or something.

risicle:
This is due to https://github.com/pypa/pip/issu es/5171 and will seemingly reappear on pip <= 10 whenever a package's build dependency doesn't itself have a binary wheel available for its exact platform.

PR #491, opened 2019-02-12
Add `sub_country_codes` filter

lfdebrux:
Being able to convert country codes to country names is a functionality that currently exists in multiple places in our codebase; this PR adds a function that can be used for this purpose generally.

The implementation relies on a new Python library [`govuk-country-register`], which is maintained by myself (@lfdebrux). The reason for doing this rather than simply adding our existing code to dmutils was twofold:

- firstly, there were a number of implementations used, some of which called the register API using requests, others which relied on an existing nodeJS library to provide the data, therefore it made sense to start with a clean slate implementation;

- secondly, having the register data in a separate library allows us to use the implementation in places where dmutils are not already present, or pin different apps to a specific version of the register data.

[`govuk-country-register`]: https://pypi.org/project/govuk-country-register

### DIFF
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -1,4 +1,6 @@
 from __future__ import unicode_literals
+
+import govuk_country_register
 import re
 from jinja2 import evalcontextfilter, Markup, escape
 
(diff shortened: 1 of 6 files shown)


dmutils/filters.py
```diff
@@ -96,3 +98,24 @@ def preserve_line_breaks(eval_ctx, value):
         result = Markup(result)
 
     return result
+
+
+def sub_country_codes(text):
+    """Replace country codes with country common name
+
+    :param text:  text containing country code in the form 'country:TLD'
+    :return:  text containg country names, or TLDs if not found
```
katstevens: A small nitpick but it's not the TLD - it's the ISO alpha-2 code (see https://www.gov.uk/government/publications/iso-country-codes--2).

lfdebrux:
```suggestion
        :param text:    text containing country codes in the form 'country:AA'
                    where AA is a two-letter ISO 3166-2 alpha-2 country code
        :return:        text containing country names, or the country code if not found
```

PR #494, opened 2019-02-20
get_app_status: log check failures

risicle:
See https://trello.com/c/CqTPp8E0/

I added this in two parts. Firstly, the sensible log message output just before the view returns - put here because it's not _just_ the "additional checks" we want to report the results of.

But Then I _also_ added a `logged_duration` wrapper around the call to each additional check, with the idea that it would give us better detail about the timing and order of these failures. It may aid in untangling things during a storm.

I've done it as two separate commits, so I can easily drop the latter. Do we think we want the latter or is it too much?

### DIFF
--- a/dmutils/status.py
+++ b/dmutils/status.py
@@ -1,6 +1,10 @@
 import math
 import os
+import sys
 from flask import jsonify, current_app
+
+
+from dmutils.timing import logged_duration
 
 
 def get_version_label(path):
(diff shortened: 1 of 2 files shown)


dmutils/status.py
```diff
@@ -39,7 +43,18 @@ def __init__(self, message):
 def _perform_additional_checks(additional_checks, response_data, error_messages):
     for additional_check in additional_checks:
```
katstevens: When do we actually carry out `additional_checks`? Is it just the APIs?

risicle: Good question. I just went and checked - it is indeed just the APIs ATM.

risicle: (search api checks es, api checks pg, av api checks clamd)

lfdebrux:
I think having the logged_duration around additional checks is a good idea

PR #498, opened 2019-03-07
Pin boto3 requirement to range

lfdebrux:
Requiring a specific version of boto3 causes issues with version conflicts

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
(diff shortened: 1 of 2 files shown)


katstevens:
Can we check this works ok with moto - this caused us problems before https://github.com/alphagov/digitalmarketplace-utils/pull/452

lfdebrux:
> Can we check this works ok with moto - this caused us problems before #452

Good point.

I haven't had any issues with `moto` in this repo, will check against the other repos now.

lfdebrux:
Should be fine as long as we're using `moto3<=1.3.7`

PR #507, opened 2019-03-21
Handle users who want to resubscribe to Mailchimp lists

katstevens:
Trello: https://trello.com/c/pzAoSoTC/434-mailchimp-resubscribe-for-users-who-have-unsubscribed

Users who try to sign up when they've already unsubscribed trigger a 503 error - not a great experience. 

Solution based on Ben's idea in the ticket: 
- new `DMMailChimpClient` method `resubscribe_email_to_list` which sends a confirmation email to the user first (i.e. no error message needs to be shown on the frontend)
- catch this type of error in the `subscribe_new_email_to_list` method, log a warning and try the resubscribe, and don't raise the error.

### DIFF
--- a/dmutils/email/dm_mailchimp.py
+++ b/dmutils/email/dm_mailchimp.py
@@ -142,6 +142,16 @@ class DMMailChimpClient(object):
                     extra={"error": str(e), "mailchimp_response": response}
                 )
                 return True
+            elif 'The contact must re-subscribe to get back on the list.' in response.get('detail', ''):
+                # User has previously unsubscribed and must confirm their resubscription via email
+                self.logger.warning(
+                    f"Expected error: Mailchimp cannot automatically subscribe user ({hashed_email}) to list "
+                    "({list_id}) as the user has been permanently deleted and must manually re-subscribe. "
+                    "A confirmation email will be sent.",
+                    extra={"error": str(e), "mailchimp_response": response}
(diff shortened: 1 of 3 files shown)


dmutils/email/dm_mailchimp.py
```diff
@@ -157,6 +167,27 @@ def subscribe_new_emails_to_list(self, list_id: str, email_addresses: str) -> bo
                     success = False
         return success
 
+    def resubscribe_email_to_list(self, list_id: str, email_address: str):
+        """Attempt to resubscribe a deleted user (a confirmation email is sent to them)"""
```
lfdebrux:
```suggestion
        """Attempt to resubscribe a deleted user (a confirmation email is sent to them by Mailchimp)"""
```

benvand:
It might be worth trying out that api 'update pending'  command by running up the dmmailchimpclient locally with a prod key and using one of our emails.
As far as I could tell from the docs this PR should work, but I'm not 100%

katstevens:
Hmm, interesting. I subscribed myself to a test list, deleted myself in the Mailchimp console, then was able to resubscribe.... with no confirmation at all?

katstevens:
This solution won't allow resubscription of deleted users - we'll need return a different error so we can display an appropriate message to the user (i.e. 'use a different email').

PR #510, opened 2019-03-26
435-spike-figure-out-why-werkzeug-0150-is-breaking-everything

benvand:
https://trello.com/c/DvUSNTbL/435-spike-figure-out-why-werkzeug-0150-is-breaking-everything

New ProxyFix class is requires kwargs to have an effect.

https://github.com/pallets/werkzeug/blob/0.14.1/werkzeug/contrib/fixers.py#L97
vs
https://github.com/pallets/werkzeug/blob/master/src/werkzeug/middleware/proxy_fix.py#L46 

It won't set the HTTP_X_FORWARDED_PROTO unless you pass in a value for `x_proto` when you instantiate it.

Do this.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '48.0.1'
+__version__ = '48.0.2'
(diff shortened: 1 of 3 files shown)


katstevens:
How do our apps fare if one is on `Werkzeug==0.15.x` and another is on `0.14.x`? Is this another `itsdangerous` multi-bump waiting to happen?

benvand:
The causes the functionality to stay the same. Other apps won't see any difference between the requests they're receiving

PR #514, opened 2019-04-10
Mailchimp invalid fix

katstevens:
Fix for the changes in Mailchimp's error response for invalid emails. The method now catches generic Mailchimp 400s (i.e. the user has made a mistake, not Mailchimp).

Previous response message: `email address looks fake or invalid`
Now: `Please provide a valid email address`

See recent log message example:
```
{'type': 'http://developer.mailchimp.com/documentation/mailchimp/guides/error-glossary/', 'title': 'Invalid Resource', 'status': 400, 'detail': 'Please provide a valid email address.', 'instance': '<instance-id-redacted>'}
```

Also 
- renamed some tests that were still referring to the old return values for errors
- changed the expected error case to log level `WARNING` rather than `ERROR`

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '48.0.2'
+__version__ = '48.0.3'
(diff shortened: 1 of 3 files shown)


katstevens:
@benvand yeah I was deliberating on that 😹 Decided on 'fix' as that part of the current major bump is genuinely broken...

PR #521, opened 2019-05-13
logged_duration: wrap get_disk_space_status, give default condition a "slow call threshold"

risicle:
The first of these I don't expect to be controversial, because we already similarly wrap all the "additional checks" for the status view.

The second adds a `SLOW_DEFAULT_CALL_THRESHOLD` of 0.5s to the default `logged_duration` condition. This will have a small effect on existing uses of it, but I don't think it's significant enough to warrant a major bump. The commit message explains more:

> the thinking here is that *any* routine we consider measurable spending more than half a second doing what it's doing is probably something we want to know about and the maximum rate it can hypothetically churn our unwanted logs is 2 per second (per thread I guess). longer than SLOW_EXTERNAL_CALL_THRESHOLD because we already try and limit the number of external calls we do, whereas we may not have the same attitude to whatever block of code this is wrapped around.

> a logged_duration user wanting to use the default condition but a shorter threshold can always add that to the default condition with an "or" clause


katstevens:
How much (roughly) do you expect the number of logged calls to increase by? Doubled? 10x?

risicle:
I'd expect it only to make a difference in weird cases.

PR #532, opened 2019-06-24
s3: add get_full_signed_url method

risicle:
This doesn't have any tests yet as I'm just floating it for general approval. If we're going to let admins delete comms files we really have to allow them to download them to see what they're about to delete. So this implies we need some kind of clone of https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/eca2133f5a9cd3df95ec41e539d930e1816cdf2b/app/main/views/frameworks.py#L701

So I'm just seeking to combine common code and it feels less silly to just add this method straight to the `S3` class instead of moving a copy of https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/eca2133f5a9cd3df95ec41e539d930e1816cdf2b/app/main/helpers/services.py#L50 to utils.

### DIFF
--- a/dmutils/s3.py
+++ b/dmutils/s3.py
@@ -6,10 +6,13 @@ from dateutil.parser import parse as parse_time
 import logging
 import mimetypes
 import os
+import urllib.parse as urlparse
 
 # a bit of a lie here - retains compatibility with consumers that were importing boto2's S3ResponseError from here. this
 # is the exception boto3 raises in (mostly) the same situations.
 from botocore.exceptions import ClientError as S3ResponseError
+
+from flask import has_app_context, current_app
 
 from .formats import DATETIME_FORMAT
 from .timing import logged_duration_for_external_request as log_external_request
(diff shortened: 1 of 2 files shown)


dmutils/s3.py
```diff
@@ -138,6 +141,25 @@ def get_signed_url(self, path, expires_in=30):
...
+            return None
+
+        if assets_url is None:
+            if not has_app_context():
+                raise TypeError("assets_url argument must be provided if no app context is available")
+            assets_url = current_app.config["DM_ASSETS_URL"]
+
+        parsed_url = urlparse.urlparse(url)
+        parsed_assets_url = urlparse.urlparse(assets_url)
+        return parsed_url._replace(netloc=parsed_assets_url.netloc, scheme=parsed_assets_url.scheme).geturl()
```
katstevens: Does this duplicate what we have in `dmutils.documents.get_signed_url`? Or is that what the `self.get_signed_url` is using?

katstevens: As far as I can see, the other views in the Admin that fetch S3 docs use `dmutils.documents.get_signed_url`

risicle: oh man i didn't even know about that...

risicle: 99% yes it does - seems the supplier FE doesn't know about that function either.

PR #534, opened 2019-07-24
documents: remove asserts in favour of explicit checks

risicle:
Just a quick one I noticed.

`assert`s are meant as debugging hints, not value enforcement - they should *not* be relied upon in live code. The python runtime is is hypothetically free to omit them.

### DIFF
--- a/dmutils/documents.py
+++ b/dmutils/documents.py
@@ -81,7 +81,8 @@ def upload_document(uploader, upload_type, documents_url, service, field, file_c
              failed
 
     """
-    assert upload_type in ['documents', 'submissions']
+    if upload_type not in ('documents', 'submissions',):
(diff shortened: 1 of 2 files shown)


dmutils/documents.py
```diff
@@ -81,7 +81,8 @@ def upload_document(uploader, upload_type, documents_url, service, field, file_c
              failed
 
     """
-    assert upload_type in ['documents', 'submissions']
+    if upload_type not in ('documents', 'submissions',):
+        raise ValueError(f"Unexpected upload_type {upload_type!r}")
```
katstevens: Oooh fancy `repr`!

risicle:
Mmmmmmnah I wasn't going to bother pulling it into apps specifically

PR #536, opened 2019-08-22
Add "cookie probe" system to allow us to properly detect disabled cookies

risicle:
https://trello.com/c/HsZ9mj7o

Called this "cookie probe" instead of "test cookie" because we already overload the term "test" enough.

This consists of two parts: a piece of middleware to set a long-lived "probe" cookie on every response and a check for this which ends up being in the `csrf_handler` for reasons explained in the comments.

The latter check needs to be enabled using the config setting `DM_COOKIE_PROBE_EXPECT_PRESENT`, but only *after* we've rolled out the cookie-setting half to all frontends, so we can reliably expect the cookie to be present.

### DIFF
--- a/dmutils/cookie_probe.py
+++ b/dmutils/cookie_probe.py
@@ -1,0 +1,29 @@
+from datetime import timedelta
+
+
+
+DEFAULT_DM_COOKIE_PROBE_COOKIE_NAME = "dm_cookie_probe"
+DEFAULT_DM_COOKIE_PROBE_COOKIE_VALUE = "yum"
+DEFAULT_DM_COOKIE_PROBE_COOKIE_MAX_AGE = timedelta(days=365).total_seconds()
+DEFAULT_DM_COOKIE_PROBE_EXPECT_PRESENT = False
+
+
+def init_app(app):
+    """
+        Unconditionally sets a long-lived cookie on all responses from `app`, which we will
+        easily be able to check for when trying to determine whether cookies are working for a
+        particular client at all.
+    """
+    app.config.setdefault("DM_COOKIE_PROBE_COOKIE_NAME", DEFAULT_DM_COOKIE_PROBE_COOKIE_NAME)
+    app.config.setdefault("DM_COOKIE_PROBE_COOKIE_VALUE", DEFAULT_DM_COOKIE_PROBE_COOKIE_VALUE)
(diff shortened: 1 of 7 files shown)


dmutils/errors/frontend.py
```diff
@@ -10,7 +12,26 @@ def csrf_handler(csrf_error):
...
+        ) != current_app.config.get("DM_COOKIE_PROBE_COOKIE_VALUE", cookie_probe.DEFAULT_DM_COOKIE_PROBE_COOKIE_VALUE)
+    ):
+        current_app.logger.info(
+            "cookie_probe.failed: cookies probably not working for this user"
+        )
+        return render_error_page(
+            status_code=400,
+            error_message="This feature requires cookies to be enabled for correct operation",
+        )
+    elif 'user_id' not in session:
```
lfdebrux: Rather than overloading the csrf handler completely can we break this out into a seperate function and then include that function in csrf handler based on the presence of the config value?

dmutils/cookie_probe.py
```diff
@@ -0,0 +1,28 @@
+from datetime import timedelta
+
+
+DEFAULT_DM_COOKIE_PROBE_COOKIE_NAME = "DM_COOKIE_PROBE"
+DEFAULT_DM_COOKIE_PROBE_COOKIE_VALUE = "yum ouch"
```
lfdebrux: Not sure about this...

dmutils/cookie_probe.py
```diff
@@ -0,0 +1,28 @@
+from datetime import timedelta
+
+
+DEFAULT_DM_COOKIE_PROBE_COOKIE_NAME = "DM_COOKIE_PROBE"
```
lfdebrux: Typically cookie names are lowercase?

dmutils/errors/frontend.py
risicle: Ahhi was on the verge of doing that yes, just couldn't decide what the call semantics should be... yeah I'll at least put the condition in a separate function.

dmutils/cookie_probe.py
risicle: "cookie" -> "yum"
"probe" -> "ouch"

Might just make it "yum"

dmutils/cookie_probe.py
risicle: orly? ok then

risicle:
I've actually tested this now and it does indeed work.

PR #537, opened 2019-09-26
add DMGzipMiddleware, a thin wrapper around Flask_gzip adding header-controllability & instrumentation

risicle:
https://trello.com/c/Kd4qas51

This should make us (me) more comfortable about rolling out compression to other apps.

Bumped the default min size to 8k to ensure it really is only used when needed.

### DIFF
--- a/dmutils/flask.py
+++ b/dmutils/flask.py
@@ -1,6 +1,7 @@
 from functools import partial
 
 from flask import render_template, render_template_string
+from flask_gzip import Gzip
 
 from dmutils.timing import logged_duration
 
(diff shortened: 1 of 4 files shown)


dmutils/flask.py
```diff
@@ -32,3 +33,24 @@
 timed_render_template_string.__doc__ = """
     See ``timed_render_template``, only for ``render_template_string``.
 """
+
+
+class DMGzipMiddleware(Gzip):
+    compress_by_default: bool
+
+    def __init__(self, *args, compress_by_default: bool = False, **kwargs):
```
lfdebrux: I think setting a default parameter has the same effect as using kwargs.setdefault (and is more readable).

```suggestion
    def __init__(self, *args, compress_by_default: bool = False, minimum_size = 8192, **kwargs):
```

dmutils/flask.py
```diff
@@ -32,3 +33,24 @@
...
+        kwargs.setdefault("minimum_size", 8192)
+        self.compress_by_default = compress_by_default
+        super().__init__(*args, **kwargs)
+
+    def after_request(self, response):
+        x_compression_safe = response.headers.pop("X-Compression-Safe", None)
+        compress = {"0": False, "1": True}.get(x_compression_safe, self.compress_by_default)
```
lfdebrux: Why 0/1 instead of false/true?

risicle: You don't tend to see false/true so much in header values, and then you also have all the uncertainty of `false`/`true`/`False`/`True`/`f`/`t`/`F`/`T`... when does it stop?

dmutils/flask.py
risicle: But then it opens the possibility of forgetting to add it back to the kwargs in the supercall...

PR #538, opened 2019-09-27
DMGzipMiddleware: emit warning when middleware receives an unknown X-Compression-Safe value

risicle:
@lfdebrux had a point in https://github.com/alphagov/digitalmarketplace-utils/pull/537 that the choice of `0`, `1` as `X-Compression-Safe`'s values may seem funny to some - I still think it's the best choice but in case anyone misses the plot let's at least have a warning so developers know when an attempted use isn't having the effect they expect it to.

### DIFF
--- a/dmutils/flask.py
+++ b/dmutils/flask.py
@@ -1,4 +1,5 @@
 from functools import partial
+import warnings
 
 from flask import render_template, render_template_string
 from flask_gzip import Gzip
@@ -45,7 +46,14 @@ class DMGzipMiddleware(Gzip):
 
     def after_request(self, response):
         x_compression_safe = response.headers.pop("X-Compression-Safe", None)
-        compress = {"0": False, "1": True}.get(x_compression_safe, self.compress_by_default)
+        known_values = {"0": False, "1": True}
+        compress = known_values.get(x_compression_safe, self.compress_by_default)
+
+        if x_compression_safe not in known_values and x_compression_safe != "" and x_compression_safe is not None:
+            warnings.warn(
+                f"{self.__class__.__name__} received unknown X-Compression-Safe header value {x_compression_safe!r} - "
+                "falling back to default"
+            )
 
         # flask_gzip makes the minimum_size comparison itself, but we want to avoid outputting a misleading
         # logged_duration message if it's going to be prevented in the superclass.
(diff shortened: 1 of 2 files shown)


dmutils/flask.py
```diff
@@ -45,7 +46,14 @@ def __init__(self, *args, compress_by_default: bool = False, **kwargs):
...
         x_compression_safe = response.headers.pop("X-Compression-Safe", None)
-        compress = {"0": False, "1": True}.get(x_compression_safe, self.compress_by_default)
+        known_values = {"0": False, "1": True}
+        compress = known_values.get(x_compression_safe, self.compress_by_default)
+
+        if x_compression_safe not in known_values and x_compression_safe != "" and x_compression_safe is not None:
+            warnings.warn(
+                f"{self.__class__.__name__} received unknown X-Compression-Safe header value {x_compression_safe!r} - "
+                "falling back to default"
+            )
```
katstevens: Do we use the built in `warnings` anywhere else? How do these look when fed through an app logger?

risicle: I don't think they come out at all in logs - I'm expecting people to notice/get annoyed by them being shown after a pytest run.

dmutils/flask.py
```diff
@@ -45,7 +46,14 @@ def __init__(self, *args, compress_by_default: bool = False, **kwargs):
 
     def after_request(self, response):
         x_compression_safe = response.headers.pop("X-Compression-Safe", None)
-        compress = {"0": False, "1": True}.get(x_compression_safe, self.compress_by_default)
+        known_values = {"0": False, "1": True}
+        compress = known_values.get(x_compression_safe, self.compress_by_default)
+
+        if x_compression_safe not in known_values and x_compression_safe != "" and x_compression_safe is not None:
```
lfdebrux: This conditional could be simplified to

```suggestion
        if x_compression_safe and x_compression_safe not in known_values:
```

lfdebrux: Oh, also, I didn't think `not in` worked in Python? 🤔

risicle: > This conditional could be simplified to

_Nearly_. Your line would not warn on actual python `False`, and I didn't go and check whether flask _always_ stringifies header values between view and middleware and don't really want to make the reader have to go & check up on that.

PR #542, opened 2019-11-12
bump werkzeug restriction to >=0.16,<0.17

risicle:
https://trello.com/c/dz3EnMnh

What do people think of me adding this `python-dateutil` restriction here? It's demanded by `botocore 1.13.15`, but of course _could_ be mandated in-app instead...?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Records breaking changes from major version bumps
 
+## 50.0.0
+
+PR [#542](https://github.com/alphagov/digitalmarketplace-scripts/pull/542)
+
+Bumping the werkzeug version restriction. Test & deploy with extra care.
+
 ## 49.0.0
 
 PR [#540](https://github.com/alphagov/digitalmarketplace-scripts/pull/540)
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '49.0.0'
+__version__ = '50.0.0'
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
          'python-json-logger<0.2,>=0.1.4',
          'pytz',
          'unicodecsv>=0.14.1',
-         'Werkzeug>=0.15.1,<0.16.0',
+         'Werkzeug>=0.16,<0.17',
          'workdays>=1.4',
     ],
     python_requires="==3.6.*",


lfdebrux:
> What do people think of me adding this python-dateutil restriction here? It's demanded by botocore 1.13.15, but of course could be mandated in-app instead...?

Not a fan; unless we have a reason to restrict what version of dateutil we use then we should let botocore's restriction take precedence

risicle:
> we should let botocore's restriction take precedence

Oh it's just that it doesn't. Instead it happily installs 2.8.1 and then complains.

lfdebrux:
> > we should let botocore's restriction take precedence
> 
> Oh it's just that it doesn't. Instead it happily installs 2.8.1 and then complains.

I wasn't able to reproduce this, what were the steps you followed?

risicle:
Clear out and rebuild your test venv for this repo.

If it doesn't work for you, I wouldn't be too surprised - pip is not a reproducible build system.

risicle:
`botocore 1.13.15 has requirement python-dateutil<2.8.1,>=2.1; python_version >= "2.7", but you'll have python-dateutil 2.8.1 which is incompatible.` is what it tells me without this restriction.

lfdebrux:
I got the same error only when doing `pip install -r requirements-dev.txt`; `pip install -r requirements.txt` resulted in installation of `python-dateutil==2.8.0` as expected.

risicle:
Rrrrright, but either way, it's flaky. And it could be considered a problem that we're testing a different version from what gets finally installed.

katstevens:
I get a bunch of similar warnings on master! I think sorting out utils' `requirements-dev` so that we have a stable test environment (that reflects the reality of it being used in an app) is probably a separate ticket...

dmutils/__init__.py
```diff
@@ -2,4 +2,4 @@
 from .flask_init import init_app, init_manager
 
 
-__version__ = '49.0.0'
+__version__ = '50.0.0'
```
katstevens: What should go in the CHANGELOG entry for this?

risicle:
Ah well shall we just leave it giving the same error as before then?

dmutils/__init__.py
risicle: "Look out!"

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 Records breaking changes from major version bumps
 
+## 50.0.0
+
+PR [#542](https://github.com/alphagov/digitalmarketplace-scripts/pull/542)
+
+Bumping the werkzeug and itsdangerous version restrictions. Test & deploy with extra care.
```
katstevens: Ooo are we going to update the pinned `itsdangerous==0.24` on every app then? Fun!

risicle: Oyes - I don't think there's any point in doing the werkzeug bump without it.

katstevens: ... I am glad this is a major bump.

katstevens: OK, looking through the `werkzeug` changelog I don't think we need the `itsdangerous` bump yet. However there are breaking changes to how `werkzeug` imports are used: https://werkzeug.palletsprojects.com/en/0.16.x/changes/

I think we're mostly OK but the AV API does a direct import, which will need changing.

risicle: > but the AV API does a direct import

I actually think what's being used in that case (`test_sns.py`, right?) doesn't count as a "top level import" because it's actually accessing a submodule of the top-level. "top level" I think means the uncategorized melange of objects which packages will often put at the top level for convenience.

lfdebrux:
```suggestion
Bumping the werkzeug version restrictions. Test & deploy with extra care.
```

PR #543, opened 2019-11-18
Add support for GOV.UK Frontend error components

aliuk2012:
Ticket: https://trello.com/c/jy1qTEQN/132-replace-validation-banner-with-govuk-frontend-error-summary-component-in-user-frontend

We want to be able to use the GOV.UK Frontend error summary component in our frontend apps.

This commit adds the expected parameters for the `govukErrorSummary` component to the output of `get_errors_from_wtform`, so that you can create an error summary for a WTForms form using

    # view.py
    errors = get_errors_from_wtform(form)

    # template.html
    {{ govukErrorSummary({
      "errorList": errors.values()
    }) }}

    {{ govukTextInput({
      "errorMessage": errors.input.message
    }) }}

This change should be backwards compatible.

---

Note that the GOV.UK Frontend versions use the error message to populate the error summary, which is different to the frontend toolkit style; see https://github.com/alphagov/digitalmarketplace-user-frontend/pull/190 for an example of these changes in use.

### DIFF
--- a/dmutils/forms/helpers.py
+++ b/dmutils/forms/helpers.py
@@ -19,13 +19,71 @@ def remove_csrf_token(data):
 
 
 def get_errors_from_wtform(form):
-    """Converts errors from a Flask-WTForm into the same format we generate from content-loader forms. This allows us
-    to treat errors from both content-loader forms and wtforms the same way inside templates.
+    """Converts Flask-WTForm errors into formats suitable for use in Digital Marketplace templates.
+
+    Returns a dictionary with the following keys, supporting three types of error display:
+
+    1) digitalmarketplace-frontend-toolkit template toolkit/forms/validation.html:
+      `input_name`, `question`, `message`
+
+    2) govuk-frontend macro govukErrorSummary:
+      `text`, `href`
+
+    3) govuk-frontend errorMessage:
+      `errorMessage`
+
+    This allows us to treat errors from both content-loader forms and wtforms the same way inside templates,
+    and provides backwards-compatible support for the migration to using govuk-frontend components.
+
+    Example usage:
+
+        # app.py
+        class Form(FlaskForm):
+            input = DMTextInput()
+
+        @flask.route("/")
+        def view():
+            form = Form()
+            errors = get_errors_from_wtform()
+            return render(
+                "template.html",
+                errors=errors,
+            )
+
+        # template.html  (govuk-frontend)
+        {{ govukErrorSummary({
+            "errorList": errors.values()
+        }) }}
+
+        {{ govukTextInput({
+            "errorMessage": errors.input.errorMessage,
+        }) }}
+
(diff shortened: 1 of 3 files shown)


lfdebrux:
I've tested this version with user frontend and no other changes and it passed tests/didn't have any visual changes.

tests/test_forms.py
```diff
@@ -202,14 +202,44 @@ def test_errors_are_formatted(self, app, email_address):
...
+                formdata=ImmutableMultiDict((
+                    ("test_email", email_address,),
+                )),
+                meta={'csrf': False},
+            )
+
+            assert form.validate() is False
+            assert form.errors
+
+            errorMessage = get_errors_from_wtform(form)["test_email"]["errorMessage"]
```
katstevens: Can we stick to snake_case in Python tests please 😸

dmutils/forms/helpers.py
```diff
@@ -19,13 +19,61 @@ def remove_csrf_token(data):
...
+    """Converts errors from a Flask-WTForm into the same format we generate from content-loader forms.
+
+    Returns a dictionary that includes three keys: `input_name`, `question`,
+    and `message`. The dictionary should be passed to the template as `errors`.
+
+    This allows us to treat errors from both content-loader forms and wtforms
+    the same way inside templates.
+
+    We also include in the dictionary errors in a format suitable for GOV.UK
+    Frontend components:
```
katstevens: I think we can rewrite this docstring to be less confusing. How about:

```
Converts Flask-WTForm errors into formats suitable for use in Digital Marketplace templates.

Returns a dictionary with the following keys, supporting three types of error display:

1) digitalmarketplace-frontend-toolkit template toolkit/forms/validation.html:
  `input_name`, `question`, `message`

2) govuk-frontend macro govukErrorSummary:
  `text`, `href`

3) govuk-frontend errorMessage:
  `errorMessage`

This allows us to treat errors from both content-loader forms and wtforms the same way inside templates, 
and provides backwards-compatible support for the migration to using govuk-frontend components.

Example usage:

        # app.py
        class Form(FlaskForm):
            input = DMTextInput()

        @flask.route("/")
        def view():
            form = Form()
            errors = get_errors_from_wtform()
            return render(
                "template.html",
                errors=errors,
            )

        # template.html  (govuk-frontend)
        {{ govukErrorSummary({
            "errorList": errors.values()
        }) }}

        {{ govukTextInput({
            "errorMessage": errors.input.errorMessage,
        }) }}

        # template.html (DM frontend toolkit uses 'errors' by default)
        {% include 'toolkit/forms/validation.html' %}        
```

lfdebrux: Nice idea 👍

lfdebrux: Thanks for writing that!

PR #546, opened 2019-12-03
Template filter: open links in new tab

katstevens:
https://trello.com/c/WSCRVNH8/149-allow-buyers-to-preview-external-links-on-their-brief

For the DOS preview feature, we want to allow buyers to check whether links in their brief content work OK. 

We already use a template filter `format_links` which picks up plain text URIs like `http:///example.com` in user-generated content, and converts them to anchor links `<a href="http://example.com" class="break-link" rel="external">http://example.com</a>`. However in the DOS preview, these links would currently open within the iframe, and are unlikely to display correctly (e.g if the site being linked to has disabled the X-Frame-Options header - we can't control this). Therefore we want the links to open in a new tab.

~This new filter is designed to wrap the `format_links` filter (which it is largely copied from!) and place a `target="_blank"` into any anchor links.  It assumes that a string-like object will be passed to it, and that any escaping has already been handled by `format_links`.~

~Expected usage:~
```
{{ item.value | format_links | open_links_in_new_tab }}
```
~(tested this in Briefs FE, works as expected)~

~The filter will skip links that already have `target="_blank"`, however the code to do this is pretty ugly. If anyone has a fancy regex suggestion to filter out that pattern from the start, please shout!~

**EDIT** Changed approach thanks to @risicle's suggestion, adding a parameter to the existing `format_links` filter. I kept the test class reorganisation though 😸  

New usage:

```
{{ item.value | format_links(open_links_in_new_tab=True) }}
```

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '50.1.0'
+__version__ = '50.2.0'
--- a/dmutils/filters.py
+++ b/dmutils/filters.py
@@ -15,7 +15,7 @@ def smartjoin(input):
         return ''
 
 
-def format_links(text):
+def format_links(text, open_links_in_new_tab=None):
     """
     Filter that searches a given string (or other string-like object) for any URIs
     and wraps them with either an anchor link or a span, depending on whether the link contains a valid protocol.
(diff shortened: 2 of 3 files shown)


lfdebrux:
Is the plan then to add this filter just for the preview, or will all links in briefs open in new tabs in future?

katstevens:
Just the preview for now. If there's a user need to open links in a new tab elsewhere, people can make the decision to add the filter on a case-by-case basis.

risicle:
Ohhh processing html/xml with regexes is baaaaaad mojo.

risicle:
Wouldn't it just make more sense to give the `format_links` filter a `new_tab` argument?

katstevens:
@risicle We'd need to change it to a context processor then - template filters don't get on well with variables.

risicle:
https://stackoverflow.com/a/48532564 ?

katstevens:
Well whaddayaknow 👍 

We only use the `format_links` filter in a couple of places - I'll test those and make sure the change doesn't have any weird side effects.

lfdebrux:
Ah, I just realised we should probably add `rel="noreferrer noopener"` as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Security_and_privacy

PR #549, opened 2019-12-06
timing tests: remove testing of timing.exceeds_slow_external_call_threshold

risicle:
It's technically no more interesting than other time threshold cases we already test and this reduces the strength of the dimensional explosion of the parametrization, speeding up the test run, which is welcome.


katstevens:
master `3521 passed in 138.59 seconds`
branch `3233 passed in 114.63 seconds`

Is this roughly what was expected?

risicle:
Yes.

risicle:
Nope

PR #550, opened 2019-12-09
168-spike-check-a-newly-deployed-app-can-connect-to-its-paas-backing-service-timebox-1d

benvand:
See changelog, change dmutils/status.py::get_app_status to always run additional_checks

https://trello.com/c/B2GfZgfO/192-cause-additionalchecks-to-always-be-run-regardless-of-ignore-dependencies-on-status-endpoint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 Records breaking changes from major version bumps
 
+## 51.0.0
+
+Signature change to:
+
+`dmutils/status.py::get_app_status`
+
+```
+get_app_status(
+     search_api_client=None,
+     ignore_dependencies=False,
+     additional_checks=None,
+-    additional_checks_internal=None,
++    additional_checks_extended=None,
+ ):
+```
+
+Additional checks will now always be called regardless of the value of
+`ignore_dependencies`.
+
+New argument `additional_checks_extended` whose checks will only be
+called when `ignore_dependencies=False`.
+
 ## 50.0.0
 
 PR [#542](https://github.com/alphagov/digitalmarketplace-scripts/pull/542)
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '50.4.0'
+__version__ = '51.0.0'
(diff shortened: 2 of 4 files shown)


risicle:
So this is really just a renaming/switcharound of the arguments to `get_app_status`?

lfdebrux:
So how does this work in practice? What's the url?

benvand:
@lfdebrux @risicle 

It means that the methods passed as `additional_checks` are always run. Even if `ignore-dependencies = True` is passed.

It adds a new argument `additional_checks_extended` which are only run when `ignore-dependencies = False`.

The APIs are the only apps that use the `additional_checks` argument and the av-api is the only app that will use the `additional_checks_extended` arg so far.

Because we decided that we should ping es and psql (and by extension the clamav process) when we get the api statuses, even if we are ignoring dependencies like the search-api client, we can just change the way the status works to include those dependencies by default leveraging the existing arguments.

It only affects the API health checks; as they;re the only ones that include `additional_checks` and the `/status?ignore-dependencies` on APIs is only hit by the healthchecks from PAAS, rather than pingdom too which can't reach the apis.

benvand:
@risicle Yeah effectively
You can see here, swapping what is passed to `additional_checks` and `additional_checks_internal` / `additional_checks_extended` https://github.com/alphagov/digitalmarketplace-antivirus-api/pull/59/files

benvand:
@lfdebrux The url will be `/status?ignore-dependencies`

lfdebrux:
I find this structure of dependencies, additional checks and extended additional checks a bit confusing; to me it's not clear on the face of it what gets run in what case, and even with the documentation I find it a bit mind-bending.

I feel like while we're making a breaking change we have the possibility of overhauling the call signature to something that's a bit clearer to use?

For instance, what about something like this?

```
def get_app_status(
        data_api_client,
        search_api_client,
        include:Optional[str] = "internal_checks,dependencies",
        *,
        dependencies:Optional[List[Callable]] = None,
        internal_checks:Optional[List[Callable]] = None,
        external_checks:Optional[List[Callable]] = None,
        **
    )
```

Then what is included by default is clear from the signature, you could have urls such as `/status?include=external_checks,dependencies`, `/status?include=none`, `/status?include=checks`, and there's scope to include other kinds of checks tests what-have yous in future.

PR #552, opened 2020-01-22
22-create-class-and-tests-for-class

benvand:
https://trello.com/c/QOPKvSd5/22-create-class-and-tests-for-class

* Add direct plus client to interface with D&B Direct Plus API
* Add tests for client
* Minor version bump for new feature
* Requirements
  * added requests to package requirements
  * added requests-mock to dev requirements

### DIFF
--- a/dmutils/direct_plus_client.py
+++ b/dmutils/direct_plus_client.py
@@ -3,6 +3,7 @@ import requests
 
 
 class DirectPlusClient(object):
+    """Client to interface with Dun and Bradstreet's Direct Plus API."""
 
     access_token = None
     protocol = 'https'
@@ -22,7 +23,7 @@ class DirectPlusClient(object):
         Set the access token and header authorisation parameter for requests to use.
         """
         basic_auth_string = requests.auth._basic_auth_str(self.username, self.password)
-        response = self._dnb_request(
+        response = self._direct_plus_request(
             'token',
             method='post',
             version='v2',
@@ -33,7 +34,7 @@ class DirectPlusClient(object):
         self.access_token = response.json().get('access_token')
         self.required_headers += (('Authorization', f'Bearer {self.access_token}'),)
 
-    def _dnb_request(
+    def _direct_plus_request(
         self,
         endpoint,
         method='get',
@@ -60,7 +61,7 @@ class DirectPlusClient(object):
             # If access token invalid (401) refresh access token manually
             # and retry request with allow_access_token_reset = False
             self._reset_access_token()
-            response = self._dnb_request(
+            response = self._direct_plus_request(
                 endpoint,
                 method=method,
                 version=version,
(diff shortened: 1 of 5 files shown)


katstevens:
Given this is a shiny new class, do we want to include `typing` on the methods?

dmutils/direct_plus_client.py
```diff
@@ -0,0 +1,85 @@
+import logging
+import requests
+
+
+class DirectPlusClient(object):
```
katstevens: We have a mixture of naming here - `DirectPlus` and `dnb`. Worth having a brief docstring explaining what they are and what this is for?

benvand: Have updated the references to direct_plus and added a docstring.

direct_plus is more specific as this isn't an interface to dun and bradstreet bu _only_ their direct plus api

benvand:
@katstevens 

I'm happy to do it if you really want. But I could probably use some pointers. And I'd really rather not use this piece of work as a learning experience because we're so pressed for time.

A couple other points:
* This was originally designed to live in its own package and hopefully at some point we can migrate it to its own package. I think that would be a good time to implement typing. https://github.com/alphagov/direct-plus-python-client
* We haven't got a consistent strategy to get our code typed yet.
* I've not done it before and I don't want this code to start throwing exceptions that I'm not expecting because it wasn't written with typing in mind
* Really want to get this out the door so I can start helping on frontend stuff

katstevens:
> * This was originally designed to live in its own package and hopefully at some point we can migrate it to its own package. I think that would be a good time to implement typing. https://github.com/alphagov/direct-plus-python-client
> * We haven't got a consistent strategy to get our code typed yet.
> * I've not done it before and I don't want this code to start throwing exceptions that I'm not expecting because it wasn't written with typing in mind
> * Really want to get this out the door so I can start helping on frontend stuff

Sounds reasonable. I think our informal strategy re: typing was to add it when introducing new classes/methods. Happy to leave it as a tech debt card for now?

benvand:
@katstevens 

> Happy to leave it as a tech debt card for now?
https://trello.com/c/D1vKNpM8/1230-add-typing-to-directplusclient

👍

dmutils/direct_plus_client.py
```diff
@@ -0,0 +1,86 @@
...
+                allow_access_token_reset=False
+            )
+        return response
+
+    def get_organization_by_duns_number(self, duns_number):
+        """
+        Request a supplier by duns number from the Direct Plus API
+        """
+        response = self._direct_plus_request(
+            f'data/duns/{duns_number}', payload={'productId': 'cmpelk', 'versionId': 'v2'}
```
katstevens: Where is `cmpelk` coming from? Worth hiving this out into a constant at the top of the file in case it changes?

benvand: https://directplus.documentation.dnb.com/html/resources/JSONSample_CMP.html#cmpelk

`CMPELK` is an internal D&B code that sort of references the format of the response.

They call it a product code. So we're getting an 'organization' object from Dun and Bradstreets 'CMPELK' product using their 'Direct Plus API' interface.

But it's specific to getting a supplier org from this endpoint rather than referencing the API as a whole.

katstevens: Makes sense, thanks. And now we have it documented in this PR 😸

PR #553, opened 2020-01-30
Add 'become a supplier' external route

katstevens:
https://trello.com/c/N85yhHQX/278-add-suppliers-supply-to-external-routes-in-dmutils

This Supplier FE URL is used in the Buyer FE, where it's currently hardcoded.

Also adds in a step to install dev dependencies before running `make test`, as it wasn't doing so. I think the only reason we didn't have it there already was to get round the `freeze-requirements` issue, which we're phasing out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ test-flake8: virtualenv
 	${VIRTUALENV_ROOT}/bin/flake8 .
 
 .PHONY: test-python
-test-python: virtualenv
+test-python: virtualenv requirements-dev
 	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
 
 .PHONY: show-environment
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '51.2.0'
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ test-flake8: virtualenv
 	${VIRTUALENV_ROOT}/bin/flake8 .
 
 .PHONY: test-python
-test-python: virtualenv
+test-python: virtualenv requirements-dev
```
lfdebrux: I'm personally not a fan of having `make test` install requirements, because I sometimes want to run `make test` a lot quickly and having it run pip every time is annoying... on the other hand, there's definitely an advantage to having make do all the stuff magically for you.

I'm guessing I'm saying that maybe we should talk as a team about what we want our Makefiles to be like? One things for sure, they are in need of a rewrite (they could be a lot more consistent, and probably simpler too.)

katstevens: The majority of our apps (and repos with Python tests) include the dev-requirements in the `make test` command, so this will bring it in line with those.

Agree that the Makefiles need a review!

lfdebrux: > The majority of our apps (and repos with Python tests) include the dev-requirements in the make test command, so this will bring it in line with those.

That's fair, this change makes sense then.

> Agree that the Makefiles need a review!

It's on my personal firebreak list 😆

PR #555, opened 2020-02-17
email tokens: rename *_SALT config variables to *_TOKEN_NS

risicle:
https://trello.com/c/AVadfnPO

These are not really salts, and calling them so is likely to confuse and raise alarm bells as they are not used in the secure manner a salt would need to be.

Continue to fall back to `*_SALT` with a `DeprecationWarning` for the time being to enable backward compatibility. We should probably remove this compatibility on the next major bump, but I'm not going to do an omnibump for this.

### DIFF
--- a/dmutils/email/tokens.py
+++ b/dmutils/email/tokens.py
@@ -1,6 +1,8 @@
 import base64
 import json
 import struct
+from warnings import warn
+
 
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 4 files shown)


dmutils/email/tokens.py
```diff
@@ -100,7 +102,11 @@ def decode_password_reset_token(token, data_api_client):
         decoded, token_timestamp = decode_token(
             token,
             current_app.config["SHARED_EMAIL_KEY"],
-            current_app.config["RESET_PASSWORD_SALT"],
+            (
+                current_app.config.get('RESET_PASSWORD_TOKEN_NS')
```
katstevens: I might be missing something here - what does `_NS` stand for? 'Not salt'?

risicle: Namespace.

PR #557, opened 2020-03-04
1504-make-loggeddurationforexternalrequestalways-log-on-error

benvand:
https://trello.com/c/pOVlqacc/1504-make-loggeddurationforexternalrequestalways-log-on-error

* Make logged_duration_for_external_request always log on exception
* Only add content type is it is not none
* Bugfix version bump to 51.5.1

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '51.5.0'
+__version__ = '51.5.1'
--- a/dmutils/timing.py
+++ b/dmutils/timing.py
@@ -116,15 +116,23 @@ def request_is_sampled(log_context):
     return has_request_context() and getattr(request, "is_sampled", False)
 
 
+def exception_in_stack():
+    """Return true if we are currently in the process of handling an exception, ie one has been caught in a try block.
+
+    https://docs.python.org/3/library/sys.html#sys.exc_info
+    """
+    return sys.exc_info()[0] is not None
+
+
 def different_message_for_success_or_error(success_message, error_message):
     """Can be passed into `logged_duration` as `message=different_message_for_success_or_error(x, y)` in order to
     generate different log messages depending on whether the block completed successfully or raised an exception."""
-    return lambda _: success_message if sys.exc_info()[0] is None else error_message
+    return lambda _: success_message if not exception_in_stack() else error_message
 
 
-def request_context_and_any_of_slow_call_or_sampled_request(log_context):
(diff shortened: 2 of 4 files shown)


dmutils/s3.py
```diff
@@ -59,6 +60,10 @@ def save(self, path, file_, acl='public-read', timestamp=None, download_filename
                 str(download_filename).encode("ascii", errors="ignore").decode(),
             )
 
+        content_type = self._get_mimetype(path)
+        if content_type is not None:
+            extra_kwargs['ContentType'] = content_type
+
```
risicle: Hold on why would we want to upload documents with no mimetype?

benvand: Why not? At this point they've passed fleep checks, but we're 'unable' to guess what the mimetype is

risicle: We should be able to guess from the combination of the file extension and the zip header. Not having a mime type means when we serve it back to the user it probably won't prompt them to open it in a word processor.

risicle: Also if anyone ever does add format validation to the bucket scanner it will need to know what to validate it _as_, having entries which are `None` will leave it puzzled.

risicle:
If I could make a minor feature request - would it be easy to make it log the md5 of the file it's about to try uploading? Why md5? Becuase S3's `etag` is the file's md5 and it would allow us to pair them up without having to download all the files...

risicle:
FWIW this is https://trello.com/c/QOosZF6E

benvand:
> would it be easy to make it log the md5 of the file it's about to try uploading

Bit worried it'll be a performance hit it we have to iterate over the data again :/
The fact that we're timing the action makes me think it's probably long running as is.

risicle:
It's only long running because it's going across the network - calculating the md5 of a 5M file should take milliseconds. What this would allow us to do is say for certain that the user was trying to upload the exact same file twice in a row, even if it failed.

dmutils/s3.py
katstevens: Is there any use/risk in having a human-recognisable value in there? `extra_kwargs['ContentType'] = content_type or 'undetectable'`

risicle: i ... think i'd quite like it to continue throwing some sort of error if it can't figure out a mimetype for it, now Ben's fixed the problem thru the docker image. Might be better if the error were more obvious or happened earlier.

benvand: @risicle Have removed this

benvand:
@risicle 
Mentioned in tech meeting but not going to parse the file for the MD5 because not sure if it's going to seek the end of the file.
Have created a tech debt ticket for this. 
https://trello.com/c/NkIj5R98/1297-make-loggeddurationforexternalrequest-log-file-md5

PR #558, opened 2020-03-10
formats: split out get_localized_datetime() function from _format_date()

risicle:
To support https://trello.com/c/RBJb5rfV

All these date processing functions are nice and all, but none of them currently allow you to recover a `datetime` object from an iso timestamp as commonly returned from the api.

`get_localized_datetime` returns a timezone-localized `datetime` object given an iso timestamp or another `datetime`.

### DIFF
--- a/dmutils/formats.py
+++ b/dmutils/formats.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
-import pytz
 import re
+from typing import Union
+
+import pytz
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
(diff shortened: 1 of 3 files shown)


dmutils/formats.py
```diff
@@ -181,20 +183,28 @@ def _use_gmt_timezone(date_string):
     return re.sub(r"UTC|BST", "GMT", date_string)
 
 
-def _format_date(value, default_value, fmt, localize=True):
-    if not value:
-        return default_value
-
+def get_datetime(value: Union[str, datetime], localize: bool = True) -> datetime:
```
katstevens: Could we call this `get_localized_datetime`?

risicle: Done

PR #561, opened 2020-05-11
Change break-link class name

domoscargin:
https://trello.com/c/bcU2s0Zs/33-2-links-in-summary-tables-eg-dos-opportunities-dont-wrap-in-safari-or-chrome

Changing the class name `break-link` to `app-break-link`.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '52.0.0'
+__version__ = '52.1.0'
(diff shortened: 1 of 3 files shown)


katstevens:
We'll need a version bump on this.

katstevens:
It's borderline whether this is a breaking change or not. We specify `.break-link` as a custom class in the User FE assets, but I don't think we actually use it there (there's no user-generated content requiring the `format_links` filter in the User FE). So we can probably get away with making this a minor bump?

domoscargin:
Yes - `.break-link` in User FE is unused. I've [submitted a PR](https://github.com/alphagov/digitalmarketplace-user-frontend/pull/260) to remove the `_text.scss` partial that contains it - since that file is no longer imported.

The only other thing that would be "broken" by this change is the bug on the buyer frontend, for which there's [another PR here](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1032).

PR #562, opened 2020-05-27
log instance index number

gidsg:
https://trello.com/c/PTdD5lEh/102-our-logs-show-paas-instance-guid-but-not-paas-instance-index

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -113,7 +113,7 @@ def configure_handler(handler, app, formatter):
     handler.addFilter(AppNameFilter(app.config['DM_APP_NAME']))
     handler.addFilter(RequestExtraContextFilter())
     handler.addFilter(AppStackLocationFilter("app_", app.root_path))
-    if app.config.get('DM_ENVIRONMENT') == 'production':
+    if os.environ.get('CF_INSTANCE_INDEX'):
         handler.addFilter(AppInstanceFilter())
 
     return handler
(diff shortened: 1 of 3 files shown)


dmutils/logging.py
```diff
@@ -113,6 +113,8 @@ def configure_handler(handler, app, formatter):
     handler.addFilter(AppNameFilter(app.config['DM_APP_NAME']))
     handler.addFilter(RequestExtraContextFilter())
     handler.addFilter(AppStackLocationFilter("app_", app.root_path))
+    if app.config.get('DM_ENVIRONMENT') == 'production':
+        handler.addFilter(AppInstanceFilter())
```
katstevens, line 117: Why are we restricting this to production only?

gidsg: Ahh, good point, I think we did this initially just so the tests would pass as this environment variable is not set. But we should run it on any PaaS environment, so I guess for should be for any value other than local?

katstevens: Are we unable to check for the var directly with `os.environ.get('...')` here then?

gidsg: good point @katstevens have updated to do it this way now.

tests/test_logging.py
```diff
@@ -12,8 +12,27 @@
...
+    assert AppStackLocationFilter in filter_classes
+    assert AppInstanceFilter not in filter_classes
+
+
+def test_configure_handler_has_instance_index_where_present(app, os_environ):
+    os_environ.update({'CF_INSTANCE_INDEX': '1'})
+    handler = mock.Mock()
+    configure_handler(handler, app, mock.Mock())
+    filter_classes = [x[0][0].__class__ for x in handler.addFilter.call_args_list]
+    assert AppInstanceFilter in filter_classes
```
katstevens, line 35: These tests are fine, but an even better assertion style would be

```
assert filter_classes == [
    AppNameFilter,
    RequestExtraContextFilter,
    AppStackLocationFilter,
    AppInstanceFilter
]
```

So for each test, it's explicit which classes are in there (and if we remove/add any classes in future, it's very obvious).

PR #564, opened 2020-06-03
Add ability to override href for govuk_error object

lfdebrux:
We want to be able to use the dmutils.forms.errors helpers everywhere we are using forms in our frontend apps. Sometimes our forms can be a little... weird. Having the ability to override the field input href is useful, especially for things like fields inside selection buttons.

This PR does this by allowing a developer to either a) specify the href for the error message in the error object passed to `govuk_errors` or b) setting the `id` of the form field in a WTForms form.

This PR also adds some logic to cover the default cases better, reducing the situations where we do need to override the href. Specifically it:

- adds prefix `input-` to id for WTForms fields
- adds suffix `-1` to id for WTForms selection button fields

See the commit messages for more details.

---

This PR is needed to make sure that error summaries work as expected in the following tickets:

https://trello.com/c/u5CoMrCP/19-1-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-buyer-frontend

https://trello.com/c/qUAR9Nm4/20-1-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-briefs-frontend

### DIFF
--- a/dmutils/forms/fields.py
+++ b/dmutils/forms/fields.py
@@ -46,6 +46,12 @@ __all__ = ['DMBooleanField', 'DMDecimalField', 'DMHiddenField', 'DMIntegerField'
 class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
     widget = DMCheckboxInput(hide_question=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # digitalmarketplace-frontend-toolkit suffixes the id with `-1`
+        if "id" not in kwargs:
+            self.id = self.id + "-1"
+
     @property
     def options(self):
         return [{"label": self.label.text, "value": self.data}]
--- a/dmutils/forms/fields.py
+++ b/dmutils/forms/fields.py
@@ -50,7 +50,7 @@ class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
         super().__init__(*args, **kwargs)
         # digitalmarketplace-frontend-toolkit suffixes the id with `-1`
         if "id" not in kwargs:
-            self.id = self.id + "-1"
+            self.href = self.href + "-1"
 
     @property
     def options(self):
--- a/dmutils/forms/mixins.py
+++ b/dmutils/forms/mixins.py
@@ -56,11 +56,13 @@ class DMFieldMixin:
     Derived classes which include this mixin should have a
     subclass of `wtforms.Field` in their base classes.
     '''
-    def __init__(self, label=None, validators=None, hint=None, question_advice=None, _id_prefix="input-", **kwargs):
+    def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
         super().__init__(label=label, validators=validators, **kwargs)
 
-        if "id" not in kwargs:
-            self.id = _id_prefix + self.id
+        if "id" in kwargs:
+            self.href = kwargs["id"]
+        else:
+            self.href = "input-" + self.id
 
         if hint:
             self.hint = hint
@@ -109,7 +111,8 @@ class DMSelectFieldMixin:
     def __init__(self, label=None, validators=None, coerce=text_type, options=None, **kwargs):
         super().__init__(label, validators=validators, coerce=coerce, **kwargs)
         if "id" not in kwargs:
-            self.id = self.id + "-1"
+            # This should be the id for the first option
+            self.href = self.href + "-1"
         if options:
             self.options = copy(options)
 
--- a/dmutils/forms/mixins.py
+++ b/dmutils/forms/mixins.py
@@ -56,8 +56,12 @@ class DMFieldMixin:
     Derived classes which include this mixin should have a
     subclass of `wtforms.Field` in their base classes.
     '''
-    def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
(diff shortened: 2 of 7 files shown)


lfdebrux:
Fingers crossed this shouldn't be a breaking change, but I'm testing it now with the briefs frontend, where I'm hoping it will also fix some bugs.

tests/forms/test_dmfields.py
```diff
@@ -73,13 +73,37 @@ class TestForm(wtforms.Form):
...
+@pytest.mark.parametrize("field_class", (
+    dmutils.forms.fields.DMRadioField, dmutils.forms.fields.DMBooleanField
+))
+def test_selection_field_id_is_suffixed_with_number(field_class):
+    class TestForm(wtforms.Form):
+        field = field_class()
+
+    form = TestForm()
+
+    assert form.field.id == "input-field-1"
```
katstevens, line 97: Can we test that DMRadioFields with more than one option don't all have suffix `-1`?

dmutils/forms/fields.py
```diff
@@ -46,6 +46,11 @@
 class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
     widget = DMCheckboxInput(hide_question=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if "id" not in kwargs:
+            self.id = self.id + "-1"
```
katstevens, line 52: I was a bit confused at first as I thought `DMBooleanField` covered Yes/No radio forms, but it's just a single checkbox, right? Can we add a comment perhaps, to help explain the hardcoded `-1` suffix?

lfdebrux: Done

katstevens:
Is there a way we can futureproof this for v3? What about adding in (yet another) kwarg flag -`govuk_v3=False` or something like that?

lfdebrux:
> Is there a way we can futureproof this for v3? What about adding in (yet another) kwarg flag -`govuk_v3=False` or something like that?

We could, but I'd rather wait until we've decided on what architecture we're going to use for GOV.UK Frontend forms. We might end up not using any of this code for GOV.UK Frontend forms.

katstevens:
> > Is there a way we can futureproof this for v3? What about adding in (yet another) kwarg flag -`govuk_v3=False` or something like that?
> 
> We could, but I'd rather wait until we've decided on what architecture we're going to use for GOV.UK Frontend forms. We might end up not using any of this code for GOV.UK Frontend form inputs.

OK, but we're going to have to decide that pretty soon!

tests/forms/test_dmfields.py
lfdebrux: They won't have the suffix `-1`... However because WTForms already suffixes a number each option will end up having the an id `input-field-1-{i}` 😅 

However, in our codebase we don't use the WTForms options for generating HTML, so I don't think those `id`s get used anywhere.

So this implementation is a hack, because we're relying on that assumption.

However, I can't think of a way to implement the correct behaviour of returning the id of the first option that isn't more broken. Any ideas?

lfdebrux: I've tried to circumvent the issue by adding a new property to the field objects, `href`. I've kept it as a separate commit for now but it can get folded in if you think it's better.

dmutils/forms/fields.py
```diff
@@ -46,6 +46,12 @@
 class DMBooleanField(DMFieldMixin, wtforms.fields.BooleanField):
     widget = DMCheckboxInput(hide_question=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # digitalmarketplace-frontend-toolkit suffixes the id with `-1`
```
phantas, line 51: Am I correct that we may need to revisit this when we migrate checkboxes?

lfdebrux: govuk-frontend v2 follows the same pattern. We would need to revisit it for govuk-frontend v3, but that could just mean overriding the property `id` for the relevant form.

dmutils/forms/mixins.py
```diff
@@ -58,6 +58,12 @@ class DMFieldMixin:
     '''
     def __init__(self, label=None, validators=None, hint=None, question_advice=None, **kwargs):
         super().__init__(label=label, validators=validators, **kwargs)
+
+        if "id" in kwargs:
+            self.href = kwargs["id"]
+        else:
+            self.href = "input-" + self.id
+
```
phantas, line 66: Have we tested if there is any weird recursiveness with this mixin or the selection one? I am happy to fire it up and have a go at it, if not.

lfdebrux: Please have a go at it :)

phantas: just saw the comments above too!

phantas: remind me, which page is best to test this?

phantas:
Testing this I noticed checkboxes get the right focus but the id looks weird:

Screenshot 2020-06-09 at 11 54 32

Screenshot 2020-06-09 at 11 47 58

lfdebrux:
> Testing this I noticed checkboxes get the right focus but the id looks weird:

Good spot! Fixed in https://github.com/alphagov/digitalmarketplace-content-loader/pull/83

phantas:
Tested again for DOS, all good
Screenshot 2020-06-09 at 12 48 33

PR #565, opened 2020-06-16
Log to stdout in addition to file

gidsg:
https://trello.com/c/VZBPctdG/106-2-our-apps-do-not-log-to-stdout-timebox-max-2-days

N.B. This should be tested on Preview to verify it is working as expected on PaaS.

### DIFF
--- a/dmutils/logging.py
+++ b/dmutils/logging.py
@@ -85,10 +85,21 @@ def init_app(app):
         return response
 
     logging.getLogger().addHandler(logging.NullHandler())
+    handlers = [logging.StreamHandler(sys.stdout)]
 
     del app.logger.handlers[:]
 
-    handler = get_handler(app)
+    if app.config.get('DM_PLAIN_TEXT_LOGS'):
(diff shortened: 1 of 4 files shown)


dmutils/logging.py
```diff
@@ -85,10 +85,21 @@ def after_request(response):
         return response
 
     logging.getLogger().addHandler(logging.NullHandler())
+    handlers = [logging.StreamHandler(sys.stdout)]
```
katstevens, line 88: Do we think this should be configurable per-app, like the other handler?

gidsg: @katstevens can we leave as-is and if we find a use case to make it configurable per-app we could always add it later?

katstevens: Sounds reasonable 😸

gidsg:
> could we update the reference to the now-deleted get_handler function in the tests.conftest.app_with_stream_logger fixture?

@katstevens great spot. I updated this now in f9c0c48.

PR #566, opened 2020-06-19
Add govuk-link class to format_links helper

katstevens:
https://trello.com/c/X4vKC3Bv/112-2-apply-govuk-frontend-styles-to-frameworks-content-for-the-create-a-dos-opportunity-journey

All our links should include the `govuk-link` class - this PR updates the filter function that converts 'text' links to `<a>` tags in user-generated-content (e.g. on public brief pages).

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app, init_manager
 
 
-__version__ = '52.5.0'
+__version__ = '52.6.0'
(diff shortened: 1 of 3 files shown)


lfdebrux:
Do we still need `app-break-link` @domoscargin?

domoscargin:
I don't think we need `app-break-link` - think `format_links` is only being used within the summary lists on Brief pages (and the preview), and Summary Lists break links already.

katstevens:
Woohoo I can take it out of here then 😸

katstevens:
@domoscargin @lfdebrux what about the case where `format_links` outputs a `<span>`? Removing `app-break-link` there means it won't have any class at all (so do we really need the `<span>`...)?

lfdebrux:
> @domoscargin @lfdebrux what about the case where `format_links` outputs a `<span>`? Removing `app-break-link` there means it won't have any class at all (so do we really need the `<span>`...)?

I vote for getting rid of the span.

PR #567, opened 2020-07-01
Look up email references directly in Notify API

katstevens:
https://trello.com/c/EnPFfpum/

After a recent failure in an email-sending job, we realised that the `allow_resend=False` option had some issues:
- the `DMNotifyClient` client instance first of all looks in its own cache, to get a list of sent email references. This cache is empty on initialisation, so for most email-sending scripts, each run on Jenkins have a new client instance and an empty cache.
- if the cache is empty, the client falls back to its `get_delivered_notifications()` (i.e., getting every recent delivered message, with a fresh API call), and looks for the reference in that list. Unfortunately, this list only contains the most recent 250 items! And that's across all scripts, jobs, apps, etc. Given most of our email sending scripts send thousands of emails each, if a job fails halfway then the majority of references won't be found.

To fix this I've added an optional flag to the `has_been_sent` and `send_email` methods, to allow a direct look up in the Notify  by reference. We should probably limit the usage to script retries only, for example if a `--resume-run-id` argument has been passed to the script. Otherwise there will be `O(n)` additional calls to Notify during any given script (as the cache is ignored and we need to check every individual email). However, 'expensive' is probably better than 'broken'...? As it stands, the `allow_resend=False` behaviour simply does not work outside the confines of a single script run.

Thoughts welcome!

### DIFF
--- a/tests/email/test_dm_notify.py
+++ b/tests/email/test_dm_notify.py
@@ -216,7 +216,7 @@ class TestDMNotifyClient(PatchExternalServiceLogConditionMixin):
                 email_reply_to_id=None
             )
 
-    def test_personalisation_passed(self, dm_notify_client, notify_send_email):
+    def test_send_email_personalisation_passed_to_client(self, dm_notify_client, notify_send_email):
         """Assert the expected existence of personalisation."""
         personalisation = {u'f\u00a3oo': u'bar\u00a3'}
         with mock.patch(self.client_class_str + '.' + 'send_email_notification') as email_mock:
@@ -236,7 +236,7 @@ class TestDMNotifyClient(PatchExternalServiceLogConditionMixin):
                 email_reply_to_id=None
             )
 
-    def test_personalisation_appears_in_reference(self, dm_notify_client, notify_send_email):
+    def test_send_email_includes_personalisation_in_reference(self, dm_notify_client, notify_send_email):
         """Assert personalisation is added to the auto generated reference."""
         personalisation = {u'f\u00a3oo': u'bar\u00a3'}
         with mock.patch(self.client_class_str + '.' + 'send_email_notification') as email_mock:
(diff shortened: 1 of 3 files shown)


lfdebrux:
When is the recent cache ever useful? I'm wondering if we should just take it out completely.

katstevens:
> When is the recent cache ever useful? I'm wondering if we should just take it out completely.

Good question. I think it's potentially useful in the apps, where the instance of the Notify client is more persistent (say, if a user submits multiple requests for an action and we only want to send one email), or if a script wants to handle retries within its own single run. But I haven't checked through all our usages to see if that's a) happening anywhere b) likely to happen anywhere.

If we do take it out, I'd rather do that as a separate PR anyway, as it would definitely be a breaking change.

lfdebrux:
> I think it's potentially useful in the apps, where the instance of the Notify client is more persistent (say, if a user submits multiple requests for an action and we only want to send one email), or if a script wants to handle retries within its own single run. But I haven't checked through all our usages to see if that's a) happening anywhere b) likely to happen anywhere.

I feel like in those instances it makes more sense to have an LRU cache rather than look at the most recent emails from Notify 🤔 But yeah, we can look at it in a separate PR.

tests/email/test_dm_notify.py
```diff
@@ -216,27 +216,7 @@ def test_send_email_with_external_reference(self, dm_notify_client, notify_send_
...
-                personalisation=personalisation
-            )
-
-            email_mock.assert_called_with(
-                self.email_address,
-                self.template_id,
-                personalisation=personalisation,
-                reference='CLkthp1ZgyeBSMCgQj-zf18netwEf3J9aJxLcm-FZ4s=',
-                email_reply_to_id=None
-            )
```
lfdebrux, line 237: Why have you deleted this test?

katstevens: I couldn't see any difference at all between it and the one below it.

PR #568, opened 2020-07-08
draft

agalamatis:
trello: https://trello.com/c/3uVc3tgm/124-2-spike-replacing-radio-selection-buttons-with-govuk-frontend-radios-components-1-day

Function and tests in utils repo for converting the dictionaries content generates and are currently used to a list of dictionaries with input and label keys with the original values of radios and checkboxes.

### DIFF
--- a/dmutils/forms/helpers.py
+++ b/dmutils/forms/helpers.py
@@ -22,6 +22,23 @@ def remove_csrf_token(data):
 
 
 def govuk_option(option: typing.Dict) -> typing.Dict:
+    """Converts one digitalmarketplace-frontend-toolkit style option (of the radio and checkboxes elements)
+    into a format suitable for use with either digitalmarketplace-frontend-toolkit
+    templates/macros or govuk-frontend macros.
+
+    Example usage:
+
+    # app.py
+
+        @flask.route("/", methods=["GET", "POST"])
+        def view():
+            lot_question = {
+            option["value"]: option
+            for option in govuk_option(
+                ContentQuestion(content_loader.get_question(framework_slug, 'services', 'lot')).get('options')
+            )
+        }
+    """
     if option:
         # DMp's yml does not requires only labels, which is used as the value if none is provided
         item = {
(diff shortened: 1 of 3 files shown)


gidsg:
@agalamatis can you mark this PR ready to review?

dmutils/forms/helpers.py
```diff
@@ -17,3 +19,21 @@ def remove_csrf_token(data):
         del cleaned_data['csrf_token']
 
     return cleaned_data
+
+
+def govuk_option(option: typing.Dict) -> typing.Dict:
+    if option:
```
katstevens, line 25: Could we add in an example of how/when this might be used in a frontend app?

gidsg:
@agalamatis can you edit the commit message to make it something descriptive?

dmutils/forms/helpers.py
```diff
@@ -17,3 +19,21 @@ def remove_csrf_token(data):
         del cleaned_data['csrf_token']
 
     return cleaned_data
+
+
+def govuk_option(option: typing.Dict) -> typing.Dict:
```
lfdebrux, line 24: If we're not being more specific about the type of Dict we can just write `dict` and save ourselves an import.

```suggestion
def govuk_option(option: dict) -> dict:
```

katstevens:
I've tried this out locally for the Choose Lot form, following the example above and the Design System documentation for adding in a `govukRadios` component. Here's what I get:

choose-lot-govukRadios

It looks pretty good but if [you look at the same page on production](https://www.digitalmarketplace.service.gov.uk/buyers/direct-award/g-cloud/choose-lot), we're missing the form hint examples. So I don't think having separate `description` and `hint` attributes will be useful here. I'm not sure about combining them - probably better to review the content itself?

katstevens:
@agalamatis I'd rather avoid having to create our own version of this component if we can avoid it! The Design System guidance recommends 'keeping it simple',  so perhaps let's stick with trying to reduce the amount of content in the hint if we can. I think keeping the conversion from `"description": "blah"` to `"hint": {"text": "blah"}` is a good idea.

katstevens:
Sorry, one last thing needed before we can merge - a version number 😸

agalamatis:
> Sorry, one last thing needed before we can merge - a version number 😸

minor version bumped

PR #569, opened 2020-07-08
Add to_link helper for creating <a> tag strings

domoscargin:
While working on https://trello.com/c/EcbpN7V0/129-2-replace-summary-tables-on-saved-searches-page-with-tables I found myself really wishing I could just create a link without writing up a long string.

`to_link` takes a url string, an optional label string, and a set of key-value args for any other attributes, and outputs an HTML <a> tag string. Future iterations could accept numbers and other types and handle them.

`to_link('http://www.gov.uk', 'GOV.UK')` = `<a href="http://www.gov.uk">GOV.UK</a>"`

As I'm not a Python expert, I'd appreciate somebody picking this apart.

### DIFF
--- a/dmutils/html.py
+++ b/dmutils/html.py
@@ -1,0 +1,21 @@
+from jinja2 import Markup, escape
+
+
+def link_to(url, label='', **attrs):
+    """Create a hyperlink with the given text pointing to the URL"""
+
+    if url == '' or not isinstance(url, str):
+        raise ValueError('link_to expects URL to be a non-empty string')
+
+    if label == '' or label is None:
+        label = url
+
+    output = ['<a']
+    output.append(f' href="{escape(url)}"')
+    for key, value in attrs.items():
+        if isinstance(value, str):
+            output.append(' {}="{}"'.format(escape(key), escape(value)))
+    output.append('>')
+    output.append(escape(label))
+    output.append('</a>')
+    return Markup("".join(output))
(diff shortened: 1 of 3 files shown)


dmutils/html.py
```diff
@@ -0,0 +1,21 @@
+from jinja2 import Markup, escape
+
+
+def link_to(url, label='', **attrs):
+    """Create a hyperlink with the given text pointing to the URL"""
```
katstevens, line 5: Could we re-use the logic in `dmutils.filters.format_links` instead?

dmutils/html.py
```diff
@@ -0,0 +1,21 @@
...
+
+    output = ['<a']
+    output.append(f' href="{escape(url)}"')
+    for key, value in attrs.items():
+        if isinstance(value, str):
+            output.append(' {}="{}"'.format(escape(key), escape(value)))
+    output.append('>')
+    output.append(escape(label))
+    output.append('</a>')
+    return Markup("".join(output))
```
lfdebrux, line 21: I like what you've done here, building the string up in a list and then using `join()` is a good idea as it's usually quicker and more flexible than appending to a string (in Python at least).

That said, I think this would be more readable if you relied more heavily on the power of Python f-strings. What I would do is assemble the attrs as a separate string using a list, and then create the entire element with one f-string. Something like

```
<a href={escape(url)}{attrs}>{escape(label)}</a>
```

domoscargin:
Thanks for your comments, @katstevens and @lfdebrux. I've redone the [buyer-fe PR](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1058) so that this is no longer needed, so I'm closing this.

PR #576, opened 2020-10-08
Remove api_stubs

lfdebrux:
We don't use these anywhere in our code base anymore (we use dmtestutils instead), we should remove them.


bjgill (approved):
:shipit:

PR #577, opened 2020-10-15
Remove Flask-Script

lfdebrux:
https://trello.com/c/kbhJdTm3/567-remove-flaskscript-from-dmutils

It is deprecated, we no longer need it, let's get rid of it! :fire:

This is a breaking change, but to install it into an app we simply need to delete most of the code in the apps' `./application.py` file :grin:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 54.0.0
+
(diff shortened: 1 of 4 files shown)


lfdebrux:
Example of how to fold this into an app: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1085

bjgill (approved):
Looks sensible.

PR #578, opened 2020-10-16
Block ')' from ending URLs

bjgill:
Trello: https://trello.com/c/U9O7zTCU/1437-fix-broken-urls-in-3-of-briefs

Previously, the URL-detection heuristic would see "something (like http://example.com) this" and detect "http://example.com)" as the URL. Whilst URLs can validly include ')', it is very rare. Indeed, of the 1000+ closed briefs we have, 49 have URLs that end in ')' where they should not. No closed brief has a URL that intentionally ends in ')'.

You can see an example of this problem in https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities/13014, which was reported in x-gov Slack.

Thus, update our URL-detection heuristic to to forbid ')' as the last character of a URL. This will do the right thing for users in the vast majority of cases. I have not found any URLs that this change will break.

Add tests to confirm the new behaviour.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app
 
 
-__version__ = '54.0.0'
+__version__ = '54.0.1'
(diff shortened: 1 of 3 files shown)


gidsg (approved):
🔗 👍

bjgill:
I did consider it. There are people on the internet who've posted regexes that should do just that. Unfortunately, none of these people seem to have released their work under an MIT-compatible licence.

So our options would be:
1. Add a new dependency to do the URL detection for us.
2. Come up with our own regex.

Either of those would turn this from something I can do in ~1 hour to something that would take significantly more effort. And given that I've not found any user-posted URLs that deliberately end with a closing parenthesis, I don't think it's worth that effort.

PR #580, opened 2020-10-27
Add ability to override s3 endpoint URL for local development

lfdebrux:
I want to run dmrunner with [localstack] instead of real AWS access. This PR changes the dmutils s3 client to use the url in the config variable in `S3_ENDPOINT_URL` as the endpoint URL for boto3, if (and only if) Flask is running in development mode.

[localstack]: https://github.com/localstack/localstack


alex9smith (approved):
Looks good!

PR #584, opened 2020-11-05
Adds Redis session support behind env var

gidsg:
https://trello.com/c/XXL04KtG/379-3-run-the-new-user-session-management-on-the-local-development-environment
This adds Redis session support behind an env var so that we can manage the rollout.
Code based largely on @lfdebrux's prior spike.

### DIFF
--- a/dmutils/flask_init.py
+++ b/dmutils/flask_init.py
@@ -63,7 +63,8 @@ def init_app(
     if login_manager:
         login_manager.init_app(application)
         import dmutils.session
-        dmutils.session.init_app(application)
+        if os.environ.get('DM_USE_REDIS_SESSION_TYPE'):
+            dmutils.session.init_app(application)
     if search_api_client:
         search_api_client.init_app(application)
 
--- a/dmutils/flask_init.py
+++ b/dmutils/flask_init.py
@@ -62,8 +62,8 @@ def init_app(
         db.init_app(application)
     if login_manager:
         login_manager.init_app(application)
-        import dmutils.session
         if os.environ.get('DM_USE_REDIS_SESSION_TYPE'):
+            import dmutils.session
             dmutils.session.init_app(application)
     if search_api_client:
         search_api_client.init_app(application)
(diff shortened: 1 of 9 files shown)


dmutils/flask_init.py
```diff
@@ -60,6 +62,9 @@ def init_app(
         db.init_app(application)
     if login_manager:
         login_manager.init_app(application)
+        import dmutils.session
+        if os.environ.get('DM_USE_REDIS_SESSION_TYPE'):
```
lfdebrux, line 66: Could we just use the existing `SESSION_TYPE` config var here?

```suggestion
        if app.config.get("SESSION_TYPE") == "redis":
```

I think the advantage of this is 1) it doesn't need to stay as an environment variable later on 2) one less thing to remember 😁

lfdebrux:
We should probably have one or two tests for this 🤔

dmutils/flask_init.py
gidsg: @lfdebrux the reason I didn't use `SESSION_TYPE` here is that this allows us to do a gradual rollout of `SESSION_TYPE` to the frontend apps without breaking existing sessions across apps. 

Once this change is rolled out though, the condition could be changed as you suggest or potentially removed entirely.

lfdebrux: I see... why doesn't using `SESSION_TYPE` let us do a gradual rollout?

lfdebrux: Or rather, doesn't it amount to the same thing? If we can change `DM_USE_REDIS_SESSION_TYPE` in one go can't we change `SESSION_TYPE` in one go? Or am I missing something? 🤔

gidsg: @lfdebrux I found in testing that trying to set `SESSION_TYPE` globally e.g. using dmrunner settings required that it already exist with some value within the app config, else it was stripped out, so it was not set successfully. Using a `DM_` prefix env var is set globally though. Happy to talk this through more with you if that would help as I also found it confusing.

lfdebrux: Ah, I see. Well that makes sense then, let's go with this method you've already worked out!

gidsg:
@lfdebrux I have added some tests now, please may you re-review? 
I am aware commit history is a bit messy but will rebase once approved.

lfdebrux (approved):

PR #585, opened 2020-11-10
Set session type in utils

gidsg:
Follow-up to https://github.com/alphagov/digitalmarketplace-utils/pull/584. On reflection, it makes sense to set `SESSION_TYPE` directly in utils, rather than having to set it in each individual frontend app. 

N.B. This will not cause redis to be automatically used as it is still behind the `DM_USE_REDIS_SESSION_TYPE` env var.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Records breaking changes from major version bumps
 
 ## 55.0.0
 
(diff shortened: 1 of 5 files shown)


alex9smith (approved):
You're right, this does seem much better than setting per app 👍

PR #586, opened 2020-11-16
get vcap_services directly from env

gidsg:
https://trello.com/c/yS6OcI31/382-2-run-the-new-user-session-management-on-the-preview-environment-temporarily-for-testing
Previously we were erroneously attempting to read `VCAP_SERVICES` from app config which it is not included in.

This PR changes it to read `VCAP_SERVICES` directly from the env, I think this makes sense as it is not something we would want to modify in per-app config.

### DIFF
--- a/dmutils/cloudfoundry.py
+++ b/dmutils/cloudfoundry.py
@@ -4,10 +4,9 @@ Support for running on a CloudFoundry platform such as GOV.UK PaaS.
 
 import json
 import os
-from typing import Union
 
 
-def get_vcap_services() -> Union[dict, KeyError]:
+def get_vcap_services() -> dict:
     return json.loads(os.environ["VCAP_SERVICES"])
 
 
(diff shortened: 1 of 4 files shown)


dmutils/cloudfoundry.py
```diff
@@ -3,15 +3,12 @@
...
 
 
-def get_vcap_services(app, default=None) -> dict:
-    if app.config.get("VCAP_SERVICES") is None:
-        return default
-
-    vcap_services = json.loads(app.config["VCAP_SERVICES"])
-
-    return vcap_services
+def get_vcap_services() -> Union[dict, KeyError]:
```
alex9smith, line 10:
```suggestion
def get_vcap_services() -> dict]:
```
I think this return type can just be `dict` It might raise a `keyError` instead, but if it does return something, that something will be a dict

alex9smith (approved):

PR #588, opened 2020-11-17
Change EmailTemplateError to be subclass of EmailError

lfdebrux:
Thanks @bjgill for spotting this when reviewing https://github.com/alphagov/digitalmarketplace-scripts/pull/580#discussion_r525106570

### DIFF
--- a/tests/email/test_dm_notify.py
+++ b/tests/email/test_dm_notify.py
@@ -507,3 +507,18 @@ class TestDMNotifyClient(PatchExternalServiceLogConditionMixin):
 
             with pytest.raises(EmailTemplateError):
                 dm_notify_client.send_email(self.email_address, self.template_id)
+
+    def test_catch_email_error_catches_email_template_error(self, dm_notify_client, notify_example_http_error):
+        notify_example_http_error._message = [
+            {"error": "BadRequestError", "message": "Missing personalisation: application_deadline"}
+        ]
+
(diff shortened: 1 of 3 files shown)


tests/email/test_dm_notify.py
```diff
@@ -507,3 +507,18 @@ def test_raises_email_template_error_if_personalisation_missing(self, dm_notify_
...
+
+        with mock.patch(self.client_class_str + '.' + 'send_email_notification') as email_mock:
+            email_mock.side_effect = notify_example_http_error
+
+            try:
+                dm_notify_client.send_email(self.email_address, self.template_id)
+            except EmailError as e:
+                assert isinstance(e, EmailTemplateError)
+            else:
+                assert False, "EmailError was not raised"
```
bjgill, line 524: For future reference, pytest has a built-in way of testing this sort of thing: https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

bjgill (approved):
One thing for future reference, but not worth delaying things to fix

tests/email/test_dm_notify.py
lfdebrux: Yeah, for this particular test I wanted to test the exact code I was using in alphagov/digitalmarketplace-scripts#580 (in lieu of proper tests in that repo) so I could feel more confident about it.

PR #590, opened 2020-12-01
Add date validator for date inputs

lfdebrux:
https://trello.com/c/Ex9qw68e/184-3-more-granular-date-error-messages

Add `DateValidator` class to validate date input submissions from user. This should (hopefully) be usable with both WTForms DMDateFields and content loader date questions.

### DIFF
--- a/dmutils/forms/validators.py
+++ b/dmutils/forms/validators.py
@@ -258,6 +258,10 @@ class DateValidator:
             # remove processing errors from field
             field.errors[:] = []
 
+            # add errors to specific form field fields
+            for form_field_field in error_fields:
+                getattr(field.form_field, form_field_field).errors = [error_message]
+
             validation_error = ValidationError(error_message)
             validation_error.fields = error_fields
             raise validation_error
--- a/dmutils/forms/validators.py
+++ b/dmutils/forms/validators.py
@@ -10,7 +10,7 @@ DateValidator -- Error messages for a date input
 import datetime
 from typing import Any, Dict, Optional
 
-from wtforms.validators import ValidationError
+from wtforms.validators import StopValidation, ValidationError
 
 from dmutils.email.helpers import validate_email_address
 
@@ -269,6 +269,6 @@ class DateValidator:
             for form_field_field in error_fields:
                 getattr(field.form_field, form_field_field).errors = [error_message]
 
-            validation_error = ValidationError(error_message)
+            validation_error = StopValidation(error_message)
             validation_error.fields = error_fields
             raise validation_error
--- a/dmutils/forms/validators.py
+++ b/dmutils/forms/validators.py
@@ -159,7 +159,14 @@ class DateValidator:
     def _error(self, error, fields, **kwargs):
         e = ValueError(self.format_error_message(error, **kwargs))
         e.error = error
-        e.fields = fields
+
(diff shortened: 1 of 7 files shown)


dmutils/forms/validators.py
```diff
@@ -82,3 +92,144 @@ def __call__(self, form, field):
...
+
+class DateValidator:
+    """Error messages for a date input
+
+    Implements the guidance in [1].
+
+    Instead of taking a `message` DateValidator has two properties,
+    `whatever_it_is` and `Whatever_it_is`, which are stand-ins for whatever it
+    is you are asking for. `whatever_it_is` is used in the middle of a
+    sentence, and should include a participle: for instance, if you are asking for the
```
domoscargin, line 105: Looks like something got cut out here

lfdebrux: I'm not seeing it...

lfdebrux: Fixed it

domoscargin (approved):
This looks good to me, though somebody more familiar with Python might want to give it a glance.

Do these 3 cases cover our date errors as is? GreaterThan looks like it covers a lot of the comparative cases (past, future, etc).

lfdebrux:
> Do these 3 cases cover our date errors as is? GreaterThan looks like it covers a lot of the comparative cases (past, future, etc).

Yes, I think so. I'll add a test demonstrating how we should be able to use `GreaterThan` with dates.

tests/forms/test_forms.py
```diff
@@ -31,3 +34,25 @@ class TestForm(wtforms.Form):
     assert form.validate() is False
     assert form.date.errors == ["Date must be a real date"]
     assert form.date.day.errors
+
+
+@pytest.mark.parametrize("start_date, end_date, validates", (
```
domoscargin, line 39: 👍

lfdebrux:
I've realised I've made a small mistake: in cases where more than one input would be highlighted the Design System guidance say we should instead highlight all the inputs. I'm going to make this PR a draft while I fix that.

dmutils/forms/validators.py
```diff
@@ -90,3 +96,179 @@ def __call__(self, form, field):
...
+            date_entered_cant_be_correct_error_message
+            or "{Whatever_it_is} must be a real date"
+        )
+
+    def _error(self, error, fields, **kwargs):
+        e = ValueError(self.format_error_message(error, **kwargs))
+        e.error = error
+
+        # Design System guidance is that if more than one field has an error
+        # we should highlight all fields.
```
domoscargin, line 164: 👍

dmutils/forms/validators.py
```diff
@@ -90,3 +96,179 @@ def __call__(self, form, field):
...
+        :raises ValueError:
+        """
+        year, month, day = data["year"], data["month"], data["day"]
+
+        invalid = set()
+        if not (1 <= day and day <= 31):
+            invalid.add("day")
+        if not (1 <= month and month <= 12):
+            invalid.add("month")
+        if not (datetime.MINYEAR <= year and year <= datetime.MAXYEAR):
```
domoscargin, line 230: TIL I learned about `MINYEAR` and `MAXYEAR`

domoscargin (approved):

PR #592, opened 2020-12-07
Fix bugs in GreaterThan validator

lfdebrux:
While working on another ticket I found some bugs in the GreaterThan 
validator:

- it will raise an exception if there is no data for field
- it will not compare itself to 0

This commit adds some tests and fixes the bugs.

Now the GreaterThan validator will not compare itself if has no input
(same as we were doing if other had no input), and will do the comparison
 if other is 0.

### DIFF
--- a/tests/forms/test_validators.py
+++ b/tests/forms/test_validators.py
@@ -1,3 +1,4 @@
+from datetime import date
 
 import pytest
 import mock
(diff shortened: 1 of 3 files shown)


tests/forms/test_validators.py
```diff
@@ -52,3 +55,55 @@ def test_email_validator_does_not_raise_on_a_valid_email_address(form, field, em
     field.data = email_address
 
     validator(form, field)
+
+
+class TestGreaterThan:
```
domoscargin, line 60: Should one set of parameters be a date format, since we use this on those questions?

I'm not sure if we're comparing strings or numbers (or date objects?) in that case.

domoscargin (approved):
Good catch - this makes sense to me (I've left one non-blocking comment).

tests/forms/test_validators.py
lfdebrux: Sure, it shouldn't in theory make a difference, but it definitely makes sense to test it!

PR #594, opened 2020-12-30
Replace make with invoke

lfdebrux:
Ticket: https://trello.com/c/hCCo28Bc/1329-add-makefiles-to-digitalmarketplace-utils-and-reduce-duplication-across-apps

We want to be able to reduce duplication of code across our repos; one of the places we have a lot of duplicated code is in our Makefiles; all of our repos have very similar Makefiles with small historical differences.

This pull request replaces Make with invoke, using the tasks from alphagov/digitalmarketplace-dev-tools.

See alphagov/digitalmarketplace-developer-tools#1 for more details.

---

Note: the tests won't pass until alphagov/digitalmarketplace-developer-tools#1 is merged, as they now depend on invoke

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -11,3 +11,4 @@ max-complexity = 12
 max-line-length = 120
 per-file-ignores =
     **/__init__.py : F401
+    tasks.py: F401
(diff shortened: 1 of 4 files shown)


alex9smith (approved):
Works locally for me! This is so much cleaner than the Makefile 👍

PR #596, opened 2021-01-15
Log Direct Plus API errors

bjgill:
Trello: https://trello.com/c/bZ4XlASA

We had a recent incident involving the Direct Plus API. We were getting 400 responses for valid DUNS numbers. This was a problem because no API errors were logged, which made debugging very difficult. So add logging for unexpected errors.

Also, understanding what the APi client did was difficult, so add some documentation.

Also, start following the documented API more closely, and only treat the errors as expected when we also get the error codes we expect. This is taken from the API documentation: https://directplus.documentation.dnb.com/errorsAndInformationMessages.html.

### DIFF
--- a/dmutils/direct_plus_client.py
+++ b/dmutils/direct_plus_client.py
@@ -1,5 +1,10 @@
 import logging
 import requests
+from requests import HTTPError
+
+# See https://directplus.documentation.dnb.com/errorsAndInformationMessages.html
+DUNS_NUMBER_NOT_FOUND = 404, "10001"
+DUNS_NUMBER_INVALID = 400, "10003"
 
 
 class DirectPlusClient(object):
@@ -79,8 +84,16 @@ class DirectPlusClient(object):
             f'data/duns/{duns_number}', payload={'productId': 'cmpelk', 'versionId': 'v2'}
         )
 
-        if response.status_code in (404, 400):
-            # 404 - DUNs number not found
-            # 400 - Incorrect format
-            return None
+        try:
+            response.raise_for_status()
+        except HTTPError as exception:
+            try:
+                error = response.json()['error']
+
+                if (response.status_code, error["errorCode"]) in [DUNS_NUMBER_INVALID, DUNS_NUMBER_NOT_FOUND]:
+                    return None
+
+                self.logger.error(f"Unable to get supplier by DUNS number: {error}")
+            except (ValueError, KeyError):
+                self.logger.error(f"Unable to get supplier by DUNS number: {exception}")
         return response.json()['organization']
--- a/dmutils/direct_plus_client.py
+++ b/dmutils/direct_plus_client.py
@@ -1,4 +1,6 @@
 import logging
+from typing import Optional
+
 import requests
 from requests import HTTPError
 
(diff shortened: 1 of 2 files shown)


dmutils/direct_plus_client.py
```diff
@@ -71,16 +78,29 @@ def _direct_plus_request(
...
+            response.raise_for_status()
+        except HTTPError as exception:
+            try:
+                error = response.json()['error']
+
+                if (response.status_code, error["errorCode"]) in [DUNS_NUMBER_INVALID, DUNS_NUMBER_NOT_FOUND]:
+                    return None
+
+                self.logger.error(f"Unable to get supplier by DUNS number: {error}")
+            except (ValueError, KeyError):
```
alex9smith, line 104: Was there a specific condition you're expecting that would raise a `ValueError`?
I wonder if we should also raise an explicit error on a `JSONDecodeError` (but that might be out of scope for this PR and might change the behaviour of how we use it)

alex9smith (approved):
Looks sensible to me. A couple of non-blocking questions

dmutils/direct_plus_client.py
bjgill: We get a `ValueError` if the response body is not valid JSON - at least according to the documentation of the `response.json` method.

I've not seen `JSONDecodeError` - what raises it?

alex9smith: Interesting - `JSONDecodeError` is the one I got when I tried to parse something that wasn't valid JSON: 

```
IPython 7.12.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import requests

In [2]: response = requests.get('https://www.google.com')

In [3]: response.json()
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
<ipython-input-3-34c975fa6377> in <module>
----> 1 response.json()

~/dmp/digitalmarketplace-scripts/venv/lib/python3.6/site-packages/requests/models.py in json(self, **kwargs)
    896                     # used.
    897                     pass
--> 898         return complexjson.loads(self.text, **kwargs)
    899
    900     @property

~/.pyenv/versions/3.6.9/lib/python3.6/json/__init__.py in loads(s, encoding, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    352             parse_int is None and parse_float is None and
    353             parse_constant is None and object_pairs_hook is None and not kw):
--> 354         return _default_decoder.decode(s)
    355     if cls is None:
    356         cls = JSONDecoder

~/.pyenv/versions/3.6.9/lib/python3.6/json/decoder.py in decode(self, s, _w)
    337
    338         """
--> 339         obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    340         end = _w(s, end).end()
    341         if end != len(s):

~/.pyenv/versions/3.6.9/lib/python3.6/json/decoder.py in raw_decode(self, s, idx)
    355             obj, end = self.scan_once(s, idx)
    356         except StopIteration as err:
--> 357             raise JSONDecodeError("Expecting value", s, err.value) from None
    358         return obj, end

JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

bjgill: Interesting. What version of `requests` do you have?

alex9smith: I agree that the doc for `response.json` says it should be `ValueError` though 
```
In [4]: response.json?
Signature: response.json(**kwargs)
Docstring:
Returns the json-encoded content of a response, if any.

:param \*\*kwargs: Optional arguments that ``json.loads`` takes.
:raises ValueError: If the response body does not contain valid json.
File:      ~/dmp/digitalmarketplace-scripts/venv/lib/python3.6/site-packages/requests/models.py
Type:      method
```

alex9smith: > What version of requests do you have?
`2.23.0`

bjgill: Oh - you're using the scripts env. So that is what it really will do. That's a bit sad. Thanks for flagging - I'll take a closer look.

bjgill: It turns out that it's all OK. `json.decoder.JSONDecodeError` is a subclass of `ValueError`, so the catch will work. But thanks for flagging.

PR #597, opened 2021-02-10
Handle all errors from Notify client

bjgill:
Trello: https://trello.com/c/ct6lxcle

Most of the time, Notify error messages are of type `List[Dict]`. However, sometimes, they are a string - see https://github.com/alphagov/notifications-python-client/pull/189.

Handle both properly. This will fix the errors we see when we get unexpected errors from Notify (for example, 503s).

### DIFF
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -103,15 +103,13 @@ class DMNotifyClient:
     def _log_email_error_message(self, to_email_address, template_name_or_id, reference, error):
         """Format a logical error message from the error response and send it to the logger"""
 
-        error_messages = []
-        for error_message in error.message:
-            error_messages.append(
-                "{status_code} {error}: {message}".format(
-                    status_code=error.status_code,
-                    error=error_message["error"],
-                    message=error_message["message"],
-                )
-            )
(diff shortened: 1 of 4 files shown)


gidsg (approved):
LGTM 👍

dmutils/email/dm_notify.py
```diff
@@ -185,7 +183,8 @@ def send_email(
 
         except HTTPError as e:
             self._log_email_error_message(to_email_address, template_name_or_id, reference, e)
-            if any(msg["message"].startswith("Missing personalisation") for msg in e.message):
+            if isinstance(e.message, list) and \
+                    any(msg["message"].startswith("Missing personalisation") for msg in e.message):
```
alex9smith, line 187: Is there no case when we'd want to raise an `EmailTemplateError` for a `str` error message?

alex9smith (approved):
The approach seems sensible and thanks for opening the PR to Notify's client as well! One non-blocking question

dmutils/email/dm_notify.py
bjgill: No - the message is only a string if the response body was not an expected error response from Notify.

PR #598, opened 2021-02-10
Use mypy to test typing

bjgill:
This PR adds [`mypy`](http://mypy-lang.org/) static type checking to the checks performed on `make test` (and hence CI).

I've also fixed/ignored the issues mypy picked up as appropriate.

I copied the approach from supplier-frontend: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1284

### DIFF
--- a/dmutils/cloudfoundry.py
+++ b/dmutils/cloudfoundry.py
@@ -4,13 +4,14 @@ Support for running on a CloudFoundry platform such as GOV.UK PaaS.
 
 import json
 import os
+from typing import Dict, List
 
 
 def get_vcap_services() -> dict:
-    return json.loads(os.environ["VCAP_SERVICES"])
+    return json.loads(os.environ["VCAP_SERVICES"])  # type: ignore
 
 
-def get_service_by_name_from_vcap_services(vcap_services: dict, name: str) -> dict:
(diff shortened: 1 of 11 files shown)


lfdebrux:
🏆

dmutils/cloudfoundry.py
```diff
@@ -4,13 +4,14 @@
 
 import json
 import os
+from typing import Dict, List
 
 
 def get_vcap_services() -> dict:
-    return json.loads(os.environ["VCAP_SERVICES"])
+    return json.loads(os.environ["VCAP_SERVICES"])  # type: ignore
```
lfdebrux, line 11: Instead of doing an ignore we can do a typecast with [`cast()`](https://docs.python.org/3/library/typing.html#typing.cast):

```suggestion
    return cast(dict, json.loads(os.environ["VCAP_SERVICES"]))
```

lfdebrux (approved):

dmutils/cloudfoundry.py
bjgill: Ah, that's a neat thing I was unaware of.

bjgill: Have added casts where I can

bjgill:
Added casts and rebased against master

lfdebrux (approved):

PR #599, opened 2021-02-11
Add a message to be logged when an email is sent successfully

lfdebrux:
We get a lot of user support queries asking whether an email was sent correctly or not; it would be nice to be able to always see a log message saying "email was sent". This PR adds a log message to the `DMNotifyClient.send_email_message()` method, so whatever logging there may be around the call you can always be sure at least one message was logged whether the email was sent successfully or not.

Oh and it does some refactoring as well.

### DIFF
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -204,6 +204,8 @@ class DMNotifyClient:
                 raise EmailTemplateError(str(e))
             raise EmailError(str(e))
 
+        self._log(logging.INFO, f"Email with reference '{reference}' sent to Notify successfully", email_obj)
+
         self._update_cache(reference)
 
         return response
--- a/dmutils/email/dm_notify.py
+++ b/dmutils/email/dm_notify.py
@@ -1,5 +1,9 @@
 # -*- coding: utf-8 -*-
 """Digital Marketplace Notify integration."""
+
+from typing import Dict, NamedTuple, Optional
+import logging
+
 from flask import current_app
 from notifications_python_client import NotificationsAPIClient
 from notifications_python_client.errors import HTTPError
@@ -7,6 +11,13 @@ from notifications_python_client.errors import HTTPError
 from dmutils.email.exceptions import EmailError, EmailTemplateError
 from dmutils.email.helpers import hash_string
 from dmutils.timing import logged_duration_for_external_request as log_external_request
+
+
+class DMNotifyEmail(NamedTuple):
+    to_email_address: str
+    template_name_or_id: str
+    reference: str
+    personalisation: Optional[Dict[str, str]] = None
 
 
 class DMNotifyClient:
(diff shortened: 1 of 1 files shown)


dmutils/email/dm_notify.py
```diff
@@ -9,6 +13,13 @@
 from dmutils.timing import logged_duration_for_external_request as log_external_request
 
 
+class DMNotifyEmail(NamedTuple):
+    to_email_address: str
+    template_name_or_id: str
+    personalisation: Optional[Dict[str, str]] = None
+    reference: Optional[str] = None
```
bjgill, line 20: Nice.

Though, why is `reference` optional?

dmutils/email/dm_notify.py
```diff
@@ -182,12 +198,14 @@ def send_email(
...
 
         except HTTPError as e:
-            self._log_email_error_message(to_email_address, template_name_or_id, reference, e)
+            self._log_email_error_message(email_obj, e)
             if isinstance(e.message, list) and \
                     any(msg["message"].startswith("Missing personalisation") for msg in e.message):
                 raise EmailTemplateError(str(e))
             raise EmailError(str(e))
 
+        self._log(logging.INFO, "Email sent to Notify successfully", email_obj)
```
bjgill, line 207: What does this look like on the command line when being run as part of a script? Can we see the email object's fields as well as the message?

dmutils/email/dm_notify.py
```diff
@@ -94,13 +105,13 @@ def get_reference(to_email_address, template_id, personalisation=None):
...
         :return: Hashed string 'reference' to be passed to client.send_email_notification or self.send_email
         """
-        personalisation_string = u','.join(
-            list(map(lambda x: u'{}'.format(x), personalisation.values()))
-        ) if personalisation else u''
-        details_string = u'|'.join([to_email_address, template_id, personalisation_string])
+        personalisation_string = ",".join(
+            list(map(lambda x: "{}".format(x), personalisation.values()))
+        ) if personalisation else ""
+        details_string = u"|".join([to_email_address, template_id, personalisation_string])
```
bjgill, line 111:
```suggestion
        details_string = "|".join([to_email_address, template_id, personalisation_string])
```

bjgill (commented):
A good improvement - this should save us a lot of repetitive logging in various scripts. Just a few questions.

dmutils/email/dm_notify.py
lfdebrux: Ah, that's a holdover from a different approach I tried and gave up on, I will fix that, thanks.

dmutils/email/dm_notify.py
lfdebrux: Thanks!

dmutils/email/dm_notify.py
lfdebrux: 🤔 I think not, but I'll get back to you on that

bjgill: :+1: If not, might it be worth adding the `reference` to the message?

lfdebrux: Yeah that's probably a good idea whatever the case; the handling of the `extra` field depends on the formatter and so could be different in different circumstances.

bjgill (approved):

PR #603, opened 2021-03-10
Rewrite DirectPlusClient.get_organization_by_duns_number()

lfdebrux:
Ticket: https://trello.com/c/52Xz6MJM/2184-users-are-able-to-create-supplier-accounts-with-invalid-duns-numbers

We are having an issue where users can create supplier accounts with invalid DUNS numbers. This appears to be because although we expect the Direct+ error code `10001` for DUNS number not found, we are actually getting 40105.

This PR fixes this by changing the code to ignore the Direct+ error code and just look at the HTTP status code to determine whether the DUNS number is valid/found or not.

Also we now raise exceptions instead of returning `None`, which makes this a PR a breaking change.


dmutils/direct_plus_client.py
```diff
@@ -95,12 +107,34 @@ def get_organization_by_duns_number(self, duns_number) -> Optional[dict]:
...
+                raise DirectPlusError(error_message, exception)
+
+        try:
+            organization = cast(dict, response.json()['organization'])
+        except (ValueError, KeyError) as e:
+            # this should never happen, so let's propogate it so it is noisy
+            self.logger.error(
+                f"Unable to parse DirectPlus API response (HTTP code {response.status_code}): {response}"
+            )
+            raise e
```
bjgill, line 138: Python nit: if you're just re-raising `e`, you don't need to explicitly catch it - https://docs.python.org/3/tutorial/errors.html#raising-exceptions

dmutils/direct_plus_client.py
```diff
@@ -78,15 +87,18 @@ def _direct_plus_request(
...
-        :raises KeyError on unexpected failure if the response body is JSON.
-        :raises ValueError on unexpected failure if the response body is not valid JSON.
+        :return: the organization details corresponding to the DUNS number
+        :raises DUNSNumberNotFound: if there is no organization for the DUNS number
+        :raises DUNSNumberInvalid: if `duns_number` is not a valid DUNS number
+        :raises DirectPlusError: if there is an unexpected error response from the API
         """
+
+        # request the company profile, linkage, and executives (v2)
+        # https://directplus.documentation.dnb.com/openAPI.html?apiID=cmpelkv2
```
bjgill, line 101: :heart:

bjgill (approved):
Looks sensible. Fixes our current problem, and also forces callers to explicitly handle failure.

Just one nit about Python use - for your information only, doesn't need to be fixed.

dmutils/direct_plus_client.py
lfdebrux: Oh, thanks for the nit, I meant to log the `e` not just reraise it

PR #606, opened 2021-04-09
Mark this package as having type hints

bjgill:
Trello: https://trello.com/c/DZ28pqhm/90-try-to-add-type-hints-to-content-loader

As per [PEP 561](https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages), we need to let type checkers know that this package has type hints. We do this by adding an empty `py.typed` file, and adding it to `setup.py`.

mypy will now recognise that this package is typed. So it will check the types when type checking things that depend on dmutils.

Also add type hints to the filters module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 58.0.0
+
+### Breaking changes
+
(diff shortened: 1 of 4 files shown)


lfdebrux (approved):
🎉 🚀

PR #608, opened 2021-04-21
Ensure that the tests are run at least weekly

bjgill:
We make changes to this repo infrequently. So there can be long gaps between test runs. Add weekly tests on a Sunday to ensure that we notice in a timely manner if the tests break.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,8 @@
 name: Run tests
-on: [pull_request]
+on: 
+  pull_request:
+  schedule:
+    - cron: "0 0 * * SUN"
 jobs:
   test:
     runs-on: ubuntu-latest


alex9smith:
The fix for the failing tests is in https://github.com/alphagov/digitalmarketplace-utils/pull/609

This is a good idea and probably one we should add to other repos if it works. How will we know if a scheduled test fails?

bjgill:
That's the main problem - it only notifies the person who last modified `.github/workflows/test.yml`, which is not necessarily helpful.

Maybe we should add a step to notify Slack on failure?

In any case, this was just an experiment. It needs a fair bit more work, so I'll create a story on the tech board rather than continuing with it now.

gidsg:
@bjgill Sorry if I am missing something here, given these are unit tests, why is running them only after code changes not sufficient?

bjgill:
The problem this tries to solve is that we have an unpinned dependency - `digitalmarketplace-developer-tools`. This means that the tests can break without any code changes to this repo.

gidsg:
> The problem this tries to solve is that we have an unpinned dependency - digitalmarketplace-developer-tools. This means that the tests could break and we wouldn't notice until we came to make an (unrelated) PR.

@bjgill Makes sense, do you want me to research how we can propagate failures to Slack?

bjgill:
Am not planning at continuing with this any time soon

PR #610, opened 2021-04-28
Add dependabot to this repo

bjgill:
We've pinned some dependencies in [`setup.py`](https://github.com/alphagov/digitalmarketplace-utils/blob/main/setup.py). We should add dependabot to make sure that they're kept up to date

We should probably be reducing the amount of pinning anyway, but this is probably still a good first step.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+- package-ecosystem: pip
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
🤖

bjgill:
Flaky test. Will re-run, and also submit a PR to fix

PR #611, opened 2021-04-29
Disable Hypothesis deadlines in spreadsheet tests

bjgill:
By default, Hypothesis imposes a 0.2s deadline on tests and treats them as failing if they exceed this (https://hypothesis.readthedocs.io/en/latest/settings.html).

However, these spreadsheet tests can be slow - they're manipulating spreadsheets after all! So the tests sometimes fail because they slightly exceed the deadline. For example, https://github.com/alphagov/digitalmarketplace-utils/runs/1880009563?check_suite_focus=true and https://github.com/alphagov/digitalmarketplace-utils/runs/2467835277?check_suite_focus=true.

Disable the deadline for these tests so that they stop being flaky.

Fixes failure seen in https://github.com/alphagov/digitalmarketplace-utils/pull/610

### DIFF
--- a/tests/test_ods.py
+++ b/tests/test_ods.py
@@ -3,7 +3,7 @@ import functools
 
 import dmutils.ods as ods
 
-from hypothesis import strategies as st
+from hypothesis import strategies as st, settings
 from hypothesis import given, example
 
 po = functools.partial(mock.patch.object, autospec=True)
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
Do we actually need hypothesis in these tests? (or is that opening a whole can of worms)

bjgill:
I don't know enough about our use of Hypothesis to be able to answer you on that. I suspect they're useful, though, in saving us from having to list all the edge cases ourselves.

bjgill:
I think Github's having trouble. I merged this PR - https://github.com/alphagov/digitalmarketplace-utils/commit/33af50bc00db79ff7347731e1c93ff9af19e1f9f. But it doesn't seem to have linked it to this PR :shrug:

PR #639, opened 2021-05-12
Stop using the mock library for tests

bjgill:
mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards. We only support Python 3.6+. Switch across to use the standard library version to reduce our maintenance burden.

https://mock.readthedocs.io/en/latest/

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -3,7 +3,6 @@
 flake8
 freezegun
 hypothesis
-mock
 moto
 mypy
 pytest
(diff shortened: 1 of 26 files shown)


alex9smith (approved):

bjgill:
Ah. It's only a failure on Python 3.8+. I'll take a closer look.

bjgill:
Closing in favour of https://github.com/alphagov/digitalmarketplace-utils/pull/621 - once we fix that, we can resurrect this one

PR #650, opened 2021-06-01
Update to latest v1 Flask

bjgill:
https://trello.com/c/obh5gpQp/2250-update-to-flask-114

Stop pinning to v1.0.x, since v1.1.x was recently released. We're not ready to move to v2, but do want to be on the most recent v1.

Flask now controls the version of wekzeug, so we no longer need to pin wekzeug ourselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Digital marketplace utils changelog
 
-Records breaking changes from major version bumps
+Records breaking changes from major version bumps.
+
+## 59.0.0
+
+### Breaking changes
+
+Upgrade to Flask 1.1. This has some known
+[breaking changes](https://flask.palletsprojects.com/en/2.0.x/changes/#version-1-1-0).
 
 ## 58.0.0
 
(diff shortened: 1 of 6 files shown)


bjgill:
Ok. Tests should be working now. But the breakage suggests that this should be a major version bump

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Digital marketplace utils changelog
 
-Records breaking changes from major version bumps
+Records breaking changes from major version bumps. This has some known 
+[breaking changes](https://flask.palletsprojects.com/en/2.0.x/changes/#version-1-1-0).
```
alex9smith, line 4: Should this be on line 10 instead?

alex9smith:
A major change makes sense to me. Hopefully this doesn't break too many things in the apps that use utils and create a load of extra work for us.

CHANGELOG.md
bjgill: Good catch.

alex9smith (approved):
👍

PR #654, opened 2021-06-02
Replace mock with stdlib unittest.mock

bjgill:
https://trello.com/c/TIWfCbjY/2232-upgrade-mock-in-digitalmarketplace-utils

Mock is now part of the standard library. Start using the version in the standard library. One fewer dependency for us to manage.

Also stop mocking `current_app` so that it still works with the version of mock in Python 3.8. This is because mock now tries to probe for asynchronicity, which causes errors because there is no application context at that time: testing-cabal/mock@ad2b38d

Fixes the errors we encountered when trying to upgrade mock: https://github.com/alphagov/digitalmarketplace-utils/pull/621#issuecomment-840494969

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -3,7 +3,6 @@
 flake8
 freezegun
 hypothesis
-mock
 moto
 mypy
 pytest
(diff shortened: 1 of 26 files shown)


gidsg (approved):
👍 Ⓜ️

PR #664, opened 2021-06-14
Stop using dependabot for development dependencies

bjgill:
This matches how we use dependabot for the frontend apps. Currently, the bulk of dependabot PRs for this repo are for development dependencies (https://github.com/alphagov/digitalmarketplace-utils/pulls). I don't think keeping up to date with every patch version increment is a good use of our time and effort.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,5 @@ updates:
     interval: weekly
     time: "03:00"
   open-pull-requests-limit: 10
+  allow:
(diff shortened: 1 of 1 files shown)


gidsg (approved):
Good spot 👀  👍 
Would other repos benefit from this too?

bjgill:
As far as I know, dmrunner is the only other repo that's missing this, and it doesn't have separate development dependencies

PR #689, opened 2021-09-02
Allow query parameters for mailchimp list members

bjgill:
https://trello.com/c/xkosxLyt/128-2-automatically-extract-the-correct-set-of-emails-from-mailchimp

We have a method for getting an iterator of email addresses that belong to a list. We want to start using this method in a more restrictive way - e.g. limiting to status='subscribed'. Rather than try to anticipate all query parameters we might want, let users specify parameters freely and let the mailchimp API decide whether they're OK or not.

Will unblock work to script getting emails for the 'framework coming' email.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app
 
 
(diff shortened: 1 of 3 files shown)


alex9smith (approved):
👍

PR #706, opened 2021-11-04
Stop using custom www_authenticate handling logic

bjgill:
https://trello.com/c/oWvOqn00/2276-upgrade-to-flask-v2

Werkzeug added this functionality in 0.15.0. We're currently using Werkzeug 1.0.1. So we can safely retire our custom logic. The test shows that this does not change the bahaviour we rely on.

This allows us to add support for Flask 2.0. The tests now all pass. Note there are deprecation warnings with Flask 2.0 that will become errors in 2.1. We cannot fix them (and thus add support for 2.1) until we drop support for 1.0.

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app
 
 
-__version__ = '60.0.0'
+__version__ = '60.1.0'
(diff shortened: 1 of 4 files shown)


bjgill:
Note - I've tested locally that the tests all still pass with both Flask 1.x and 2.0

iceraluk (approved):

PR #722, opened 2021-12-09
Add support for arbitrary marketplace.team environments

bjgill:
https://crowncommercialservice.atlassian.net/browse/OST-114

We're starting to get a much larger number of such environments - currently 4, soon to rise to 5.

Rather than have to explicitly add support for each new environment, it would be much easier to assume that the naming convention we've used thus far is going to continue.

Related to https://github.com/Crown-Commercial-Service/digitalmarketplace-scripts/pull/745

### DIFF
--- a/dmutils/__init__.py
+++ b/dmutils/__init__.py
@@ -2,4 +2,4 @@ from . import config, formats, logging, proxy_fix, request_id
 from .flask_init import init_app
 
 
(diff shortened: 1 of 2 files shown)


tim-s-ccs (approved):
Good idea, and should hopefully save us some work going forward 😄

PR #744, opened 2022-03-01
OST-509-Update-Banner-Datetimes-For-G13

tberey:
Updates the banner controls, so that variables/controls are G13 related, and also updates the times for the new G13 open/close windows.

### DIFF
--- a/dmutils/dmp_so_status.py
+++ b/dmutils/dmp_so_status.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 G13_OPEN = datetime(2022, 3, 9, 10, 0, 0)
-G13_CLOSE = datetime(2022, 5, 18, 15, 0, 0)
+G13_CLOSE = datetime(2022, 5, 18, 16, 0, 0)
 
 
 def are_new_frameworks_live(params) -> bool:
--- a/dmutils/dmp_so_status.py
+++ b/dmutils/dmp_so_status.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 G13_OPEN = datetime(2022, 3, 9, 10, 0, 0)
-G13_CLOSE = datetime(2022, 5, 18, 16, 0, 0)
+G13_CLOSE = datetime(2022, 5, 18, 15, 0, 0)
 
 
 def are_new_frameworks_live(params) -> bool:
--- a/dmutils/dmp_so_status.py
+++ b/dmutils/dmp_so_status.py
@@ -1,8 +1,8 @@
 from datetime import datetime
 
-DOS6_OPEN = datetime(2022, 1, 14, 12, 0, 0)
-DOS6_CLOSE = datetime(2022, 2, 24, 15, 0, 0)
+G13_OPEN = datetime(2022, 3, 9, 10, 0, 0)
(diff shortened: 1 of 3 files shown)


dmutils/dmp_so_status.py
```diff
@@ -1,8 +1,8 @@
 from datetime import datetime
 
-DOS6_OPEN = datetime(2022, 1, 14, 12, 0, 0)
-DOS6_CLOSE = datetime(2022, 2, 24, 15, 0, 0)
+G13_OPEN = datetime(2022, 3, 9, 10, 0, 0)
+G13_CLOSE = datetime(2022, 5, 18, 15, 0, 0)
```
bjgill, line 4: Is this BST or UTC?

bjgill (commented):
Change looks sensible - just some obvious test failures.

Also, I believe we're closing at 3pm BST. Does that match with the `datetime` default?

dmutils/dmp_so_status.py
tberey: I'm not sure, but there's no clock going forwards and backwards between DOS6 and G13, which means it won't matter which (whoever did this before me, will be working on the same BST or UTC times)

tberey: if you agree with that?

tberey: (if i've understood you correctly, that is)

tberey: I believe this is on our GMT/BST times, as G13 has the same times (not dates) as DOS6, and the entry was `15, 0, 0` for DOS6, and we know that this was 3pm our time on the 24th Feb

tberey: also, i think at the moment GMT = UTC, we're not on BST right now

bjgill: At the moment, you're correct that BST == UTC. However, will this still be true in May when G13 closes?

tberey: oooooohhh snap, I was so focused on the open times (and that it fell before clock changes)! Excellent spot!

tberey: Fixed!

bjgill: I've just looked at the API docs (https://docs.python.org/3/library/datetime.html#datetime.datetime), and I think what you've got here works fine.

> [](https://docs.python.org/3/library/datetime.html#datetime.datetime.now)classmethod datetime.now(tz=None)
> Return the current local date and time.

It looks like both `datetime.datetime` and `datetime.now` default to the local timezone.

bjgill: @tberey - see my previous comment. I think you need to either:

1. set the closing time to 3pm and leave tzinfo=None (the default)
2. set the closing time to 4pm and set tzinfo=UTC (?)

tberey: Oh, in python trying to be convenient, that's really quite inconvenient that is does that - everything I've worked with uses UTC, although I guess we're lucky 50% of the time that lines up with GMT/BST

tberey: @bjgill I've left as the default, i.e. tzinfo=None

bjgill (approved):
🎉  🕒

